PR #9, opened 2020-02-06
make cross-compilation easier

jpellegrini:
Hello!

This is to make it easier to cross-compile. I have used it to compile STklos for OpenWRT and run it on a wireless router (see here: https://gitlab.com/jpellegrini/openwrt-packages).

* Always use `COMP`, `STKLOS_BINARY`, `GENLEX` variables in
  Makefile.am files, and honor their previous values
  if they exist. This makes it easier to cross-compile:
  having STklos locally installed, we may just do

```
  COMP=/usr/local/bin/stklos-compile \
  GENLEX=/usr/local/bin/stklos-gelnex \
  STKLOS_BINARY=/usr/local/bin/stklos \
  ./configure && make
```

  For this to work, the utils/tmpcomp script had to be
  slightly modified, so it will accept the same arguments
  as stklos-compile, with the '-o' switch (but with this
  patch, it also works if called without it).

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -9,8 +9,10 @@ schemedemo_SCRIPTS  = fork-test hello secho \
 
 SUFFIXES            = .stk
 
+COMP ?= ../utils/tmpcomp
+
 .stk:
-	../utils/tmpcomp $*.stk $*
+	$(COMP) $*.stk -o $*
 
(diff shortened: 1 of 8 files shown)


egallesio:
I have merged your PR. 
Thank you for this contribution

PR #12, opened 2020-05-13
Fix read-bytevector! to return actual read count

okuoku:
R7RS states `read-bytevector!` should return actual read count, instead of to-be read count.

https://small.r7rs.org/attachment/r7rs.pdf

> Returns the number of bytes read.

### DIFF
--- a/src/port.c
+++ b/src/port.c
@@ -464,7 +464,7 @@ DEFINE_PRIMITIVE("%read-bytevector!", d_read_bytevector, subr4,
 
   if (n && !count)
     return STk_eof;
-  return MAKE_INT(n);
+  return MAKE_INT(count);
 }
 
 


okuoku:
@egallesio : Hmm it seems you have reverted the change
https://github.com/egallesio/STklos/blob/9c1cd13da890310b0704a3dc2452947859e67809/src/port.c#L467

Was there anything wrong?

egallesio:
> @egallesio : Hmm it seems you have reverted the change
> 
> https://github.com/egallesio/STklos/blob/9c1cd13da890310b0704a3dc2452947859e67809/src/port.c#L467
> 
> Was there anything wrong?

Oops, my bad. I don't know  how I have reverted it. It is restored now. 
Thanks a lot for seeing it .

okuoku:
@egallesio : Reverted again

https://github.com/egallesio/STklos/commit/51d26adc224a718ad7f8b13bd82b7f2f024458af#diff-9040643e0656f2a0330f5b2bf95a4f60

egallesio:
I really don't understand how it is possible. I have restored your patch (and added a couple of test to avoid it happens again). 
Hopefully, you saw it again. Thank you so much.

PR #13, opened 2020-05-15
Use va_end() after va_start()

jpellegrini:
No big deal, but makes gcc happier when running with --pedantic


egallesio:
Thanks for this patch

Issue #11, opened 2020-02-24
STklos tests crash on some big-endian machines


jpellegrini:
This is a separate issue from #10 so I opened this new one.

I have ran the STklos tests on some machines:

```
machine		end	mem	stklos-compile	test results
qemu arm	L	256	fail		OK
qemu malta be	B	256	fail		segf call/cc
			128	fail		segf call/cc
qemu malta le	L	256	fail		OK
Archer c7v4	B	128+64	fail		segf call/cc
DIR-505		B	64	fail		segf call/cc,circular (*)
Asus RT-AC51U	L	64+32	fail		OK
			64	fail		OK
x86_64		L	huge	OK		OK
```

(*) I don't know why this one segfaults the circular structures test.
    It is an old router, running with the root filesystem on a pendrive,
    and low on memory. Maybe it had not enough memory to finish.

Memory is in Mb.
"malta" is emulated MIPS.
In the memory column, "128+64" means "128Mb RAM plus 64 swap-on-compressed-ram". Tested with different amounts of memory to see if it made a difference in the call/cc test
I wanted to test the qemu MIPS be version with 64Mb, but it's not enough for it to boot.

On little-endian machines, all runs well. On big-endian ones,
```
# stklos --no-init-file --utf8-encoding=yes -f do-test.stk
==== Testing Numbers  ...                                        passed
==== Testing Keywords  ...                                       passed
==== Testing call/cc & dynamic-wind  ...                         Segmentation fault
```
If I comment out the call/cc tests, the rest runs fine.

The test logs end like this:
```
==== Testing call/cc & dynamic-wind  ==================================
call/cc =========================================================
  testing foo.1 expects 1200 ==> OK.
  testing foo.1 expects 120 ==> OK.
  testing Al Petrofsky, 15 June 2002 expects 1 ==> OK.
```
So it seems that the test that failed is the one right after Al Petrofsky.

I have ran gdb, it seems to happen when cloning an environment:

```
# gdb stklos
...
(gdb) run --no-init-file --utf8-encoding=yes -f do-test.stk
Starting program: /usr/bin/stklos --no-init-file --utf8-encoding=yes -f do-test.stk
warning: Unable to find dynamic linker breakpoint function.
GDB will be unable to debug shared library initializers
and track explicitly loaded dynamic code.
==== Testing Numbers  ...                                        passed
==== Testing Keywords  ...                                       passed
==== Testing call/cc & dynamic-wind  ...                         
[...]

PR #14, opened 2020-06-08
Fix 32-bit big-endian MIPS crash, and another cosmetic change

jpellegrini:
- fix the problem mentioned in bug  #11
- do not return value from a void function


egallesio:
I have applied the PR you made. It is OK for me (even if I don't understand why changing the test order is significant for big-endian machines). Do you think that the issue #14 can be closed now? 
 
Thanks for you help.

PR #17, opened 2020-06-15
Adds SRFI-158 -- generators and accumulators

jpellegrini:
This is the example implementation from the SRFI.

* Relevant change in generator-fold:
    STklos includes #void where the SRFI reference implementation
    would expect #eof, and that would end up giving wrong results:
    for example, generator->list (gselect (generator) (generator)
    would return (#void), which has length one. Comparing against
    STklos' #void here seems to result in correct behavior.
* license: is it OK to use GPLv3? it is compatible with GPLv2. If not, I can
  change it.
* I added tests, but they're not complete. I used the exact examples from the
  SRFI itself as tests.


egallesio:
Hi @jpellegrini,
Thanks a lot for this contribution. It is merged in the master branch now.
To be a complete a new entry must be added in srfi-0.stk (I'll do that in a next commit).
I'll look also to add  SRFI 190 which is closely related (and easy to implement)

egallesio:
About the GPL3, it is OK for me.

PR #20, opened 2020-06-18
Adds SRFI-173

jpellegrini:
This is the example implementation from the SRFI, adapted to STklos.

One procedure was changed: instead of "assume" in hook-run,
an if followed by a call to error is done. This is consistent
with what is done in other SRFI implementations in STklos.

The tests are simple, but all procedures are tested.

### DIFF
--- a/doc/skb/srfi.skb
+++ b/doc/skb/srfi.skb
@@ -444,6 +444,7 @@ figure ,(ref :figure "Feature identifiers") in a
          (tr (tdl "random")      (tdl "srfi-27"))
          (tr (tdl "conditions")  (tdl "srfi-34, srfi-35, srfi-36"))
          (tr (tdl "hash-tables") (tdl "srfi-69"))
+         (tr (tdl "hooks")       (tdl "srfi-173"))
          (tr (tdl "generators")  (tdl "srfi-158, srfi-190"))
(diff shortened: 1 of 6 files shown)


jpellegrini:
Ok, let's see if I got it right this time. I added an entry to srfi-0, and also the entry in doc/skb/srfi.skb.

jpellegrini:
Oops. Missed one thing. 5 minutes and I'll fix it.

jpellegrini:
Ok, it's ready.

egallesio:

I had to modify some files to be complete 

- In `Makefile.am`, I have added `srfi173.ostk`. It is really a pain
  to have to do it by hand, whereas it is so easy to do it with `GNUmake` extensions.  However, I don't want to induce a dependency on the `make` command, so ...
- I have added the symbol `hooks` in srfi-0.stk since you added as a synonym in the documentation of SRFI-55
- I have added an entry in the `*srfi*` A-list which is used to find SRFI names in the documentation and for the construction of the web page.

Definitively, the process needed to add a new SRFI is too much complicated, but I don't see how to simplify it. 

BTW, I have changed the copyright year in srfi-173 (and srfi-158 too) since it is  anterior to the srfi document.

PR #25, opened 2020-06-22
Adds SRFI-51

jpellegrini:
I don't know what the criteria would be to include synonims in SRFI-0.
I have added this as "rest-list", but I'm not sure it's needed.


egallesio:
Merged. 
Thanks

egallesio:
BTW, everything is OK with this PR ;-)

PR #26, opened 2020-06-23
Add SRFI-54 (Formatting)

jpellegrini:
This adds SRFI-54, Formatting.

Some remarks:
* I used the reference implementation, but had to change the order of procedures (otherwise it doesn't work)
* The tests are from the examples in the SRFI document. A few tests had to be slightly adapted; one of them will not run, but it is a somewhat convoluted test (uses SRFI-9, SRFI-26 and SRFI-14, and a similar one doesn't work on Racket)
* also included a synonim for SRFI-0

If antything needs to be changed, I can rework the code and update the pull request.


jpellegrini:
Wait, I forgot to include the name strings on the tests. I'll fix that and update the PR

egallesio:
I have modified a little bit the tests for SRFI-54, to mute some result displays which appears on standard output. 

I've seen tat you had an (unemployed) ` cat-output-and-value ` function for that in the tests. I have replaced it by the macro `mute-cat` to makes things easier. 

Anyway, there is still a problem  with the values printed (result always seems to be correct). For instance, `(cat 3 #t (cat 's #t))` will produce under the REPL `s3s"3s"`. The fist `s` should not be here (if I have correctly understood). 

I have placed this on my TODO list, but in the meantime all the tests pass and there is no weird traces when running the tests.

PR #31, opened 2020-06-25
Fix SRFI-45 name in SUPPORTED-SRFIS

jpellegrini:
Hi!
The same problem that was on the website seems to also affect the SUPPORTED-SRFIS file...


jpellegrini:
The recent commits seem to have fixed it.

PR #34, opened 2020-06-29
Add INSTALL-android

jpellegrini:
This file contains instructions to compile in Android platforms.

@egallesio - is the format OK? Markdown i sgreat for processing and producing all kinds of output, but not the best thing to read. If you prefer, I can convert it to plain text.


egallesio:
Thanks **a lot** for this contribution. 
Markdown is fine for me, since it is easy to read from the *github* or *gitlab* interfaces. 
I'm really happy of this port, which I would had even not envisioned or started without your help.

PR #37, opened 2020-06-30
Add SRFI-117

jpellegrini:
This is SRFI-117; The tests from the SRFI itself are included!

### DIFF
--- a/lib/srfi-117.stk
+++ b/lib/srfi-117.stk
@@ -1,0 +1,257 @@
+;;;;
+;;;; srfi-117.stk		-- Implementation of SRFI-117
+;;;;
+;;;; Copyright © 2020 Jeronimo Pellegrini - <j_p@aleph0.info>
+;;;;
+;;;;
+;;;; This program is free software; you can redistribute it and/or modify
+;;;; it under the terms of the GNU General Public License as published by
+;;;; the Free Software Foundation; either version 3 of the License, or
+;;;; (at your option) any later version.
+;;;;
+;;;; This program is distributed in the hope that it will be useful,
+;;;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;;;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;;;; GNU General Public License for more details.
+;;;;
+;;;; You should have received a copy of the GNU General Public License
+;;;; along with this program; if not, write to the Free Software
+;;;; Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307,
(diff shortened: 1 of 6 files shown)


lib/srfi-117.stk
```diff
@@ -0,0 +1,261 @@
...
+          list-queue? list-queue-empty?
+          list-queue-front list-queue-back list-queue-list list-queue-first-last
+          list-queue-add-front! list-queue-add-back! list-queue-remove-front! list-queue-remove-back!
+          list-queue-remove-all! list-queue-set-list!
+          list-queue-append list-queue-append! list-queue-concatenate
+          list-queue-map list-queue-map! list-queue-for-each)
+ 
+;;; This definition is from Chibi's SRFI-1 implementation.
+
+(define (last-pair ls) (if (null? (cdr ls)) ls (last-pair (cdr ls))))
```
egallesio, line 65: You can probably delete the definition of `last-pair` since it is is a primitive in STklos (and it is compatible with this definition).

jpellegrini: OK, I have deleted it and it passes all tests.

jpellegrini: One thing: it defines map!. If the user does

```
(require "srfi-1")
(requie "srfi-117")
```

Then SRFI-1's `map!` would be shadowed. Is it better to make SRFI-117 require SRFI-1, or to rename `map!` in SRFI-117?

jpellegrini:
One thing: it defines map!. If the user does

```
(require "srfi-1")
(requie "srfi-117")
```

Then SRFI-1's `map!` would be shadowed. Is it better to make SRFI-117 require SRFI-1, or to rename `map!` in SRFI-117?

egallesio:
> One thing: it defines map!. If the user does
> 
> ```
> (require "srfi-1")
> (requie "srfi-117")
> ```
> 
> Then SRFI-1's `map!` would be shadowed. Is it better to make SRFI-117 require SRFI-1, or to rename `map!` in SRFI-117?
> Currently it's renamed to `%list-queue-map!`.

Perhaps, you ca use the `required?` function. The code could be:
```scheme
(unless (provided? "srfi-1")
  (define (map! f lis)
    (let lp ((lis lis))
      (if (pair? lis)
        (begin
          (set-car! lis (f (car lis)))
          (lp (cdr lis)))))))
```
So  if `srfi-1` is already loaded, its `map!` function is used, otherwise you define a new one. If `srfi-1` is loaded after that, your code  will continue to work since sfi-1's `map!` is more general. 
What do you think?

jpellegrini:
Yes, that's perfect! I've pushed it again.

PR #39, opened 2020-06-30
Add SRFI-169 - Underscores in numbers

jpellegrini:
Hi,
SRFI-169, underscores in numbers.
Since this one makes changes to the C number parser, and
introduces a test/error macro, I'll be fine if you don't accept this 
PR. :)

* number.c was changed so it will first acept underscores
  where it'd only accept digitp, and then call a function
  to validate the positions of underscores, and remove them.
  After underscores are removed, processing continues as
  STklos already did in the past.
* The examples from the SRFI are included, and a few others.
  The tests cover parsing of integers, rationals, binary, hex
  and octal numbers.
* Included in SRFI-0
* Since only number.c was patched, no require is necessary to
  use this.
* A test/error macro was introduced in test.stk, to make it
  easier to test expected errors (otherwise we'd need to write
  a long "(with-handler ..." in every error test


egallesio:
Great, I really like to be able to use underscores in numbers, as in Go or Python. I think that this is a welcome addition to Scheme. 

The (small) problem I have with this SRFi is that it introduces a potential backward incompatibility. For now, `1_234` is a valid STklos symbol, and it should be considered as a number  now. 

I have merged your code in a private branch and I have already implemented a Scheme parameter (called `accept-srfi-169-numbers`) to change the behavior of the reader. Now we have two choices

   - use a simple `(require "srfi-169")` to accept the new syntax. The body of srfi-169.stk would be solely
      ```scheme
      (accep-srfi-169-numbers #t)
      ```
   - define the defaut value of the parameter to `#t` and let the user change back to the old behavior if needed.

I don't know what is the better solution. What do you think? 

BTW,
> A test/error macro was introduced in test.stk, to make it easier to test expected errors (otherwise we'd need to write a long "(with-handler ..." in every error test

You don't need to. If an error occurs the test will return  `*test-failed*`:

```scheme
(test "this test should fail" *test-failed* unbound-variable)
```

jpellegrini:
If you can change to SRFI-169 on-the-fly, you can change back, right? Both parsers are included?

Maybe a gradual change, with a deprecation period, and a note like this in the release notes and documentation:

"STklos reads sequences of numbers and underscores as symbols. STklos now can behave as it did in the past, or it can behave according to SRFI-169, reading those sequences as numbers.
The variable `accep-srfi-169-numbers`  controls this: 
* when it is #t, reading will be done as per SRFI-169; 
* when it is #f, the old STklos parser, which reads numbers and underscores a symbols, will be used.
The default, for now, is #f, but this is deprecated: in some future release, STklos will default to have this variable bound to #t "

Also, what about a command-line option to tell STklos what value should be used for the variable initially when it boots?

egallesio:
Finally, I have opted to define the default value of `accept-srfi-169-numbers` to `#t`. After all, the risk to have such symbols is low, and we can go back to the old behavior by setting it to `#f`. Having a deprecation period is OK if a message tell you that the construction you use will change, otherwise it is very probable you don't know it. Using it for numbers is probably to cumbersom  here. 

> Also, what about a command-line option to tell STklos what value should be used for the variable initially when it boots?

I try to avoid to have too numerous command line options. All the parameter objects could be potential options and the list can grow quickly.

lassik:
Thank you very much for implementing this SRFI! STklos may in fact be the first Scheme implementation to do it.

Reader flags have sometimes been used to do it:

* `#!r6rs`
* `#!fold-case`
* `#!no-fold-case`

There could perhaps be `#!srfi-169` and `#!no-srfi-169` reader flags.

lassik:
`#!r6rs` says to read using R6RS standard lexical syntax with no extra syntax from SRFIs or the implementation. There isn't a standard `#!r7rs` flag, but if there was, it could be used to disable extra SRFIs.

lassik:
Gauche has a `-r 7` command line flag for R7RS mode (if this flag is not given, the default is Gauche's classic mode).

Kawa has a `--r7rs` flag.

For comparison, https://github.com/schemedoc/usage lists the flags of many Scheme implementations.

PR #42, opened 2020-07-03
SRFI 87 already supported - document it

jpellegrini:
STklos already supports SRFI-87 out of the box, as part of
R7RS support. This commit only documents it, and adds a
couple of tests to test-srfi.stk (besides the tests already
in test-r7rs.stk).


egallesio:
Great. I did't see this one. Thanks! 
BTW, as you will see it is easier now to document SRFIs, I have made twos macro which generate the text for SRFIS. I have used them for the all the SRFIs starting from 100.

PR #45, opened 2020-07-03
Adds SRFI-37 - args-fold: a program argument processor

jpellegrini:
Hi @egallesio !

I know STklos already has a command line argument processor, but I thought it'd be interesting to add a standardized one anyway... Here's SRFI-37!

The implementation was adapted from Guile's, as well as the tests.

(Actually, I found that several implementations of SRFI-37 do not comply with the spec; I'll file bug reports on those)


jpellegrini:
The possible bug I mentioned is being discussed in the SRFI-37 mailing list... There may be a post-finalization note to the SRFI.

Issue #46, opened 2020-07-04
(R7RS) current-{output,input,error}-port should be  parameter objects


jpellegrini:
Hello,

According to R7RS, `current-output-port` should be a parameter object, so this:

```
(define f (open-output-file "oo"))
(parameterize ((current-output-port f)) (display "hey"))
(close-output-port f)
```

should write `"hey"` to the file `"oo"`. But STklos does not let me do that:

```
stklos> (define f (open-output-file "oo"))
;; f
stklos> (parameterize ((current-output-port f)) (display "hey"))
**** Error:
%parameter-dynenv-pop!: bad parameter `#[primitive current-output-port]'
	(type ",help" for more information)
```

I see that these are implemented as primitives in `port.c`.

Anyway -- the current  STklos documentation on `current-output-port` says it returns the current output port, and doesn't mention parameterize, but it is listed with an argument "`obj'" in the documentation (but it will refuse taking an argument).

egallesio:
I'm aware of this problem and my first attempts to transform these functions in parameters was a fail. If I remember, I had problems with the integration with threads, as well as  the way they are  actually implemented). 

It is on my TODO list, but keeping it in this list of issues will force me to finish the work.

lassik:
Are there general difficulties with parameters in the presence of threads, or just with the current-*-port parameters?

egallesio:
It should be OK now.

jpellegrini:
I suppose we can close this... I have tested `current-output-port` with `parameterize` and also as `(current-output-port some-port)`, and it works fine. Same for error/input.

PR #47, opened 2020-07-04
Adds SRFI-64 -  A Scheme API for test suites

jpellegrini:
Some modifications were necessary:

* The test suite does not export certain %-named symbols, such as
  %test-begin, %test-end etc, because they are used internally
  only. However, they are used inside macros, and these are
  expanded outside the SRFI-64 module, where the symbols are
  not available.
  The simple solution for that was to export them. They are
  named in a similar fashion as STklos internal procedures,
  so there isn't much harm being done.

* Some unnecesary cond-expand were removed

* The SRFI-64 test API comes with a test suite, but STklos tests
  run STklos own test suite. The foreign (SRFI-64) test suite was
  modified to:
  - make no output to stdout;
  - at every test group end, record the test name and number of
    failures in a list;
  - after all tests are done, run through the list, calling STklos'
    "test" procedure with the original test name, expected value
    "0" and actual value equal to the number of tests failed for
    that group.
  This is not perfect: STklos will not report "expected a, got b"
  for individual tests, but rather "expected 0, but got 2" for
  the number of failures in a test group -- but it works, since
  the SRFI testing suite itself does record the individual test
  results in a file.

* Also, since te test suit for SRFI-64 is too large, it was
  put in a separate file (test-srfi-64.stk), which is loaded
  during the tests, in test-srfi.stk


jpellegrini:
Remarks:

* The test suite creates a log file in the `tests/` directory. I have not tried to inhibit this, since that log may be useful.
* Since STklos does not yet have `current-output-port` as a parameter object, I have used `%set-std-port!`.

I hope these are not a problem!

egallesio:
Now that `current-output-port` is a parameter, no need to use `set-std-port!` anymore (see Issue #46) 
To assure compatibility, the function is now defined in Scheme and placed in `lib/obsolete.stk`.

jpellegrini:
OK I'll re-work this PR as soon as I get home in a couple hours

jpellegrini:
I have changed it now. But it seems that wrapping the whole test suite in a `parameterize` doesn't work anyway... But no internal procedure is being used anymore!

egallesio:
> I have changed it now. But it seems that wrapping the whole test suite in a `parameterize` doesn't work anyway... But no internal procedure is being used anymore!

Weird! It seems to work for me and I don't see any use of `set-std-port!`

jpellegrini:
> It seems to work for me and I don't see any use of `set-std-port!`

Yes - I used

```
(define old-output-port (current-output-port))
(define the-output (open-output-string))
(current-output-port the-output)

(close-output-port the-output)
(current-output-port old-output-port)
```

in order to make the tests silent, because when I tried

```
(define the-output (open-output-string))
(parameterize ((current-output-port the-output))
...
```

STklos complained about an internal define not being allowed.

egallesio:
OK I see. This is kind of unavoidable: in fact, `parameterize` is a macro which places the body inside the body of the second lambda of a `dynamic-wind` .  Consequently, the code of the tests is something like 
```scheme
(dynamic-wind
    (lambda () ...)
    (lambda ()
      (define ....)
      (test-begin "...")
      (define ....))           ;; this define is unallowed
    (lambda () ...))
```
and Scheme imposes to place all  the definitions  at the beginning of a function.There are two solutions:
   - place the functions definitions at the top of the test file, or 
   - use in the `test-srfi.stk` a parameterize which contains the loading of `test-srfi64.stk`. 

I must admit that I have quickly  tried the second solution with little success (some prints (3) are on stdout), but perhaps I have not completely understood the logic of all the stuff. 

Anyway, I think that we can merge this PR as is.

PR #48, opened 2020-07-06
Adds SRFI-141 - Integer Division

jpellegrini:
Another SRFI... :)


egallesio:
> Another SRFI... :)

You are impressing! Merged

PR #49, opened 2020-07-08
Adds SRFI-143 - Fixnums

jpellegrini:
Hello @egallesio !

This is a pull request, but consider it also a question: how do you feel about unsafe optimizations for fixnums (and possibly flonums later)?
I have implemnted SRFI-143 (fixnums), fully -- including the "/carry" procedures.
Now, the SRFI mentions that implementations can -- if they want -- skip type checks for these procedures, in order to run faster. 

So, I have ran one simple loop in three different ways:

1. using standard Scheme procedures
2. using fx... procedures as they are currently implemented in STklos
3. using fx... procedures without type checks

The loop is this:
```
(define n 100_000_000)  ;; look, there are underscores in the numbers ! :)

(define (sum n)
  (let ((s 0))
    (dotimes (i n)
      (set! s (- (+ (- s) (* 2 i)) 5)))
    s))

(define (fxsum n)
  (let ((s 0))
    (dotimes (i n)
      (set! s (fx- (fx+ (fxneg s) (fx* 2 i)) 5)))
    s))

(begin (print "SUM")
       (time (sum n))
       (print "FXSUM")
       (time (fxsum n)))
```

In the interpreter,
```
generic int ==> 16s
fx, type checked ==> 12s
fx, not type checked ==> 11s
```

Compiled with `stklos-compile`:
```
generic int ==> 14s
fx, type checked ==> 11s
fx, not type checked ==> 10s
```

It's not a real benchmark, I know... It only hints at the relevance of those type checks.

What do you think? Right now this pull request removes the type checks on `fx` procedures, but I can quickly rework it so the procedures are all type checked. Or, if you don't think it's ok to change the current fixnum code, I'm fine if you don't accept it.

About the pull request:

* it introduces `INT_LENGTH` in `stklos.h`, since the length of fixnums is used a few times in the code;
* includes `stdlib.h` and `math.h` in `fixnum.c`, because these are needed in a few places for SRFI-143
* introduces the inline, static `exp_2_fxwidth` function, which returns the highest power of 2 representable as fixnum (also used a couple of times)
* includes an inline, static helper function for division with carry
* includes a couple of new error functions for fixnums
* passes all tests for the reference implementation (although no code is shared with it)
*I'm not sure how to work the documentation in this case -- there's no Scheme file like `srfi-143.stk`; it's all in `fixnum.c`, `stklos.h` and `boot.stk`...
The rest is probably what you'd expect for fixnum code, I suppose.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -78,6 +78,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-135: Immutable Texts
     - SRFI-137: Minimal Unique Types
     - SRFI-141: Integer Division
+    - SRFI-143: Fixnums
     - SRFI-145: Assumptions
     - SRFI-151: Bitwise Operations
     - SRFI-156: Syntactic combiners for binary predicates
--- a/lib/boot.stk
+++ b/lib/boot.stk
@@ -69,6 +69,15 @@
 (autoload        "srfi-48"        srfi48:help srfi48:format-fixed)
 
 ;==============================================================================
+;; give new names to some primitives and define some constants so as to comply
+;; with SRFI-143 (fixnums):
+(define fxquotient fxdiv)
+(define fxremainder fxrem)
+(define fx-width (fixnum-width))
+(define fx-greatest (greatest-fixnum))
+(define fx-least (least-fixnum))
+
+;==============================================================================
 ;; Execute the REPL only if a file was not given on the command line
 (define %before-exit-hook void)
 
--- a/lib/srfis.stk
+++ b/lib/srfis.stk
@@ -181,7 +181,7 @@
     ;; 140  Immutable Strings
     (141 "Integer Division" (integer-division) "srfi-141")
     ;; 142  ....... withdrawn
-    ;; 143  Fixnums
+    (143  "Fixnums")
     ;; 144  Flonums
     (145 "Assumptions")
     ;; 146  Mappings
--- a/src/fixnum.c
+++ b/src/fixnum.c
@@ -25,6 +25,9 @@
  */
 
 #include "stklos.h"
+#include <stdlib.h>  /* for ldiv */
+#include <math.h>    /* for sqrt & pow */
+
 
 static void error_bad_fixnum(SCM obj)
 {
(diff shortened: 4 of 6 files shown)


jpellegrini:
One thing: I tried to optimize as much as reasonable the bit-mangling procedures (rotate, reverse, bit field stuff), using bit shifts and all. If anything breaks, I'll try to fix! :)

jpellegrini:
Hi @egallesio ! 
I had some time today so I adapted this one and pushed it again. 
If you don't have interest in having fixnums tell me, then we can close this one... Or, if there are things to change, I may look into them later when I come back.

jpellegrini:
Hi @egallesio !
SRFI 217 (integer sets) has a reference implementation that uses fixnums for speed... Maybe it would be nice to have SRFI-143 (this one) and then implement 217?
Or, if you think it'd be necessary to change my PR, tell me (include type checks, maybe?)

egallesio:
Hi @jpellegrini,

> I had some time today so I adapted this one and pushed it again.

Glad to see that you are well.

I will try to merge this PR asap. 
I  was undecided  about doing checks or not.  However, since results can be incorrect with perfect fixnums (mainly because of overflows), perhaps it is superfluous to test if the arguments are correct fixnums.

```scheme
tklos> (fx+ (- (expt 2 61) 1) 0)
2305843009213693951
stklos> (fx+ (- (expt 2 61) 1) 1)
-2305843009213693952
```

So,you ca let the code as is. 
Thanks for this PR.

jpellegrini:
> I was undecided about doing checks or not. However, since results can be incorrect with perfect fixnums (mainly because of overflows), perhaps it is superfluous to test if the arguments are correct fixnums.

Well, yes. Common Lisp allows people to choose how safe they want to be with arithmetic; SRFI-143 has this,

>  If any argument is not a fixnum, or if the mathematical result is not representable as a fixnum, it is an error: this is known as the fixnum rule. In particular, this means that fixnum operations may return a mathematically incorrect fixnum in these situations without raising an error

And this:

> Existing implementations employ different implementation strategies for fixnum procedures. Some implement the model specified by R6RS (overflows cause exceptions), some implement modular arithmetic (overflows “wrap around”), and others fail catastrophically. In programs that use fixnums instead of generic arithmetic, overflows are typically programming mistakes. 

So I suppose it is in the spirit of the SRFI that each implementation decides wether checks should be made or not.


Anyway... Checking overflow and underflow would be somewhat more complex, and I personally don't think it's necessary. But we could do that if you'd like.

jpellegrini:
> Common Lisp allows people to choose how safe they want to be with arithmetic

I forgot to finish that statement. :)

"but I think it's too complex and not necessary to allow that level of fine adjustment"

egallesio:
Hi @jpellegrini ,

I'm finally back. Sorry for the long absence.
I have finally merged this PR.
Thanks for your patience.

jpellegrini:
Hello @egallesio !
That's great. I'm working now on SRFI 217, which has a reference implementation using SRFI 143.

egallesio:
Anyway, we have not finished with this one :cry: . I had some warnings and there are functions which should be n-ary that are still binary. I correct all that stuff and commit asap.

jpellegrini:
> I had some warnings and there are functions which should be n-ary that are still binary.

Ops, sorry! I see now that `fxand`, `fxor`, `fxxor` were not implemented as n-ary!
As far as I can see, those are the only ones, right? Or did I miss something else?

egallesio:
> > I had some warnings and there are functions which should be n-ary that are still binary.
> 
> Ops, sorry! I see now that `fxand`, `fxior`, `fxxor` were not implemented as n-ary!
> As far as I can see, those are the only ones, right? Or did I miss something else?

Yes they are the only one, and I have seen also a small bug, that I'll correct soon
Anyway, I'm still trying to find a way to enable tests if they are wanted (using a parameter). I have a partial solution, and hope to finish it tomorrow or the day after.  The thing that I don't like actually  is that a call such as `(fxmax 'bug)` returns `bug`, that is something that is not even a number. The fact that the computation of a result with a fixnum function is not correct is probably acceptable, but that the result is not a number is more arguable.  Of course, we can actually have a special treatment for this particular case, but that mean that we'll have a control for this function and not for the others, which is something which is not suitable imho. 
More on this soon, I hope (need to  prepare my tomorrow course).

PR #50, opened 2020-07-09
Add SRFI-128 - Comparators (reduced)

jpellegrini:
Only one test from the SRFI reference implementation doesn't pass:

```
test 102 on (= (comparator-hash default-comparator "t") (string-hash "t")) expected #t but got #f
```

However, if I do it on the REPL,

```
stklos> (load "lib/srfi-128.stk")
stklos> (define default-comparator (make-default-comparator))
;; default-comparator
stklos>  (= (comparator-hash default-comparator "t") (string-hash "t"))
#t
stklos> (comparator-hash default-comparator "t")
26797187
stklos> (string-hash "t")
26797187
```

So I don't know what's happening...

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -65,7 +65,8 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-111: Boxes
     - SRFI-112: Environment Inquiry
     - SRFI-117: Queues based on lists
-    - SRFI-118: Simple adjustable-size strings 
+    - SRFI-118: Simple adjustable-size strings
+    - SRFI-128: Comparators (reduced)
     - SRFI-129: Titlecase procedures
     - SRFI-141: Integer Division
     - SRFI-145: Assumptions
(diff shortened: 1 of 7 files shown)


jpellegrini:
If fails if `srfi-13` is loaded before this SRFI (and the test suite does that). I'll keep checking.

jpellegrini:
This is the problem:

If SRFI-128 is loaded first,
```
stklos> (load "lib/srfi-128.stk")
stklos> (string-hash "r")
26797185
stklos> (require "srfi-13")
"srfi-13"
stklos> (string-hash "r")
26797185
```

If SRFI-13 is loaded first,

```
stklos>  (require "srfi-13")
"srfi-13"
stklos> (string-hash "r")
114
stklos>  (load "lib/srfi-128.stk")
stklos> (string-hash "r")
114
stklos>
```
and in this last case, the test fails.

jpellegrini:
Okay - the `string-hash` procedure in this SRFI conflicted with the one in SRFI-13. So -- I made it use SRFI-13's `string-hash` instead, and now all tests pass! :)

egallesio:
Merged (finally!). Thanks again.

PR #54, opened 2020-07-10
Add SRFI-156 - Syntactic combiners for binary predicates

jpellegrini:
There was one macro which the reference implementation contains, using `syntax-rules`, which did not work in STklos, so I write another, using `define-macro`. All tests pass.

* the problem was that STklos introduces identical symbols in different phases of macro expansion using syntax rules:

```
(define-syntax f
  (syntax-rules (k)
    ((f) #f)
    ((f k) 'K)
    ((f k x ...) (a (f x ...)))))
```

Here's the STklos expansion:

```
stklos> (macro-expand* '(f k k k))
(a (a 'K))
```

some other implementations will return
```
(a.0 (a.1 'K))
```

Anyway, not a problem, since we have `define-macro`!

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -66,6 +66,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-117: Queues based on lists
     - SRFI-141: Integer Division
     - SRFI-145: Assumptions
+    - SRFI-156: Syntactic combiners for binary predicates
     - SRFI-158: Generators and Accumulators
     - SRFI-169: Underscores in numbers
     - SRFI-171: Transducers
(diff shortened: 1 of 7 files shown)


jpellegrini:
> `stklos> (macro-expand* '(f k k k)) (a (a 'K))`

Actually, this seems to affect other SRFIs too. I'll take a look. It seems like, when a macro introduces a new symbol, it should be generated by `gensym`, and this is not happening...

jpellegrini:
I have filed issue #55 about this.

egallesio:
OK. I have sen issue #55. I suppose it will not be simple to correct (I did not  wrote this code, and I always have difficulties with it). 
I 'll look at the issue asap.
In the meantime, I merge this pool request.

Issue #55, opened 2020-07-11
STklos introduces identical symbols in different phases of macro expansion using syntax rules


jpellegrini:
STklos introduces identical symbols in different phases of macro expansion using syntax rules:

```
(define-syntax f
  (syntax-rules (k)
    ((f) #f)
    ((f k) 'K)
    ((f k x ...) (a (f x ...)))))
```

Here's the STklos expansion:

```
stklos> (macro-expand* '(f k k k))
(a (a 'K))
```

some other implementations will return

```
(a.0 (a.1 'K))
```

I rewrote some macros for SRFI-156 and SRFI-41 using `define-macro`, but it feels like the wrong way to go... As far as I could see, SRFI-86 would also need some macro rewriting (and its implementation is quite large).

jpellegrini:
Perhaps, when fixing this, it would be nice to consider using the new macro-related SRFis by Marc Nieper-Wißkirchen?
https://srfi.schemers.org/?keywords=syntax

egallesio:
Hi @jpellegrini. 

It is clear that the  macro system must be rewritten, but hygienic macros are far above my head. This is why the implementation of macro is so poor in STklos :disappointed: .

 About this bug, I have found that there is already a kludge in the source that I have added for this problem ( it was added in 2002 from the comment, and I don't remember all the story). By using a `%%` in front of the symbol, it will be gensymed.

So 
```scheme
(define-syntax f
  (syntax-rules (k)
    ((f) #f)
    ((f k) 'K)
    ((f k x ...) (%%a a %%a (f x ...)))))

stklos> (macro-expand '(f k k k))
(G200 a G200 (f k k))
```

I know that this is far form ideal, but in the meantime you could perhaps use this *turnaround*.

jpellegrini:
Thanks @egallesio ! :)
I'll try to rework this PR today to use that, then...

jpellegrini:
Anyway - where is the macro code of STklos located?

egallesio:
> Anyway - where is the macro code of STklos located?

The code for non hygienic macro is in the compiler and the code for hygienic macro is in `lib/mbe.stk`

jpellegrini:
Ok, I understand how it works now...

I am reviewing the macro system options and SRFIs, articles, etc, related to them. It may take a while, but if you don't mind @egallesio , I'd like to try to come up with a new implementation for STklos.

egallesio:
[...]

PR #56, opened 2020-07-11
Add SRFI-41 - Streams

jpellegrini:
So, here is SRFI-41!

But I have some remarks. 

* modified to not use R6RS records
* one macro was taken from Chibi, because the reference implementation uses syntax-case
* I have created three modules, "streams-primitive", "streams-derived" and "srfi-41", which loads the other two. But I only included "srfi-41" in SRFI-0. I suppose it would be nice to have "srfi-41 primitive" and "srfi-41 derived", but I'm not sure. What do you think?
* Also corrected a little problem in SRFI-0 (SRFI-41 did not disappear! And SRFI-40 was withdrawn...)
* In order to take full advantage of the tests written by the original author, I have redefined (temporarilly) the error procedure, so it will print the error and continue (the tests actually match the exact error string)
 
And a more complicated one:

* I had to rewrite one macro using `define-macro`, see issue #55 .

Maybe it would be better to postpone accepting this PR until this is resolved (but I thought it would be nice to have it "almost ready" -- I can make small adjustments and push when a solution is chosen)

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -79,6 +79,8 @@ SRC_STK = bigloo-support.stk  \
           pretty-print.stk    \
           recette.stk         \
           slib.stk            \
+          streams-primitive.stk \
+          streams-derived.stk \
           srfi-1.stk          \
           srfi-2.stk          \
           srfi-4.stk          \
@@ -95,6 +97,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-35.stk         \
           srfi-36.stk         \
           srfi-37.stk         \
+          srfi-41.stk         \
           srfi-48.stk         \
           srfi-51.stk         \
           srfi-54.stk         \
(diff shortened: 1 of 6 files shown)


jpellegrini:
@egallesio I am not sure what is going on with this one. I have tested it back in July, and it worked OK.
Now tests don't seem to pass, with STklos complaining that variable `<stream>` is unbound. I thought this was due to a missing require on SRFI-9, but it doesn't seem to be the case (I did require it in even more places than necessary).

Since this is an important SRFI, part of R7RS large, I feel like well, there must be a way to get it to work... Do you have any idea why it is failing the tests?

egallesio:
It seems that the culprit is the rewriting of the compile function that I have done previously. It was supposed to be equivalent to the previous writing (module the order of the tests which should permit to redefine standard forms). But obviously this is not equivalent.
Commenting the `(not (symbol-in-env? first env))` at the start of the `compile` function, permits to load "srfi-61", but breaks (with a SIGSEGV :disappointed:) when creating the primes stream in the  tests.

I will try to have a look to see why the behavior is different asap.

jpellegrini:
Hm, I have seen some segfaults recently, but I thought it was bad RAM in my PC. I was going to test the memory before reporting.
Some of those segfaults happened when I compiled STklos in parallel, and vanished when I compiled with -j1. But some do happen with -j1.
I'll try to send a backtrace if that happens again.

jpellegrini:
@egallesio I reworked this PR, and it now applies cleanly to your master branch,
Also... It seems to pass the tests now! :)

So - I'll be away for a few weeks maybe, then I'll start working on porting SRFIs again!

egallesio:
I have merged it, but there is potentially  a bug  (as said in [previous comment](https://github.com/egallesio/STklos/pull/56#issuecomment-723115489)). I keep it on my bug list.

Thanks for this SRFI @jpellegrini

PR #58, opened 2020-07-11
Add a HACKING.md file

jpellegrini:
Wouldn't it be nice if newcomers could more easily work on STklos? I thought that maybe a `HACKING.md` file would be a good idea. (No?)

Is markdown OK?

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,234 @@
+# Hacking STklos
+
+This is a quick guide to STklos hacking. It's not detailed, so the
+document doesn't become huge, and also because after understanding the basics,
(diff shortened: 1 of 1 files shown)


egallesio:
> Wouldn't it be nice if newcomers could more easily work on STklos? I thought that maybe a `HACKING.md` file would be a good idea. (No?)

Definitively: YES

> Is markdown OK?

YES too

Thanks. 
Merged

egallesio:
I have fixed some typos and corrected the comment of the `gmp` directory (`gmp` is not a local copy of `GMP` ( it was, but `gmp` is really too big now to be embedded in the STklos distribution.

PR #62, opened 2020-07-13
Add SRFI-129 - Titlecase Procedures

jpellegrini:
I suppose it would make sense to use SRFI-129, since it is more general than the already present `string-titlecase` in STklos, and it offers two more procedures for characters. The tests added have a couple of comment marks at the end of the line, like this:

```
(test "string 3" "Ss" (string-titlecase "\x00DF;"))         ;
```

to show where this SRFI implementation differs from the embedded `string-titlecase` in STklos.

This SRFI exports

```
char-title-case?
char-titlecase
string-titlecase
```

One thing: if I create the module and export `string-titlecase`, the old one (defined as a primitive) will still be used, so I had to export a procedure `%string-titlecase` and later, after importing `SRFI-129` into the `STklos` module, redefin it:

```
;; If we just define the new string-titlecase in  SRFI-129
;; and import it, STklos will still use the old definition,
;; so we define %string-titlecase and explicitly redefine
;; string-titlecase here:
(define string-titlecase %string-titlecase)
```

I'm not sure if you think this is OK, or if there is a better way of doing this.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -64,6 +64,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-111: Boxes
     - SRFI-112: Environment Inquiry
     - SRFI-117: Queues based on lists
+    - SRFI-129: Titlecase procedures
     - SRFI-141: Integer Division
     - SRFI-145: Assumptions
     - SRFI-156: Syntactic combiners for binary predicates
--- a/doc/skb/srfi.skb
+++ b/doc/skb/srfi.skb
@@ -628,6 +628,9 @@ described in this document (procedures
 ;; SRFI 117 -- Queues based on lists
 (gen-loaded-srfi 117)
 
+;; SRFI 129 -- Titlecase Procedures
+(gen-loaded-srfi 129)
+
 ;; SRFI 141 -- Integer division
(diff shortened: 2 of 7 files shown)


egallesio:
Hi @jpellegrini,

Imho, if  the `string-titlecase` is incorrect, it should be corrected. I will have a look at that and try to have a behavior compatible with SRFI-129. Furthermore,  by adding `char-title-case?` and `char-titlecase` in STklos, we could have a native implementation of this SRFI ! 

I prefer this option, to the pure SRFI one because, the actual `string-titlecase` (as `string-downcase` and `string-upcase`)  accepts two or three parameters). It will be more coherent. 

> ```
> ;; If we just define the new string-titlecase in  SRFI-129
> ;; and import it, STklos will still use the old definition,
> ;; so we define %string-titlecase and explicitly redefine
> ;; string-titlecase here:
> (define string-titlecase %string-titlecase)
> ```
> 
> 
> I'm not sure if you think this is OK, or if there is a better way of doing this.

I'm afraid that we have to do that. Note that I prefer here a `set!` which implicitly implies that  `string-titlecase` already exists and is overloaded with `%string-titlecase`.

egallesio:
In the meantime, I have merged this PR

Issue #63, opened 2020-07-13
SRFI-13: string-prefix? and string-suffix?  seems to always answer #f


jpellegrini:
Isn't this strange?

```
stklos> (require "srfi-13")
"srfi-13"
stklos> (string-prefix? "ab" "abc")
#f
stklos> (string-prefix? "abc" "ab")
#f
stklos> (string-prefix? "a" "ab")
#f
stklos> (string-prefix? "" "ab")
#f

stklos> (string-suffix? "" "ab")
#f
stklos> (string-suffix? "b" "ab")
#f
```

jpellegrini:
This is porbably related:

```
stklos>  (string-prefix-length ""  "")
**** Error:
string-prefix-length: bad starting index 0 on ""
	(type ",help" for more information)
```

jpellegrini:
`%%check-indexes-func` seems to calculate `max` as `-1`, then it fails here:

```
 (unless (and (integer? start) (<= 0 start max))
	(error proc "bad starting index ~S on ~S" start str))
```

jpellegrini:
And if I skip the index checking in `string-prefix-length`,  `%string-suffix-length` is called with `end1`=`-1`, which makes it bail out returning zero...

jpellegrini:
See pull request #64 ... I'm working on this.

jpellegrini:
I think PR #64  resolves this... Maybe you'll want to add more tests, or check something else?

jpellegrini:
Checked again, the bug is gone! :)

egallesio:
It seems OK. Thanks for the correction of this bug (and for the tests)

PR #64, opened 2020-07-13
Fix a couple of SRFI-13 bugs

jpellegrini:
Hi @egallesio !

I found these while working on SRFI-130...

* SRFI-13 should export `string-trim`, there is no `string-trim-left`
* in string-trim, the arguments to `%check-indexes` were in wrong order

I included a few tests, not yet complete (we can add more later)...

### DIFF
--- a/lib/srfi-13.stk
+++ b/lib/srfi-13.stk
@@ -45,7 +45,7 @@
 	  ;; Selection
 	  string-copy substring/shared string-copy!
 	  string-take string-drop string-take-right string-drop-right
-	  string-pad string-pad-right string-trim-left string-trim-right
+	  string-pad string-pad-right string-trim string-trim-right
 	  string-trim-both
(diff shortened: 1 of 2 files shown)


jpellegrini:
Ok, I have included more tests.
These are from the Chicken people, who assembled it using tests mostly from Gauche but also from Guile (as far as I could see).

jpellegrini:
Fixed several other little things; now, the KMP implementation seems to not be working. I'll look into this.

jpellegrini:
Okay, we have a large test set for SRFI-13...
And all tests pass now! :)

This also seems to be the fix to issue #63 ...

jpellegrini:
One more thing: I have modified SRFI-13 so it will properly shadow some of the string-handling procedures from R7RS.

egallesio:
Great work!!!

PR #66, opened 2020-07-15
More draft SRFIs

jpellegrini:
Just making the list complete...


jpellegrini:
And SRFI-189 is (as of today) final! Changed that also.

jpellegrini:
Github seems to be stuck checking mergeability of the PR... For several minues now!

It's certainyl not the commit -- it is only 6 lines.

jpellegrini:
WIll close and open again...

lassik:
Not sure if it's related, but the `master` branch does not build right now due to some problem with srfi-19.

PR #68, opened 2020-07-15
More draft SRFIs

jpellegrini:
Just updating SRFI-0...


jpellegrini:
Hey, it looks like we've found a bug in Github...

egallesio:
Merged. Thanks

Issue #72, opened 2020-07-16
ANSI color codes - repeated escape sequences


jpellegrini:
Hi,

It looks like the `ansi-color` procedure produces repeated color escape codes, and this confuses at least one terminal emulator (terminology, the one which comes with Enlightenment).

```
(string->list (string-append (ansi-color  'normal "aaa" 'clear) (do-color (get-repl-color :prompt) "bbb" 'clear  (get-repl-color :input) " -")
))
(#\escape #\[ #\0 #\m #\a #\a #\a #\escape #\[ #\m #\escape #\[ #\1 #\; #\3 #\5 #\m #\b #\b #\b #\escape #\[ #\escape #\[ #\0 #\; #\1 #\m #\space #\-)
```

The opening of the color code right after "bbb" is repeated, `\e[\e[`.

A bach script with this single line,

```
echo -e "\e[m\e[1;35mHello\e[\e[0;1m!!!\e[0m"
```
for example, produces 

```
Hello0;1m!!!
```

because the escape opening before "!!!" is repeated.

So, Terminology  prints the prompt with the correct color, but also prints `0;1m` between the prompt and the input. This also complicates readline, since the number of characters from cursor to the beginning of editable area is wrongly computed:

```
stklos>0;1m
```

As far as I could see, this happens when two adjacent color codes are passed to `do-color`.
For the prompt, I removed the `'clear` between prompt and input, and it worked  fine:

```
(define (make-prompt module)
  (let ((str1 (if (> repl-level 1) (format "[~A] " repl-level) ""))
        (str2 (format "~A>" (module-name module))))
    (repl-prompt (if (repl-prompt-use-color?)
                     (string-append
                        (do-color (get-repl-color :repl-level) str1 'clear)
                        (do-color (get-repl-color :prompt) str2 ;'clear   <== here, commented out
                                  (get-repl-color :input))
                        " ")
                     (string-append str1 str2 " ")))))
```

For the initial banner, I see that either adding an empty string,
```
      (display (do-color 'normal "" col1 "  \\    " col2 line1 'clear))
```
or not using the initial `normal`,
```
      (display (do-color 'normal "" col1 "  \\    " col2 line1 'clear))
```

also works. But these are workarounds...

I don;t know what the ANSI escape code spec says about those (and wether the fault is Stklos' or Terminology's), but it would be nice anyway to not output repeated or non-closed codes.

lassik:
ANSI codes are of the form:

[...]

PR #76, opened 2020-07-19
Fix sublist flattening bug in ansi-color (#72)

lassik:


egallesio:
Thanks @lassik for this one.

PR #77, opened 2020-07-19
Fix indentation in compiler.stk

jpellegrini:
(no big deal...)


egallesio:
You are right. This is because I use emacs `prettify-symbols-mode`. I'm really looking forward to have a decent support of ligatures
in emacs to avoid this (even if probably not useful in this particular case).

jpellegrini:
> I'm really looking forward to have a decent support of ligatures in emacs to avoid this

Have you tried FiraCode? 

https://github.com/tonsky/FiraCode

Although it's listed as "not working" in Emacs, there is a link to a workaround there...

PR #78, opened 2020-07-19
Add SRFI 175 - ASCII character library

jpellegrini:
Here's the ASCII character library!

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -76,6 +76,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-171: Transducers
     - SRFI-173: Hooks
     - SRFI-174: POSIX Timespecs
+    - SRFI-175: ASCII character library
     - SRFI-176: Version flag
     - SRFI-180: JSON
     - SRFI-190: Coroutines Generators
--- a/doc/skb/srfi.skb
+++ b/doc/skb/srfi.skb
@@ -665,6 +665,9 @@ described in this document (procedures
 ;; SRFI 174 -- POSIX Timespecs
 (gen-loaded-srfi 174)
 
+;; SRFI 175 -- ASCII character library
+(gen-loaded-srfi 175)
+
 ;; ----------------------------------------------------------------------
 ;; SRFI 176 -- Version flag
 ;; ----------------------------------------------------------------------
--- a/doc/skb/srfi.stk
+++ b/doc/skb/srfi.stk
@@ -87,6 +87,7 @@
      (171 . "Transducers")
      (173 . "Hooks")
      (174 . "POSIX Timespecs")
+     (175 . "ASCII character library")
      (176 . "Version flag")
      (180 . "JSON")
      (190 . "Coroutines Generators")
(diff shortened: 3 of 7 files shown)


lassik:
Thank you for submitting this as well :)

Many/most of the procedures in this SRFI are direct equivalents of char/string procedures in standard Scheme, only limited to ASCII. It may be possible to extend the code in `src/char.c` to implement them in C.

lassik:
For example:

```C
uint32_t STk_to_upper(uint32_t c) {
  if (STk_use_utf8) {
    int res = search_conversion_table(c, lower_table, lower_table_length);
    return (res <= 0) ? c : (uint32_t) res;   // -1: not a lowercase, 0 lower without upper
  } else
    return toupper(c);
}
```

Could become something like:

```C
uint32_t STk_to_upper(uint32_t c, int ascii_only) {
  if (STk_use_utf8 && !ascii_only) {
    int res = search_conversion_table(c, lower_table, lower_table_length);
    return (res <= 0) ? c : (uint32_t) res;   // -1: not a lowercase, 0 lower without upper
  } else
    return toupper(c);
}
```

jpellegrini:
Hmm, yes. That would be better. I'll do that

lassik:
If it's ok with @egallesio to do this at the C level. I would expect the result to be faster with less code.

lassik:
Note: The RnRS `char*` procedures only take char objects, but the SRFI 175 procedures can take either char objects or integer (fixnum) codepoints. The fixnum case will probably need to be handled somewhere in C.

jpellegrini:
I'll wait for @egallesio to decide then if he wants an optimized version or if this PR is OK

egallesio:
> I'll wait for @egallesio to decide then if he wants an optimized version or if this PR is OK

Sorry for the silence. Actually is a bad period for me with lots of problems . I will probably not be able to read all your contributions this week. I apologize for that and hope to be back soon.

jpellegrini:
> > I'll wait for @egallesio to decide then if he wants an optimized version or if this PR is OK
> 
> Sorry for the silence. Actually is a bad period for me with lots of problems . I will probably not be able to read all your contributions this week. I apologize for that and hope to be back soon.

Of course! I hope all problems get sorted out.

lassik:
Sorry to hear about the problems. Hope things turn out well, and no hurry - take as much time as you need.

egallesio:
Implementing this SRFI in C is definitively the good approach. However to avoid to overload the  initial interpreter, I have implemented the SRFI in a dynamically loaded file. I have added support for that in the Makefile of the `lib` directory (it was possible to have dynamic loaded binaries but only when STklos was installed, not at compilation time). I was reluctant to do this before, because there was too many systems, but things tends to be simpler now. Anyways, this implementation will need to be tested on *exotic* architectures (I have already tested it on GNU/Linux, MacOS and Android). The problem is that it doesn't compile on a given architecture, it will block the installation on it.

egallesio:
I will merge this implementation to steal the tests, the doc and so and replace the srfi-175.stk file to use the C implementation.

PR #79, opened 2020-07-20
Add SRFI-118 - Simple adjustable-size strings

jpellegrini:
This adds two procedures, written in C and placed in
str.c:

* string-append!
* string-replace!

An auxiliary function get_substring_size() was added also.

### DIFF
--- a/src/str.c
+++ b/src/str.c
@@ -41,6 +41,11 @@ extern SCM STk_make_bytevector_from_string(char *str, long len);
  *
  */
 
+static void error_bad_index(SCM s)
+{
+  STk_error("bad index ~s", s);
+}
+
 static void error_bad_string(SCM s)
 {
   STk_error("bad string ~s", s);
@@ -49,6 +54,11 @@ static void error_bad_string(SCM s)
 static void error_bad_character(SCM s)
 {
   STk_error("bad character ~S", s);
+}
+
+static void error_bad_string_or_character(SCM s)
+{
+  STk_error("bad string or character ~S", s);
 }
 
 static void error_change_const_string(SCM s)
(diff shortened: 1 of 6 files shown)


jpellegrini:
The first argumet may be a literal, because it seems to work:
```
stklos> (string-append! "a" "b")
"ab"
```
Then, there is SRFI-185, which seems to be trivially implemented on top of this one, and offers macros that require the first argument to be a place...

egallesio:
Great, another interesting SRFI!

> The first argumet may be a literal, because it seems to work in STklos:
> 
> ```
> stklos> (string-append! "a" "b")
> "ab"
> ```
> 

I think that we must verify that the fist string is not a constant one. Otherwise, we could have 

```scheme
stklos>  (define hello "Hello")
;; hello
stklos>   (define s1 hello)
;; s1
stklos>   (define s2 hello)
;; s2
stklos>   (string-append! s1 " world.")
"Hello world."
stklos> (list s1 s2)
("Hello world." "Hello world.")
```
that is the constant string `hello` whose value is changed! 

Testing that `str` is constant can be done with 

```c
if (BOXED_INFO(str) & STRING_CONST)   error_change_const_string(str);
```
What do you think?

src/str.c
```diff
@@ -677,6 +687,212 @@ DEFINE_PRIMITIVE("string-append", string_append, vsubr, (int argc, SCM* argv))
...
+      total_length += STRING_LENGTH(*(argv-i));
+    } else { /* we have already checked, it is a char! */
+      p += STk_char2utf8(CHARACTER_VAL(*(argv-i)),p);
+      /*      p += STk_utf8_char_bytes_needed(CHARACTER_VAL(p));*/
+      total_length ++;
+    }
+  }
+
+  STRING_LENGTH(q) = total_length;
+  fflush(stdout);
```
egallesio, line 751: could be deleted?

src/str.c
```diff
@@ -1381,6 +1597,8 @@ int STk_init_string(void)
   ADD_PRIMITIVE(strgei);
   ADD_PRIMITIVE(substring);
   ADD_PRIMITIVE(string_append);
+  ADD_PRIMITIVE(string_append_adjsize);
```
egallesio, line 1600: Could you use please dstring_append (for destructive string-append)? This is how function are named in C when they modify their first parameter and a non destructive version of the function exists.

jpellegrini: Sure! I'll work on this later (maybe tomorrow in the morning)

src/str.c
jpellegrini: Ha ha! Sorry, that's from debugging, I forgot to delete. Will do that.

jpellegrini:
> I think that we must verify that the fist string is not a constant one. Otherwise, we could have

Yes, I misread the SRFI! It says "The string must be a variable-size mutable string".

I'll fix that.

jpellegrini:
Ok @egallesio !

I think it's all fixed now -- the debugging leftovers, the mutable-setting bug, and the function name (it's `string_d...` as I see in the source code).

See if it's ok.

egallesio:
Once again, **great job** @jpellegrini !!!

PR #80, opened 2020-07-21
Add SRFI-94 - Type-Restricted Numerical Functions

jpellegrini:
This SRFI changes the semantics of some standard Scheme functions, so
I had to save the old definitions and recover them back at the end of
the tests:

```
(define make-rectangular %make-rectangular)
(define make-polar %make-polar)
(define quotient %quotient)
(define remainder %remainder)
(define modulo %modulo)
```

Also... The procedures are of course slow. I am willing to reimplement them in C, but I'm not sure it would be a good idea to grow the C code so often. Maybe, if this is to be optimized/reimplemented in C, this SRFI could dynamically load a `.so` file?

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -63,6 +63,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-87: => in case clauses
     - SRFI-88: Keyword Objects
     - SRFI-89: Optional Positional and Named Parameters
+    - SRFI-94: Type-Restricted Numerical Functions
     - SRFI-96: SLIB Prerequisites
     - SRFI-98: Interface to access environment variables
     - SRFI-100: define-lambda-object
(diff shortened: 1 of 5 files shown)


jpellegrini:
A dynamically loaded C library?

jpellegrini:
Only Chicken and SLIB seem to have this one. Is it worth working more on it?

jpellegrini:
Hi @egallesio !

I have fixed this PR so it applies cleanly. I'm still not sure if you'd like it to be added, but anyway -- here it is!

It's not a popular SRFI, but since you mention in STklos' home page that you'd like to ssupport as much as possible the SRFIs that are in "final" state, well, I thought I'd make the PR! :) 

If you think procedures need to be implemented in C, we may do that, no problem.

jpellegrini:
I forgot one thing!
This SRFI is using the same configure option to control parameters that is used for fixnums. We should probably change that...

egallesio:
Hi @jpellegrini,

This one is old, and it is time to merge it.  I think that we can keep it in Scheme, and if someone needs some speed, rewrite it in C, now that we have other SRFI written in C. 
Thanks for this work.

PR #82, opened 2020-07-24
Add "-V" to the manpage

jpellegrini:
Also included some missing punctuation


egallesio:
Oops, forgot the doc. Thanks.

PR #86, opened 2020-07-31
Add SRFI-185 - Linear adjustable-length strings

jpellegrini:
Some people dislike this SRFI, but it may be interesting at least for compatibility (if one needs to run a program that relies on this SRFI, it would be easy if it were available).

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -80,4 +80,5 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-175: ASCII character library
     - SRFI-176: Version flag
     - SRFI-180: JSON
+    - SRFI-185: Linear adjustable-length strings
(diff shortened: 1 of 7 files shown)


egallesio:
I'm not a fan of this SRFI too, I really prefer the SRFI 118 interface. Anyway,  I think that but I'm not obliged to use it :grinning: 
Merged.

egallesio:
I have changed the tests, because they are some invalid tests (mostly tests which uses the return value of the function string-append! which doesn't return anymore a value.
Two tests are tagged as deleted. Can you confirm that they cans be deleted?

jpellegrini:
> return value of the function string-append! which doesn't return anymore a value.
Ooops! Sorry!  

> Two tests are tagged as deleted. Can you confirm that they cans be deleted?
Yes, I guess so.

PR #92, opened 2020-08-02
Procedure signatures

jpellegrini:
Hello!

I was working on Geiser support for STklos, and noticed STklos has no way to obtain the signature of a procedure.
So I implemented it this way:

* in `proc.c`, a new field "signature", of type SCM, is added to closures
* When the compiler emits code for a lambda, it will do this:
  * The *first* `DOCSTRG` opcode brings the signature, so it fetches the next object, which is a list (the same argument list passed when the lambda was defined)
  * The *second* `DOCSTRG` opcode, if it exists, is the actual docstring.
* A primitive `"%procedure-signature` was added. It returns the list of formal parameters of a procedure
* In a second commit, the `help` method was changed, so it will separately show the signature before the docstring.

I thought it would be nice to have a parameter that tells the VM wether it should store signatures or not (if I want short memory usage and no debug info, I can turn signatures off). But the VM would not know when the bytecode is generated with or without signatures... A different opcode would solve it, but I'm not sure what you think of using another opcode just for this -- and I don;t know how to implement a new opcode in STklos)

Also -- methods won't show signatures, because (I suppose it is this, tell me if I'm wrong) `extended-lambda->lambda` will not attach their names to the closures.

Anyway - do you think this is OK?


jpellegrini:
I forgot to say -- with `%procedure-signature`, Geiser's autodoc would show the signature of a procedure when you position the cursor on top of its name.

jpellegrini:
> the VM would not know when the bytecode is generated with or without signatures

Of course it would! The docstring is a string, and the signature is a Scheme list! It just needs to check the type of the operand! 8-)
One could even emit the opcodes in different order (docstring <--> signature)...

I could make the parameter object in `boot.stk`, and while it's not bound (before boot), the VM would default to not generate signatures - right?

jpellegrini:
> the VM would default to not generate signatures

Obviously, I meant the compiler. The VM wouldn't need to know about the parameter object.

jpellegrini:
I also created a parameter object in the compiler, so one can disable signature generation to save memory.

I'm not sure where to document this parameter - or even if it should really be inside the compiler module (maybe it would be better if it was outside?)

* also added tests, for both signatures and docstrings...

jpellegrini:
Hm. Could also include the line number.

`DOCSTRG`: if it's a string, its the docstring; if it's a lis, it's the signature; if it's a fixnum, it's the line number where the closure was defined.

What about that?

egallesio:
Hi @jpellegrini,

It is a really interesting PR. I really like it and a support for STklos in Geiser is really a great news. Kudos
for this.


> * in `proc.c`, a new field "signature", of type SCM, is added to closures

This was the intent of the formal field (for the formal parameter list) which was in comment in the `closure_obj`. It was my intent to implement something like that,but **hooray** you did it.

> * When the compiler emits code for a lambda, it will do this:
>
>   * The _first_ `DOCSTRG` opcode brings the signature, so it fetches the next object, which is a list (the same argument list passed when the lambda was defined)
>   * The _second_ `DOCSTRG` opcode, if it exists, is the actual docstring.
> * A primitive `"%procedure-signature` was added. It returns the list of formal parameters of a procedure
> * In a second commit, the `help` method was changed, so it will separately show the signature before the docstring.
>
> I thought it would be nice to have a parameter that tells the VM wether it should store signatures or not (if I want short memory usage and no debug info, I can turn signatures off). But the VM would not know when the bytecode is generated with or without signatures... A different opcode would solve it, but I'm not sure what you think of using another opcode just for this -- and I don;t know how to implement a new opcode in STklos)

A different opcode is probably the best solution and less "hacky" for me.

I propose you use a dedicated opcode for that. You can replace `UNUSED-2`, which seems to have been reserved for that, by your opcode (PROC-SIG for instance). Declare the new opcode in the `assembler.stk` file by overriding `UNUSED-2`. That is all that you have to do normally. The UNUSED instructions are here to reserve the opcode numbers (and avoid the bootstrap headaches). Anyway, the simpler way to succeed in your bootstrap is to set all the support necessary for your implementation **without** producing the new opcode in the compiler.  Compile a new version with the creation of the 3 images and after that only produce effectively the new opcode. Since  your code does not modify the way the VM runs, it should not be a source of majors problems.

>
> Also -- methods won't show signatures, because (I suppose it is this, tell me if I'm wrong) `extended-lambda->lambda` will not attach their names to the closures.
>

Extended-lambdas are in fact  lambdas using  DSSSL style (lambdas with keywords an optional parameters). This function
rewrite a closure using the DSSSL notation in a standard Scheme function

```scheme
STKLOS-COMPILER> (pp (extended-lambda->lambda '(lambda (a :optional b :rest c :key d e)
                                                   (list a b c d e)))
(lambda (a . c)
  (let* ((b (and (pair? c)
                 (not (keyword? (car c)))
                 (let ((G176 (car c))) (set! c (cdr c)) G176)))
         (d (key-get c :d #f))
         (e (key-get c :e #f)))
    (let () (list a b c d e))))
STKLOS-COMPILER>
```

Methods are a different affair, even if I suspect that there is not a lot do. The problem is a method is associated to a generic function and that you'll have to use the MOP to find the methods of the generic function and then their signature. For instance (with %procedure-doc for the example):

```scheme
stklos> (define-method foo ((x <integer>)) "Doc of foo::int" 10)
stklos> (define-method foo ((x <string>)) "Doc of foo::string" 20)
stklos> (foo 1)
10
stklos> (foo "")
20
;; Find the methods of gf foo
stklos> (generic-function-methods foo)
(#[<method> 7f0baba3d720] #[<method> 7f0babcfd480])
;; find the procedures for this methods
stklos> (map method-procedure (generic-function-methods foo))
(#[closure 7f0baba3f0c0] #[closure 7f0babc87dc0])
;; find the documentation of this methods
stklos> (map (lambda (x) (%procedure-doc (method-procedure x))) (generic-function-methods foo))
("Doc of foo::string" "Doc of foo::int")
```
> Anyway - do you think this is OK?
Sure.

------------------------------------

> I also created a parameter object in the compiler, so one can disable signature generation to save memory.

This parameter could be initialized to #f if we are not in debug mode and to #t otherwise.
This permits to have a smaller image when we are not in debug mode and skip those info when unneeded.
Anyway I'm a little bit worried by the fact that all the functions will needed of the lib will generate a new signature and we will need to see if it does not cost too much for the lib. We can see that later of course.


About the line numbers: they are embedded in the source code . Source files are build with extended cons which contains the file, the line and the position of the pair. In fact `econs` are built by read when the read is told to be constant (this is the case when we compile a sexpr (hencefile information is already embedded in the signature of functions :smile:). How it could interact with Emacs is another story and I have no idea if it is easy to keep the source when sending a function form an emacs buffer to the compiler. For now the function `econs` is not exported to Scheme, but perhaps it will be needed here.

Then, the line information could be sent to the new opcode too with the signature. Consequently, we could
have an opcode with 3 (or 4) parameters
  - the signature,
  - the file name
  - the line number
  - eventually the position (which should be equivalent to the emacs point, btw)

What do you think?

jpellegrini:
Wow, thanks @egallesio ! I'll review all this later and work more on the PR.

jpellegrini:
So... One single opcode with several arguments.

What if there is no file name, line number etc (for example, when I start the REPL from the command line)?
Do we just put a negative number on line/position and false on the file name?
Wouldn't that waste space?

The signature is always available when the procedure is created, so it's OK to always set it. But when STklos runs from Emacs or some other IDE -- then it will be as if it was just a REPL runinng, and STklos will not know line numbers or file names.
So maybe a primitive to set such information would be needed, I guess?

jpellegrini:
Also, with STkos patched to include signatures, I get this:

```
stklos> (define (f a) a)
;; f
stklos> (%procedure-signature f)
(a)
stklos> (%epair? '(1))
#t
stklos> (define sig (%procedure-signature f))
;; sig
stklos> (%epair? sig)
#f
```

So the signature is a plain list, and not an extended one, it seems.

egallesio:
> So... One single opcode with several arguments.
> 
> What if there is no file name, line number etc (for example, when I start the REPL from the command line)?
> Do we just put a negative number on line/position and false on the file name?

Normally this information should always be present 
```scheme
stklos> (define x '(1))
;; x
stklos> (%epair? x)
#t
stklos> (%epair-file x)
"virtual input port"       ;; This could be enhanced
stklos> (%epair-line x)
1
stklos> (%epair-position x)
14
```

> Wouldn't that waste space?

Yes but you wanted to have file information to go back to the source. No? 
If so, I think that we can avoid it. It will take one pointer by function and we can use a STklos struct (or a vector) to put the information in the closure object. Hence, if we do not use signature, we only have one more (unused) pointer than now. If we use signatures, the triplet or quadruplet will be plugged in this supplemental  field. Note the file name could be shared between all the functions of the file. So, it costs only when we use it.

> The signature is always available when the procedure is created, so it's OK to always set it. But when STklos runs from Emacs or some other IDE -- then it will be as if it was just a REPL runinng, and STklos will not know line numbers or file names.

As said before it will know line numbers but they are probably false, since I suppose that each time you eval the code of a function; STklos will see it as a new expression and the line number will be augmented at each eval (furthermore, the file will be probably false).

> So maybe a primitive to set such information would be needed, I guess?
Probably. 

```scheme
> stklos> (define (f a) a)
> ;; f
> stklos> (%procedure-signature f)
> (a)
```

I suppose that this is the rewriting of the function in a lambda which introduce that (since it uses cons and not econs)
Is is also incorrect id you define `f` as

```scheme
(define f (lambda (a) a))
```

jpellegrini:
Ok... So I think line numbers are really harder than just the signature.

```
(define f #f)  ;; will be redefined later
...
(let ((...
          ...
            (set! f (lambda (...) ...))
           )))
```

So Emacs (or any IDE) will jut not know the line number where the procedure `f` was defined, unless it stores the lines where *all* variables are defined (not only procedures) --and this has to be done in Emacs (when `f` was defined it was`#f`, which cannot be an `epair`). It's probably not worth it.

And the example above is not just theoretical... I've seen that a lot.

This would be possible:

1. The REPL receives a form and evaluates it, *and* it changes bindings to variables
2. The form immediately following it is `(get-changed-bindings-in-last-exp)`
3. STklos answers with a list of identifiers that were mutated in the form sent in step `1`.
4. The IDE now knows not the position, but at least which form caused the mutation/definition.And it could inform the user that it was "inside this S-exp here", and point to the beginning of that form.

This is theoretically possible, and I even thought,not too long ago, of an API for questioning the history of mutations to bindings (this would include mutations with `set!` and similar; definitions with `define`; and the history would contain calls to continuations, just so the information is complete). But I gave up. 

Also, another interesting idea:

1. The REPL receives a form and evaluates it, *and* it changes bindings to variables
2. The form immediately following it is `(set-line-number)`
3. The next form is `(set-file-name)`
3. STklos records that information in the `econs` of any procedures, lambdas (or even lists) that were defined in the S-expression of (1).
4. The IDE now can ask not the position, but at least which form caused the mutation/definition.And it could inform the user that it was "inside this S-exp here", and point to the beginning of that form.

I'm not sure wether these two ideas are worth pursuing. What do you think?

** Important: some people are trying to come up with a new SRFI that would standardize communication between Scheme and IDE, and this would probably cover some of what we said (this is in the [srfi-discuss](https://srfi-email.schemers.org/srfi-discuss/) mailing list)

jpellegrini:
And even after what I said... If I am editing the file, line numbers vary all the time. Maybe we could do only the formals for now?

egallesio:
Yep, having the formals would be a good start.

jpellegrini:
Ok... I pushed a new version:

* uses a new opcode
* the parameter shouldbe initialized to #t only in debug mode

However, it seems that even if I start stklos with `-d`, the parameter still initializes with #f... This is the code I used:

```
(define compiler:generate-signature             (make-parameter
                                                 (positive? (stklos-debug-level))))
```
in the module `STKOS-COMPILRER`.

When I try ` (positive? (stklos-debug-level))` at the REPL it correctly says `#t`. But 

` ((symbol-value 'compiler:generate-signature (find-module 'STKLOS-COMPILER)))` is  `#f` (why?)

egallesio:
I have corrected the problem with the initialization of `compiler:generate-signature` (it cannot be initialized in the compiler, since the code located here constitutes the boot image which is loaded when STklos starts, it reflect then the value of the debug variable of the compiler which build the image, not the one that the user runs). So the code which test the debug level is now in `boot.stk`.

BTW, I have seen that the was support for line numbers which was already present (I didn't remember that :frowning_face: ). However, it was not fully functional and I have modified it so that backtraces are more useful in debug mode. 

I have also modified a little bit the help function to add some color. I have also suppressed the two calls to `eval` that were in your code. I think that they ere useless (and since *eval is evil*, I prefer to avoid it here). Tell me if I'm wrong.

With signatures and line numbers (even if of limited use when used from an IDE as we discussed) , I think that we have a better environment. 

Time to freeze things, and release a 1.50  version, even if not all the Issues and PR are purged.

PR #93, opened 2020-08-03
Small enhancements to hygienic macro code

jpellegrini:
* When using `gensym`, pass it the symbol, so the generated symbols will be a bit easier to understand when debugging

* Add documentation to some procedures. Until another macro system is in place, we could enhance the documentation of this one a little bit. :)


egallesio:
Good idea. Merged.

PR #94, opened 2020-08-04
Add SRFI-5

jpellegrini:
Since STklos compiles `LET` (and does not define it as a macro),
the changes are on compiler.stk only (besides the usual changes in
srfi-0.stk, test-srfi.stk and documentation)

The SRFI has no accompaining test suite, so I have written a couple
of tests.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -20,6 +20,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-1: List Library
     - SRFI-2: AND-LET*: an AND with local bindings, a guarded LET* special form
     - SRFI-4: Homogeneous numeric vector datatypes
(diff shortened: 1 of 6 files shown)


jpellegrini:
This one changes the compiler, but in a not very intrusive way -- it's not a complex change...

jpellegrini:
Hey @egallesio !
I see that SRFI-5 is now implemented! And in a cleaner way (no need to change the compiler) -- that's great news!
No need for this PR anymore then...

egallesio:
Yep Jeronimo, I wanted to tell you that it is no more necessary to change the compiler for implementing SRFI-5  but you were faster than me!!!
The rewrite of the function `compile` should permit to ease the support of internal macros, and a side effect is that we can now redefine special forms without  modifying the compiler. 

BTW, I have not advanced a lot on local macros. Some initial implementation choice complicate things :-(

PR #95, opened 2020-08-04
Mention IDEs in README

jpellegrini:


egallesio:
Thanks, @jpellegrini . It is merged.

PR #98, opened 2020-08-07
SRFI 189 -- Maybe and Either: optional container types

jpellegrini:
Now that raise-continuable works, here's one more SRFI! :)


egallesio:
Damn, you are fast! I'll look at it tomorrow.

egallesio:
Merged!

PR #99, opened 2020-08-07
Add git info to SRFI-176 output

jpellegrini:
As per @lassik 's request, this patch will add git info to `version-alist`.

One thing: there seems to also be a `%git-push` procedure, but at least on my system it doesn't seem to be update during builds.

@egallesio if you want I may change the `%git-push` procedure to use the information from `GIT_SUMMARY`, which this PR introduces. Or  maybe there is a better way to handle the git info? This is the best I could do, after going through a sea of autotools-related information. :)

One more thing (@lassik I think you'll be interested) -- I just noticed that, if the user did not fetch the tags, the tag shown by git is outdated. I only realized this  when I ran `stklos -V` and saw it was showing `1.31` as tag... So I suppose there is no way to reliably know if the tag being shown is really the tag for the code version being used?


lassik:
Works perfectly and looks awesome! Really clean way to solve such a messy problem. Thanks you for this, I'm happy :)

lassik:
> One more thing (@lassik I think you'll be interested) -- I just noticed that, if the user did not fetch the tags, the tag shown by git is outdated. I only realized this when I ran `stklos -V` and saw it was showing `1.31` as tag... So I suppose there is no way to reliably know if the tag being shown is really the tag for the code version being used?

Excellent point. Unfortunately there is no way to solve this in general, since new tags can be added to old commits at any time. So there's no way to prove that a remote hasn't added a tag without fetching from that remote. And accessing the internet from a Makefile should generally be avoided, except in special targets.

lassik:
When `git describe` shows a tag, it also shows how many commits have been made after that tag. I guess that solves part of the problem: if it shows 100 more commits than you'd expect, it means that the tags must be out of date. However, in this PR you show the commit hash, so if there is a weird situation the hash can be looked up in `git log` to figure out the situation.

egallesio:
Thanks a lot for this new contribution.

However, I have changed it a bit to enhance some points. The approach  used with generating a new `git-config.h` when running `make` using the special `BUILT_SOURCE` provokes a new link of STklos each time a `make` command is run. For instance, after an initial successful `make`, running a `make install` provokes two new calls (why two, I don't know) to the loader (and generally in `sudo` mode, it will create a file belonging to root in the user directory). It implies also  that the file that is installed is not exactly the one which has been tested (normally not a problem since we can suppose that the C compiler produces the same file).

I have also modified the `lib/Makefile.am` to update the `git-config.h` if there are modifications in the lib directory. Otherwise, if the user implementation resides only in the lib directory, the information given by the `-V` option is false. Normally, we should do the same thing in all the directories (for instance, modifying only the documentation change the git index, but should not be a source of bugs). I suppose that this is not necessary to be so strict.

Since a new link is not necessary if the newly generated `git-config.h` is identical to the previous one, the new version of the generation script will produce a new version in a temporary file and if a change is detected, the new version replaces the old one. Hence, `git-config.h` is now a standard dependence of `misc.c` and a new link is done iff  there are changes in the information that should be displayed by the `-V` option.

I will push merge this  PR, and add my mods asap (I still need to change the ID which is displayed in the STklos banner to be the commit hash instaed in the `GIT_SUMMARY` macro, to be complete).

lassik:
Great work!

> I have also modified the `lib/Makefile.am` to update the `git-config.h` if there are modifications in the lib directory. Otherwise, if the user implementation resides only in the lib directory, the information given by the `-V` option is false. Normally, we should do the same thing in all the directories (for instance, modifying only the documentation change the git index, but should not be a source of bugs). I suppose that this is not necessary to be so strict.

It's a matter of taste. Usually it's not expected to be so strict about it - it doesn't matter if a documentation change also updates the commit information in the executable. Good thinking - I hadn't even thought about this issue :)

> Since a new link is not necessary if the newly generated `git-config.h` is identical to the previous one, the new version of the generation script will produce a new version in a temporary file and if a change is detected, the new version replaces the old one. Hence, `git-config.h` is now a standard dependence of `misc.c` and a new link is done iff there are changes in the information that should be displayed by the `-V` option.

Excellent idea. Writing first to a tempfile solves it in a clean way.

> I will push merge this PR, and add my mods asap (I still need to change the ID which is displayed in the STklos banner to be the commit hash instaed in the `GIT_SUMMARY` macro, to be complete).

The commit hash in the STklos startup banner is a very good idea. Gambit also has one now. It makes it a lot easier to notice when you are running a development version vs. a release version. Especially helps if logging into different computers (via ssh) which have different versions, or a release version in `/usr/local` and development in `$HOME`, etc.

PR #101, opened 2020-08-13
Remove double semicolon from C files

jpellegrini:
@egallesio I guess you got used to press ";" twice after so much Lisp coding... :)


egallesio:
That's crazy! Thanks.

PR #103, opened 2020-08-15
Add SRFI-61 - A more general cond clause

jpellegrini:
Just one extra condition in the compilation of COND gives SRFI-61... :)

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -52,6 +52,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-55: Require-extension
     - SRFI-59: Vicinity
     - SRFI-60: Integers as bits
+    - SRFI-61: A more general COND clause
     - SRFI-62: S-expression comments
     - SRFI-64: A Scheme API for test suites
     - SRFI-66: Octet Vectors
--- a/doc/skb/srfi.skb
+++ b/doc/skb/srfi.skb
@@ -517,6 +517,11 @@ you need to insert the following expression])
 (p [,(quick-link-srfi 62) is fully supported. See ,(ref :subsection "Comments")]))
 
 ;; ----------------------------------------------------------------------
+;; SRFI 61 -- A more general COND clause
+;; ----------------------------------------------------------------------
+(gen-loaded-srfi 61)
+
+;; ----------------------------------------------------------------------
 ;; SRFI 64 -- A Scheme API for test suites
 ;; ----------------------------------------------------------------------
 (srfi-section 64
--- a/doc/skb/srfi.stk
+++ b/doc/skb/srfi.stk
@@ -63,6 +63,7 @@
      (55 . "Require-extension")
      (59 . "Vicinity")
      (60 . "Integers as bits")
+     (61 . "A more general COND clause")
      (62 . "S-expression comments")
      (64 . "A Scheme API for test suites")
      (66 . "Octet Vectors")
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -99,6 +99,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-54.stk         \
           srfi-59.stk         \
           srfi-60.stk         \
+          srfi-61.stk         \
           srfi-64.stk         \
           srfi-66.stk         \
           srfi-69.stk         \
@@ -163,6 +164,7 @@ scheme_OBJS = compfile.ostk    \
           srfi-54.ostk         \
           srfi-59.ostk         \
           srfi-60.ostk         \
+          srfi-61.ostk         \
           srfi-64.ostk         \
           srfi-66.ostk         \
           srfi-69.ostk         \
--- a/lib/srfi-0.stk
+++ b/lib/srfi-0.stk
@@ -114,9 +114,9 @@
     ;; srfi-56                          ; ....... withdrawn
     ;; srfi-57                          ; Records
     ;; srfi-58                          ; Array Notation
-    (srfi-59 "srfi-59")                         ; Vicinity
+    (srfi-59 "srfi-59")                 ; Vicinity
     (srfi-60 "srfi-60")                 ; Integers as Bits
-    ;; srfi-61                          ; A more general cond clause
+    (srfi-61 "srfi-61")                 ; A more general cond clause
     srfi-62                             ; S-expression comments
     ;; srfi-63                          ; Homogeneous and Heterogeneous Arrays
     ((srfi-64 testing) "srfi-64")       ; A Scheme API for test suites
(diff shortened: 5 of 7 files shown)


jpellegrini:
I suppose it would be better to have this also as a macro instead of a patch to the compiler (as with SRFI-5), right?

egallesio:
> I suppose it would be better to have this also as a macro instead of a patch to the compiler (as with SRFI-5), right?

Yes this was the idea. It is probably safer to avoid touching the compiler, if we can.

jpellegrini:
Also I suppose you'd prefer `define-macro` over `syntax-rules`, since speed for `COND` is essential...

jpellegrini:
I'm not sure why this isn't working for `COND`. I have tried patching the procedure `rewrite-cond-clauses` from the compiler and make a new `cond`, but...


```
(define (rewrite-cond-clauses  c)
  (cond
   ((null? c)                 (void))
   ((not (pair? (car c)))     (compiler-error 'cond c "invalid clause ~S" (car c)))
   ((eq? (caar c) 'else)      (if (null? (cdr c))
                                  `(begin ,@(cdar c))
                                  (compiler-error 'cond
                                                   c
                                                   "else not in last clause ~S"
                                                   c)))
   ((and (pair? (cdar c)) (eq? (cadar c) '=>))
                              (if (and (list? (car c)) (= (length (car c)) 3))
                                  (let ((test-var (gensym)))
                                    `(let ((,test-var ,(caar c)))
                                       (if ,test-var
                                           (,(caddar c) ,test-var)
                                           ,(rewrite-cond-clauses (cdr c)))))
                                  (compiler-error 'cond
                                                   c
                                                   "bad '=>' clause syntax ~S"
                                                   (car c))))
   ((and (pair? (cdar c)) (pair? (cddar c)) (eq? (caddar c) '=>))
                              (if (and (list? (car c)) (= (length (car c)) 4))
                                  (let ((test-var (gensym)))
                                    `(let ((,test-var ,(caar c)))
                                       (if (and ,test-var
                                                (,(cadar c) ,test-var))
                                           (,(car (cdddar c)) ,test-var)
                                           ,(rewrite-cond-clauses (cdr c)))))
                                  (compiler-error 'cond
                                                   c
                                                   "bad '=>' clause syntax ~S"
                                                   (car c))))
   ((null? (cdar c))          (let ((test-var (gensym)))
                                `(let ((,test-var ,(caar c)))
                                   (or ,test-var
                                       ,(rewrite-cond-clauses (cdr c))))))
   (else                      `(if ,(caar c)
                                   (begin ,@(cdar c))
                                   ,(rewrite-cond-clauses (cdr c))))))
```

Define both `cnd` and `cond`:

```
(define-macro (cnd . expr)
  (rewrite-cond-clauses expr))

(define-macro (cond . expr)
  (rewrite-cond-clauses expr))
```

`cnd` works, but `cond` doesn't!

```
stklos> (define (get-nat) 10)

stklos> (define (get-sym) 'x)

stklos> (cond ((get-nat) integer? => (lambda (x) (- x))) ((get-sym) 'y))
**** Error:
%execute: variable `=>' unbound
stklos> (cnd ((get-nat) integer? => (lambda (x) (- x))) ((get-sym) 'y))
-10
```

Even though their macro-expansions are identical:

```
stklos> (macro-expand '(cnd ((get-nat) integer? => (lambda (x) (- x))) ((get-sym) 'y)))
(let ((G265 (get-nat))) (if (and G265 (integer? G265)) ((lambda (x) (- x)) G265) (if (get-sym) (begin 'y) #void)))
stklos> (macro-expand '(cond ((get-nat) integer? => (lambda (x) (- x))) ((get-sym) 'y)))
(let ((G267 (get-nat))) (if (and G267 (integer? G267)) ((lambda (x) (- x)) G267) (if (get-sym) (begin 'y) #void)))
```

jpellegrini:
@egallesio I think it's done!

Loaded, not embedded; 
Using `macro-expand`, so it won't be slow;
Based on the ooriginal `COND` code in the compiler, so should not introduce any strange behavior;
Also, it only adds one clause to the expansion cases, so nothing else should be affected.

Seems to pass my tests and the STklos R5RS and R7RS tests...

egallesio:
Hi @jpellegrini ,


> Also I suppose you'd prefer define-macro over syntax-rules, since speed for COND is essential...

I don't see what you mean here. In fact the expanded code should be identical in both cases.

Given the sample implementation in the SRFI (code case-lowered scheme)

```here
(define-syntax cond
  (syntax-rules (=> else)

    ((cond (else else1 else2 ...))
     (if #t (begin else1 else2 ...)))

    ((cond (test => receiver) more-clause ...)
     (let ((t test))
       (cond/maybe-more t
                        (receiver t)
                        more-clause ...)))

    ((cond (generator guard => receiver) more-clause ...)
     (call-with-values (lambda () generator)
       (lambda t
         (cond/maybe-more (apply guard    t)
                          (apply receiver t)
                          more-clause ...))))

    ((cond (test) more-clause ...)
     (let ((t test))
       (cond/maybe-more t t more-clause ...)))

    ((cond (test body1 body2 ...) more-clause ...)
     (cond/maybe-more test
                      (begin body1 body2 ...)
                      more-clause ...))))

(define-syntax cond/maybe-more
  (syntax-rules ()
    ((cond/maybe-more test consequent)
     (if test
         consequent))
    ((cond/maybe-more test consequent clause ...)
     (if test
         consequent
         (cond clause ...)))))
```

We obtain:

```scheme
stklos> (pp (macro-expand* '(cond
                             ((integer? x) 'int)
                             ((number? x)  'number)
                             ((assoc x '((1 . 2) ( 3 . 4))) => cdr)
                             (else "???"))))

(if (integer? x)
  (begin 'int)
  (if (number? x)
    (begin 'number)
    (let ((t236 (assoc x '((1 . 2) (3 . 4)))))
      (if t236 (cdr t236) (if #t (begin "???"))))))

```

which is more or less the code that we would obtain  with the function `rewrite-cnd-clauses` of your macro.
Furthermore, `rewrite-cnd-clauses` uses `compiler-error` which is not exported by the compiler :disappointed: 

So, if that doesn't bother you, I can merge your PR and immediately replace the code in `lib/srfi-61.stk` by the SRFI sample implementation.  I have this already ready in a local branch.

jpellegrini:
> I don't see what you mean here. In fact the expanded code should be identical in both cases.

I meant compilation time -- COND is very common, so the syntax-rules machinery could make compilation slower -- but it's probably not a problem!

> So, if that doesn't bother you, I can merge your PR and immediately replace the code in `lib/srfi-61.stk` by the SRFI sample implementation. I have this already ready in a local branch.

Sure, no problem!

jpellegrini:
>  Furthermore, `rewrite-cnd-clauses` uses `compiler-error` which is not exported by the compiler disappointed

Oh, my mistake. I haven't checked that...

PR #105, opened 2020-08-18
Some more updates on the VM documentation

jpellegrini:
The `DEFINE_SYMBOL` opcode seems to only be used in top level, so it's impossible to make an example with`disassemble`, right?

### DIFF
--- a/vm.md
+++ b/vm.md
@@ -580,10 +580,12 @@ NOP
 ```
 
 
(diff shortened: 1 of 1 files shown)


jpellegrini:
I just included another update, using the new `disassemble-expr` - and an example of `define` disassembled.

PR #106, opened 2020-08-25
Add %%local-alias

jpellegrini:
@egallesio  this is a work in progress. I am including it here just so you can see it, but I suppose its not yet time to accept it.

This is the idea I was talking about in issue #55 .
It was actually not complex at all. With `%%local-alias`, it would be possible to implement hygienic macros in a clean and not complex way.

`(%%local-alias b a)` defines `b` as an alias to the `a` variable _visible where the `%%local-alias` is used_. It does so by pushing one more item onto the `env` stack: instead of a variable name (symbol), it pushes a list (the alias). The rest of the code is adapted to cope with this.


jpellegrini:
For full compliance with the standards, it would also be necessary to have a similar `%syntax-alias`.

```
> (define-syntax let (syntax-rules () ((let a b c) (+ a b c))))
> (let 2 3 4)
9
```

(All the issues with variables arise with syntax too, but the solution then is very similar. Working on it)

jpellegrini:
This PR needs a complete rewrite -- I'll do that later when we have lexically-scoped define-macro

lassik:
Please note that [SRFI 212: Aliases](https://srfi.schemers.org/srfi-212/srfi-212.html) is currently in draft status. Possibly related, or STklos might have some ideas that are good to add to that SRFI.

jpellegrini:
> Please note that [SRFI 212: Aliases](https://srfi.schemers.org/srfi-212/srfi-212.html) is currently in draft status. Possibly related, or STklos might have some ideas that are good to add to that SRFI.

Yes - but it would make more sense to implement that after the work on lexically-scoped macros is done, because both variables and macros could have their names aliased (indeed, my idea was to use both in order to implement R7RS semantics for syntax-rules).

PR #109, opened 2020-08-30
Add SRFI-113 - Sets and Bags

jpellegrini:
Some tests were not included because they use test-assert to detect syntax errors. Other than that, all tests pass and the implementation seems to work perfectly.


egallesio:
And one more!!! 
Thanks

egallesio:
Hi @jpellegrini 

Tests 101 and 102 don't pass. It seems to be again a problem with the hash function. 
On my machine:
```
Errors found in this section:
test 101 on (= (comparator-hash default-comparator #\t) (char-hash #\t)) expected #t but got #f
test 102 on (begin (let ((a (comparator-hash default-comparator "t")) (b (string-hash "t"))) (print "a= " a "\nb= " b "\n") (= a b))) expected #t but got #f
```

jpellegrini:
This is really strange. All tests passed when I made the PR, now they don't. And even stranger: what I get here is not the errors you mentioned. It's this:
```
==== Testing SRFIs  ...                                          **** Error while executing file "do-test.stk"
	 Where: in try-load
	Reason: variable `srfi-37:option' unbound

  - <<let/call>>
  - try-load
  - %try-load
  - %load
  - #[closure 7f7258051f00]
  - %execute
  - #[closure 7f723bdb0230]
  - call-with-values
  - dynamic-wind
  - try-load
  - ...
Set shell variable STKLOS_FRAMES to set visible frames
EXIT
make[1]: *** [Makefile:459: test] Error 70
```
I'll take a look

jpellegrini:
The tests that don't pass are from the comparators SRFI, because the SRFI-113 tests redefine some things from SRFI-128. I'm working on this

jpellegrini:
I am confused. I got all tests to pass again,but:

* In SRFI 37 there was a record named "srfi-37:option" when it should be "option".
* In the comparators procedure (128) there was a print, which clutters the tests. But I had never seen those before!
* I removed the string-hash and char-hash procedures from the SRFI-113 tests, they're not necessary. But I wonder why the tests passed before here.

Anyway - Ind a new PR.

PR #110, opened 2020-08-30
Use  $(DESTDIR) in install-sources target

jpellegrini:
Otherwise installation on alternative locations won't work...

See in `Makefile.in` in other directories, for example: it will expand into `$(DESTDIR)/$(schemedir)`. In the `lib/` directory, since the Make target was written by hand, the `$(DESTDIR)` part is missing.


egallesio:
Included in commit  80e4ef2a731b84274369d64c1d768d6dca9f05b5

PR #112, opened 2020-09-04
Use an env var for strip

jpellegrini:
When cross-compiling, the strip binary will not work on the stklos
executable (binary format mismatch). One solution is to use a STRIP
variable, and set it to "stklos" if it is either missing or empty.

Tested on Debian GNU/Linux on amd64, and on the OpenWRT build root
(for MIPS).


jpellegrini:
Hi @egallesio -- I have changed the PR so it preserves the `install-base-no-strip` target, which may be useful perhaps, to those willing to debug STklos.

egallesio:
Merged!

PR #114, opened 2020-09-14
Make clean really clean in src/

jpellegrini:
* Remove git-info.h, not git-config.h
* Also remove boot.ok
 * And fix the comment in Makefile.am

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -40,7 +40,7 @@ stklos_SOURCES = base64.c boolean.c boot.c box.c char.c \
 
 # HACK:  Normally BUILT_SOURCES contains the files that must be generated before
 # starting the construction of the main target. We use it with a fictive target
-# which runs a script which may update the git-config.h file. If it is updated, 
+# which runs a script which may update the git-info.h file. If it is updated, 
 # normal rules apply and a new global link is done.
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @jpellegrini,

Imho, `boot.ok` should not be deleted. It is the last boot file which is know to work. This is particulary useful if you have a bug in `boot.img`.  Restarting from a too mush old image can also be a problem, and force you to unset a bunch of modification you have done in one of the intervening files in the boot process. It served me more than one time, trust me. 

Of course, committing often in the git tree minimizes the utility to keep it, but it is a good safety belt to prevent long rollbacks. Is it really a problem to keep it?

jpellegrini:
Hm, ok. Actually, the Debian package expects the clean target to delete it, but this is not a problem -- it can be done ad-hoc in `debian/rules`.

egallesio:
Hi,
I have renamed the file `boot.ok` in `boot.keep`. However, instead of deleting is now done when doing a `make distclean` (instead of a `make clean` as done in this PR). I don't know if will be OK for Debian, but as I said before, it is important to keep this file to be able to backtrack when the boot image is broken.  I suppose you'll have to modify the `debian/rules` file.

PR #116, opened 2020-09-14
Update SRFIs information in srfi-0.stk

jpellegrini:
Some were withdrawn, some went final, and some were renamed!


egallesio:
Merged!

Issue #115, opened 2020-09-14
.so files should not be in /usr/share


jpellegrini:
The FHS determines that shared objects should be in `$(prefix)/lib`, and not `$(prefix)/share`; and the Debian packaging system enforces that, so it would be necessary to move `srfi-175-impl.so` to `$(prefix)/lib`. I have tried, but I was once again eluded by autotools! I can't see how it's supposed to be done.

I thought of using something like `$(prefix)/lib/stklos/$(STKLOS_VERSION)`, and configuring STklos to look for C libraries there. But I can't get `lib/Makefile.am` to install the `.so` file in a different directory.

**EDIT:** I think I got it right. Please take a look at PR #117 .

jpellegrini:
This doesn't really fix it, although I managed to `(require "srfi-175")`. It doesn't seem to pass the tests. @egallesio how do we make STklos look for libraries in `$PREFIX/lib/stklos/1.50`?

egallesio:
Your PR #117 effectively corrects this point. 
Since it has been merge, I close this issue.
Thanks.

PR #117, opened 2020-09-14
Install .so files in lib/

jpellegrini:
I am not sure if this is correct, but it seems to work!
This addresses bug #115 which has been brought up when packaging for Debian.


jpellegrini:
This doesn't really fix it, although I managed to `(require "srfi-175")`. It doesn't seem to pass the tests. @egallesio how do we make STklos look for libraries in `$PREFIX/lib/stklos/1.50`?

jpellegrini:
Hm, it only doesn't work when compiling the Debian package. I'll investigate

PR #120, opened 2020-09-16
Add manpage to stklos-script

jpellegrini:
(Debian packages want manpages for all executables)


egallesio:
Thanks for this page. Merged!

PR #121, opened 2020-09-16
Change encoding to UTF-8

jpellegrini:
One more issue detected by the Debian packaging tools... Two files use ISO-8859-1 instead of UTF-8. This PR converts them to UTF-8.


egallesio:
Thanks. It is merged

Issue #119, opened 2020-09-16
Clarify copyright of some files


jpellegrini:
Hello @egallesio !

I was checking the licenses in order to put them in debian/copyright, and it isn't clear from some files what their licenses are.

Some of them seem to have an MIT-like license (like `process.c` for example). Is that correct, or should it be GPLv2 as most of STklos?

Anyway, these are the files:

```
src/base64.c: UNKNOWN
src/process.c: UNKNOWN
src/system.c: UNKNOWN


lib/bigloo-support.stk: UNKNOWN
lib/bigmatch.stk: UNKNOWN
lib/expand.pp: *No copyright* UNKNOWN
lib/expand.ss: UNKNOWN
lib/full-syntax.stk: UNKNOWN
lib/mbe.stk: *No copyright* UNKNOWN

lib/srfi-1.stk: UNKNOWN                              (SRFI license; what about the adaptation?)
lib/srfi-100.stk: UNKNOWN
lib/srfi-2.stk: *No copyright* UNKNOWN
lib/srfi-45.stk: UNKNOWN
lib/srfi-59.stk: *No copyright* UNKNOWN
lib/srfi-74.stk: Expat License
lib/srfi-89.stk: UNKNOWN
lib/srfi-9.stk: UNKNOWN

pkgman/lang-bigloo.l: *No copyright* UNKNOWN
pkgman/lang-chicken.l: *No copyright* UNKNOWN
pkgman/lang-mzscheme.l: *No copyright* UNKNOWN

etc/stk-dynload.c: UNKNOWN
```

What about the Makefiles? Can I state they all have the same license as the code?

egallesio:
Hi, 
For most of the files I'm the only author, so it is simple (files with MIT licences are files that comes from STk, since it was the licence it used). For some file it will be a little bit more difficult, but all the files used in STklos have a license which is compatible with GPL2 (as far as I understand licences terms subtleties).  

I clarify things in the files themselves before answering more precisely to this issue. 
Debian tools seems to be terrific for tracking all the  corners of a package. I'm impressed. 
Thanks for signalling those points and for your help with the Debian packaging.

egallesio:
Hi I have pushed some commits to clarify the copyrights.
What we have:
Deleted files

 - etc/stk-dynload.c: UNKNOWN ==> deleted. This file was useless.
 - lib/srfi-45.stk: UNKNOWN ==> the file has been deleted and this SRFI is now embedded. 

[...]

Issue #118, opened 2020-09-16
What is the license of the documentation?


jpellegrini:
Hello!
What is the license of STklos' documentation? (the autogenerated docs, the manpages, README, etc)
The Debian packaging team wants to know. :)

egallesio:
Hi!

That is an excellent question. For the parts I wrote I don't really care (GFGL?) but large parts of it come from R5RS or R7RS, where the copyright is not really clear....

jpellegrini:
Hi @egallesio -- did you mean GFDL? Would that be 1.3 "or later"?
I'll check the RnRS license status. I am sure this have been sorted out before, because other Schemes have been packaged for Debian, and I think lots of them do include parts of RnRS.

egallesio:
Yep GFDL (or any other license). 
If GFDL,  don't know the diffrence between versions.

lassik:
> I'll check the RnRS license status. I am sure this have been sorted out before, because other Schemes have been packaged for Debian, and I think lots of them do include parts of RnRS.

From [the R5RS PDF](https://www.schemers.org/Documents/Standards/R5RS/r5rs.pdf):

> We intend this report to belong to the entire Scheme community, and so we grant permission to copy it in whole or in part without fee. In particular, we encourage implementors of Scheme to use this report as a starting point for manuals and other documentation, modifying it as necessary.

lassik:
Not sure if there is a more formal license somewhere. The TeX files are collected here: https://github.com/schemedoc/rnrs-metadata There may be other files in the original tar archives that contain those TeX files that are not included in that repo.

jpellegrini:
Hi @lassik . Do you think the statement (we grant permission to copy it in whole or in part without fee) is clear enough? Can it be relicensed under a different license? And if, for example, the STklos manual mixes R5RS with @egallesio 's own text, is it necessary to identify which part is taken from R5RS?

(The same applies to R7RS -- which has the same statement)

lassik:
We'd best ask @weinholt about that. I'm not well versed in the details of copyright law or Debian policy.

weinholt:
Whenever this comes up on a Scheme list the answer is something to the effect that you just need to read the paragraph, it says everything you need to know:

> We intend this report to belong to the entire Scheme community, and so we grant permission to copy it in whole or in part without fee. In particular, we encourage implementors of Scheme to use this report as a starting point for manuals and other documentation, modifying it as necessary.

[...]

PR #124, opened 2020-09-18
Licensing overhaul

jpellegrini:
HI,

@egallesio since you mentioned in issue #118 that GFDL would be OK I added a copy of it.

I'm not sure if this is good, but since there is code licenses and documentation license, I modified COPYING to only mention authorship and license, referring to the license files from there. 
The diff shows a change in the GPL-2 file, but that is only because I put a standard copy of it, and git recorded it as a modification to COPYING.

I'm not sure you agree -- this is really only a proposal!

(Identifying the licenses of some files, as mentioned in issue #119 would still be pending, but this PR would perhaps be good anyway?)

The changes:

* Only mention copyright in COPYING, referring to the license files
  from there. Also mention that some files have different license and
  authors.

* ADD GPL-2 and GFDL-1.3.

* Mention copyright in README.md.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,1 +1,7 @@
-LICENSE
+
+STklos: Copyright (C) 1993-2020 Erick Gallesio. Most STklos source files are distributed under the GNU General Public License version 2 or any later version (SPDX GPL-2.0-or-later); some are distributed under different licenses and have been adapted from works by other authors -- please see each file for its specific license.
+
+STklos documentation: Copyright (C) 1993-2020 Erick Gallesio. The documentation is distributed under the GNU Free Documentation License version 1.3 or any later version (SPDX GFDL-1.3-or-later).
+
+HACKING amd vm.md: Copyright (C) 2020 Jerônimo Pellegrini. These documents are distributed under the GNU Free Documentation License version 1.3 or any later version (SPDX GFDL-1.3-or-later).
+
--- a/GFDL-1.3
+++ b/GFDL-1.3
@@ -1,0 +1,451 @@
+
+                GNU Free Documentation License
+                 Version 1.3, 3 November 2008
+
+
+ Copyright (C) 2000, 2001, 2002, 2007, 2008 Free Software Foundation, Inc.
+     <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+0. PREAMBLE
+
+The purpose of this License is to make a manual, textbook, or other
+functional and useful document "free" in the sense of freedom: to
+assure everyone the effective freedom to copy and redistribute it,
+with or without modifying it, either commercially or noncommercially.
+Secondarily, this License preserves for the author and publisher a way
+to get credit for their work, while not being considered responsible
+for modifications made by others.
+
+This License is a kind of "copyleft", which means that derivative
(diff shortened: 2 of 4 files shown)


lassik:
The https://reuse.software/spec/ specification recommends a `LICENSES/` directory in the repo, with the license files named like:

* `LICENSES/GPL-2.0-or-later.txt`
* `LICENSES/GFDL-1.3-or-later.txt`

@weinholt is using this layout in the Scheme package manager [Akku](https://gitlab.com/akkuscm/akku). Do Debian tools support this?

weinholt:
@lassik Are you asking me? I don't know what such support would look like, but maybe one of the tools at <https://wiki.debian.org/CopyrightReviewTools> does something special with it. We usually try to figure out upstream's licenses ourselves, which is always a manual job (though possibly supported by tools). Sometimes it can be like detective work, but upstream is rarely directly involved.

jpellegrini:
@egallesio is it OK for you to have a `LICENSES/` directory, as @lassik mentioned?

Anyway -- if a file was adapted, then it would be, for example, "orginally MIT with GPL modifications by Erick Gallesio", for example. It should *not* be considered as an MIT-licensed file, but rather GPL, as I understand. However, the text of the MIT license should stay there, because it *is* the original license, and was "sublicensed" as GPL. 
So I suppose it would be nice to have the text of all licenses in that directory, and have the top-level `COPYING` file explain that the whole thing was relicensed as GPL. Is this reasonable?

egallesio:
> @egallesio is it OK for you to have a LICENSES/ directory, as @lassik mentioned?

Yes, this is the best way to avoid cluttering the main directory. 

> So I suppose it would be nice to have the text of all licenses in that directory, and have the top-level COPYING file explain that the whole thing was relicensed as GPL. Is this reasonable?

I'm not a lawyer, but it seems OK for me.

lassik:
[SPDX](https://spdx.dev) is an organization that is trying to standardize how license information is marked in open source projects, and their conventions are now being used by the Linux kernel et.al. The information is marked in a source code comment at the top of each file. [Here is the explanation from Linux](https://github.com/torvalds/linux/blob/8c79f4cd441b27df6cadd11b70a50e06b3b3a2bf/Documentation/process/license-rules.rst#license-identifier-syntax). What do you think about using these `SPDX-License-Identifier:` comments in the STklos source?

lassik:
The more complex comments would look something like this: `/* SPDX-License-Identifier: MIT AND GPL-2.0-or-later */`

Each well-known license is referred to using a standard identifier; [here is the list of them all](https://spdx.org/licenses/).

lassik:
Some projects are developing tools to auto-extract these comments to figure out the license of a project, including the above REUSE initiative.

jpellegrini:
@egallesio -- I remember some of my adaptations of SRFIs were released under GPLv3-or-later. Now I realize one thing: if I understand correctly, that means any redistribution of STklos with them would also need to be GPLv3-or-later, so if you want, I may change those to GPLv2-or-later (the same as STklos currently).
Or if you want to change the whole code base to GPLv3-or-later, I'll be happy too :)  -- it's your choice of course!

jpellegrini:
Also -- @egallesio -- what do you think of @lassik 's idea? Can we add SPDX headers to the source files? I can make a PR if it's OK with you.

Issue #122, opened 2020-09-16
.so files do not have prerequisite information


jpellegrini:
No big deal... One more thing from Debian lint:

```
W: stklos: shared-library-lacks-prerequisites usr/lib/stklos/1.50/srfi-175-impl.so
```

The warning tag means:

> _The listed shared library doesn't include information about the other libraries against which it was linked.
> More specifically, "ldd foo.so" should report such other libraries. In your case, it reports "statically linked".
> The fix is to specify the libraries. One way to do so is to add something like "-lc" to the command-line options for "ld"._

So I did include `-lc` in `lib/Makefile.am`, 

```
.c.@SH_SUFFIX@ :
	@CC@ @CFLAGS@ @STKCFLAGS@ @SH_COMP_FLAGS@ -I../src @GCINC@ \
	-c -o $*.o $*.c
	@SH_LOADER@ @SH_LOAD_FLAGS@  $*.@SH_SUFFIX@ -lc $*.o
	/bin/rm -f $*.o
```

But I'm not sure this is the right way to handle it. Should it somehow be included in `SH_LOAD_FLAGS`?
Or is hardcoding `-lc` there is OK? (If it is, I can make a PR)

jpellegrini:
PR #125 would include this.

PR #125, opened 2020-09-20
Add @CPPFLAGS@ and -lc to shared lib compiling

jpellegrini:
`@CPPFLAGS@` is important in hardening steps (Debian hardening testing tools complain -- this was an error reported by the CI tools)

The `-lc` part is to get dependency information. (This one was just a warning, not an error)

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -190,9 +190,9 @@ SUFFIXES = .stk .ostk .scm
 	$(COMP) -o $*.ostk $*.scm
 
 .c.@SH_SUFFIX@ :
-	@CC@ @CFLAGS@ @STKCFLAGS@ @SH_COMP_FLAGS@ -I../src @GCINC@ \
+	@CC@ @CPPFLAGS@ @CFLAGS@ @STKCFLAGS@ @SH_COMP_FLAGS@ -I../src @GCINC@ \
 	-c -o $*.o $*.c
-	@SH_LOADER@ @SH_LOAD_FLAGS@  $*.@SH_SUFFIX@ $*.o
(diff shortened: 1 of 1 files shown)


jpellegrini:
This would resolve issue #122

jpellegrini:
@egallesio it seems that there may be a better way of handling this. Maybe wait a little?

egallesio:
I'll wait ....
BTW, I have changed the ways CFLAGS is used when removing automake warnings. Perhaps it has some influence on what you are doing (hope it is not the case).

jpellegrini:
As I mentioned in PR #195 , it seems that we need to also pass `-lgc`, and this fixes compiling/loading shared libs in Android.

egallesio:
> As I mentioned in PR #195 , it seems that we need to also pass -lgc, and this fixes compiling/loading shared libs in Android.

I'm trying to access my Android phone, but for some reason I cannot see it anymore. :disappointed:  !!! That's really weird since it worked until now. 
Anyway, I'm quite sure that adding it will cause problems on some systems, and since it can be statically loaded or not depending on the configure flags  (MacOs probably since it is always different from other systems).  As soon as I have access to my phone, I'll come back here. 

> But do we need to pass those explicitly? Autotools is such a huge tool, shouldn't it take care of this?

Probably, but it will not.. Autotool can do a lot of things, .... if you tell it how to do those things :wink:

egallesio:
Hi @jpellegrini ,

I close this PR, because I think it is corrected now. In fact, I add all the dynamic libraries used when compiling STklos to the `ld` line. For now, I only do that for Android.

PR #132, opened 2020-11-01
Fix SRFI-193 comment in srfi.skb

jpellegrini:


egallesio:
Ouch well seen @jpellegrini . Merged

PR #133, opened 2020-11-01
Export a procedure used by a macro in SRFI-64

jpellegrini:
Procedure %test-approximate= is used by macro test-approximate=,
but wasn't being exported


egallesio:
Merged!

PR #135, opened 2020-11-08
Fix list= in SRFI-1

jpellegrini:
I found this while preparing SRFI-134.

In `list=`, the reference implementation has a loop as this:

`(let lp2 ((pair-a list-a) (pair-b list-b))`

The STklos version had

`(let lp2 ((list-a list-a) (list-b list-b))`

but this makes the original list-b not accessible from inside the
loop, so for example

`(apply list= char-ci=? '((#\a #\b) (#\A #\B) (#\a #\B)))`

would fail when it should not.

This commit reverts to the reference implementation version of
`list=`.

### DIFF
--- a/lib/srfi-1.stk
+++ b/lib/srfi-1.stk
@@ -487,18 +487,18 @@
   (or (null? lists) ; special case
 
       (let lp1 ((list-a (car lists)) (others (cdr lists)))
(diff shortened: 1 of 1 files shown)


egallesio:
Thanks for the correction . Merged!

Issue #136, opened 2020-11-08
There are no tests for SRFI-1


jpellegrini:
The SRFI repository has no tests, but it should be easy, although boring, to port tests from some other Scheme implementation.

PR #140, opened 2020-11-11
Add tests for SRFI-1

jpellegrini:
Not complete yet. These were adapted from Larceny to Chicken,
and now to STklos.

The Guile test suite is more complete, but also longer and it uses a different test framework; it would be harder to port.

### DIFF
--- a/tests/test-srfi.stk
+++ b/tests/test-srfi.stk
@@ -30,6 +30,634 @@
 (test-section "SRFIs")
 
 ;; ----------------------------------------------------------------------
+;;  SRFI 1 ...
+;; ----------------------------------------------------------------------
+(test-subsection "SRFI 1 - List Library ...")
+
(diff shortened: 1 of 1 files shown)


jpellegrini:
The tests will fail unless PR #135 is applied.

jpellegrini:
This would fix #136

egallesio:
**Thanks a lot** @jpellegrini for this port. It is probably sufficient to have this test set and add some tests from Guile if needed.

Merged.

lassik:
https://github.com/srfi-explorations/srfi-test

jpellegrini:
> https://github.com/srfi-explorations/srfi-test

That is great @lassik !
I'll work on a converter from SRFI-64 to STklos tests, and then we'll have all those tests available!

Issue #141, opened 2020-11-12
SRFI-4: s64 vectors refuse to hold negative values


jpellegrini:
```
stklos> (s64vector -3)
**** Error:
%uvector: value `-3' is out of bounds or incorrect for a s64vector
```

```
stklos> (define a (s64vector 3))
stklos> (s64vector-set! a 0 -1)
**** Error:
%uvector-set!: value `-1' is out of bounds or incorrect for a s64vector
```

I'll look into this.

jpellegrini:
Also, f32 does not accept exact numbers:

```
stklos> (f32vector 0)
**** Error:
%uvector: value `0' is out of bounds or incorrect for a f32vector
	(type ",help" for more information)
stklos> (f32vector 1.0)
#f32(1.0)
```

jpellegrini:
Ok, found it!
`u64` and `s64` are swapped in `uvector.c`.
`u64vector`s accept negatives, and `s64vector`s don't -- it should be the other way around.

PR #142, opened 2020-11-12
Add tests for SRFI-4

jpellegrini:
The only failing test is related to #141 .

Tell me if it's Ok. We should probably delete the tests that are commented out.


egallesio:
OK. I merge and will uncomment the tests now that you have fixed the problem with signed ans unsigned vectors. 
Thanks again for all your work.

PR #144, opened 2020-11-12
Fix u64/s64 vectors mixup

### LINKED ISSUE #141: SRFI-4: s64 vectors refuse to hold negative values

jpellegrini:
The condition for setting values was swapped in both
(signed vectores did not accept signed numbers, and
 unsigned vectors did)
Fix #141

### DIFF
--- a/src/uvector.c
+++ b/src/uvector.c
@@ -248,14 +248,14 @@ static void uvector_set(int _UNUSED(type), SCM v, long i, SCM value)
       break;
     case UVECT_S64:
       if (INTP(value) || BIGNUMP(value))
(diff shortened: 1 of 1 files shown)


egallesio:
Thanks for the fix!

PR #147, opened 2020-11-14
Add SRFI-130 - Cursor-based string library

jpellegrini:
Hello @egallesio !

The tests are huge, so I put them in a separate file.

SRFI-130 redefines some procedures from SRFI-13. I am not sure if those are supposed to  be different, but the tests fail unless I force the use of the new ones, so I did this:

```
(require "srfi-13")

(define-module SRFI-130

;;;;;;;;; code goes here

) ;; END OF DEFINE-MODULE
;;;; ======================================================================
(select-module STklos)

(define srfi-13:string-index string-index)
(define srfi-13:string-index-right string-index-right)
(define srfi-13:string-split string-split)
(define srfi-13:string-skip string-skip)
(define srfi-13:string-skip-right string-skip)

(import SRFI-130)

(define string-index (in-module SRFI-130 string-index))
(define string-index-right (in-module SRFI-130 string-index-right))
(define string-split (in-module SRFI-130 string-split))
(define string-skip (in-module SRFI-130 string-skip))
(define string-skip-right (in-module SRFI-130 string-skip-right))

(provide "srfi-130")
```

Is this ok?

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -72,6 +72,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-118: Simple adjustable-size strings 
     - SRFI-128: Comparators (reduced)
     - SRFI-129: Titlecase procedures
+    - SRFI-130: Cursor-based string library
     - SRFI-141: Integer Division
     - SRFI-145: Assumptions
     - SRFI-151: Bitwise Operations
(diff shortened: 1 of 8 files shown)


egallesio:
Thanks  for this new SRFI 

I  have adapted and committed this PR. However, your `lib/Makefile.am` references two files
- srfi-130-macros.stk
- srfi-130-portable.stk

It seems to work without it, so I have committed. 
I let the PR open just in case.

jpellegrini:
> your `lib/Makefile.am` references two files

Oops! Those are not needed, I have incorporated them in `srfi-130.stk`.

PR #148, opened 2020-11-14
Let reader accept #i for rationals

jpellegrini:
Motivation: the tests for SRFI 194 (random data generators) use this, and several Scheme implementations accept it also (Guile, Racket, Chibi, Chicken, Gauche, Chez).

Using the inexact flag, the reader now accepts

 #i0/1    => 0.0
 #i2/3    => 0.666666666666667

 #i1+2/5i => 1.0+0.4i
 #i1@1    => (make-polar 1.0 1.0)

but it will *not* accept fractions made with floats, such
as

1/2.0
1/2.
1.0/2

I also uncommented some unrelated tests from the number test suite which actually pass (!)


egallesio:
Thanks for this fix @jpellegrini . 
Merged.

PR #150, opened 2020-11-20
Add SRFI-135 - Immutable Texts

jpellegrini:
Hello @egallesio !
One more SRFI -- R7RS Large's Immutable Texts!

The only changes I made are:

- check wether string `x` is empty before doing `(string-copy s i x)` (some other Schemes do not complain with `(string-copy! "a" 1000 "")`, some do -- STklos does, so I added the check.
```
                        ((string? x)
                         (when (> (string-length x) 0)
                           (string-copy! s i x))
                         (inner-loop (cdr stuff) (+ i (string-length x))))
                        (else
                         (when (> (textual-length x) 0)
                           (string-copy! s i (textual->string x)))
```
- adapt William Clinger's tests to the STklos test suite (automated, using an Emacs Lisp function)


egallesio:
Thanks a lot  for this implementing this SRFI. 
Your PR was missing the test  file (btw, that geat to have an emacs macro for building it). 
Anyway I have migrated your PR in the master tree with a couple of simple tests in `test/srfis/135.stk`. 
Can you please place the file you have in place of this file when you  have time? 
I don't know what is the more convenient (close this PR and open another one, or modify the current one), so I let this PR open for now.

jpellegrini:
Ok @egallesio - I'll make a second PR then!

PR #153, opened 2020-11-22
Fix reading of illegal hex escape sequences in strings

### LINKED ISSUE #152: STklos reader gets confused with illegal escape sequences in strings
If I try to enter the string `"\x9 +"` STklos correctly sigals an error, but interprets the rest of the string as if it were input to be read and evaluated:

```
stklos> "\x9 +"
**** Error:
%read: bad inline hexa sequence (9  1) on port `#[input-virtual-port 7f45d77140b0]'
#[primitive +]
stklos>                       ;; I entred ^D here!
**** Error:
%read: end of file while reading a string on `#[input-virtual-port 7f45d77140b0]'
```

The same for `" \xo "`...

Found it while porting SRFI-207.

jpellegrini:
Fix #152

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -136,14 +136,18 @@ static int read_hex_sequence(SCM port, char* utf8_seq)
   while ((i < sizeof(buffer) - 1) && isxdigit(c) && (c != ';') && (c != EOF));
   buffer[i] = '\0';
 
-  if (c != ';')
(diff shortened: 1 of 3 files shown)


egallesio:
I close this PR (but not the related issue), because I have used another method to fix the problem (ut I sole your tests :wink: )

PR #154, opened 2020-11-23
Add SRFI-137 - Minimal Unique Types

jpellegrini:
A small and very simple SRFI...


jpellegrini:
Updated to the new SRFI infrastructure (great work, BTW!)

egallesio:
Hello, 

Thanks for this PR (even if i have not used your code): I have implemented it using the `make-struct-type` builin which supports sub-typing. The implementation is even simpler and consists just in a unique function returning the five values.

PR #156, opened 2020-11-26
Use original tests for SRFI-135

jpellegrini:


egallesio:
Cool. I merge it immediately.

egallesio:
Wow!! there are now more than 10k tests. 
Thanks again for all your help.

PR #157, opened 2020-11-27
Fix any and every in bonus.stk

### LINKED ISSUE #137: "every" from bonus.stk is not fully compatible with SRFI-1
Hello @egallesio !

`every` from `bonus.stk` does not return the last PRED value, as the SRFI requires:

> The iteration stops when a false value is produced or one of the lists runs out of values. In the latter case, every returns the true value produced by its final application of `pred`.

This makes the reference tests for SRFI-134 fail.

Since this is not in `srfi-1.stk`, but in `bonus.stk`, I am not sure which you'd prefer:

* keep the version of `every` in `bonus.stk` as is, and makea new one in `srfi-1.stk` (then requiring srfi-1 would override the embedded version of `every`)
* replace the `every` code in `bonus.stk`

If you tell me which one you'd like to have, I can make a PR.

jpellegrini:
every from bonus.stk does not return the last PRED value, as the SRFI
requires:

 | The iteration stops when a false value is produced or one of the
 | lists runs out of values. In the latter case, every returns the
 | true value produced by its final application of pred.

This commit brings any and every from the SRFI-1 reference implementation
(adapted).

Fix #137

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -1,4 +1,4 @@
--;;;;
+;;;;
 ;;;; bonus.stk  -- Useful primitives not in R5RS
 ;;;;
 ;;;; Copyright © 2000-2020 Erick Gallesio - I3S-CNRS/ESSI <eg@unice.fr>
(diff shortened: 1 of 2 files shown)


egallesio:
Hi @jpellegrini, 

I have finally implemented it in another way to avoid the `call/ec` . I suppose that this version is faster than the one in this PR (not tested in fact). BTW, this code was not correct since 
```
(every (lambda (x y) (and (positive? x) (positive? y) (+ x y))) '(1 2 3) '(4 5 6))
```
returned 5 instead of 9.
 BTW, fast path was correct  `(every (lambda (x) x) '(1 2 3)) => 3` 

I have added some tests to yours.

PR #158, opened 2020-11-28
Add SRFI-134 - Immutable Deques

jpellegrini:
Here's SRFI-134!
It will pass tests after issue #137 is fixed (see PR #157 )


egallesio:
Thanks for it. Merged!

PR #160, opened 2020-12-12
write: check if port is textual

### LINKED ISSUE #159: STklos crashes when trying to output characters into bytevector port with write
Hello @egallesio -- still while trying to port SRFI-207, I found another problem. So, R7RS says bytevector ports should work with `write-u8` and `read-u8`, and that works fine in STklos; but when `write` is used STklos crashes:

```
stklos> (call-with-port
 (open-output-bytevector)
 (lambda (out)
   (write #\a out)
   (write #\b out)
   (get-output-bytevector out)))
Received a SIGSGV signal.
Try to augment stack size (--stack-size option).
If the problem persists, send a mail to <bugs@stklos.net>
```

When `display` is used, then STklos raises the error and doesn't crash:

```
(call-with-port
 (open-output-bytevector)
 (lambda (out)
   (display #\b out)
   (display #\a out)
   (get-output-bytevector out)))
**** Error:
call-with-port: bad textual port `#[output-bytevector-port 7f43d64f6f20]'
```

One interesting thing: both Chibi and Gauche do accept `write` on bytevector ports... It would be an interesting addition to STklos in the future (automatically call `string->u8` and other auxiliary functions when writing to binary ports).

PS: something else is also strange when porting the reference implementation of SRFI-207; I'm looking into it.

jpellegrini:
Also include a test...

This would fix #159

### DIFF
--- a/src/port.c
+++ b/src/port.c
@@ -624,7 +624,7 @@ doc>
  */
 DEFINE_PRIMITIVE("write", write, subr12, (SCM expr, SCM port))
 {
-  port = verify_port(port, PORT_WRITE);
+  port = verify_port(port, PORT_WRITE | PORT_TEXTUAL);
   STk_print(expr, port, WRT_MODE);
   return STk_void;
 }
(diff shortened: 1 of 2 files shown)


lassik:
LGTM

PR #161, opened 2020-12-15
Add SRFI-207: String-notated bytevectors

jpellegrini:
Hello @egallesio ! Here's SRFI-207...

* Changed read.c:

  - read_hex_sequence accepts a new argument, "use-utf_8",
    because string-notated bytevectors should *not* be read
    as UTF-8.
  - read_string accepts a new argument, "srfi_207".
    When it is true (nonzero), some escapes are forbidden.
    Also, at the end of the function, if it was called for
    reading a SRFI-207 string-notated bytevector, we *must not*
    interpret it as UTF-8, so we do not call STk_makestring,
    but do the same as it (STk_makestring) wold do,
    but without the UTF8 stuff.
  - maybe_read_uniform_vector: accept #u8" for reading SRFI-207
    bytevectors

* Tests from the reference implementation were used: some procedures
were adapted, and a wrapper macro was created, so the tests themselves
did not need to be changed.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -132,6 +132,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-189.stk        \
           srfi-190.stk        \
           srfi-196.stk        \
+          srfi-207.stk        \
           tar.stk             \
           trace.stk
 
@@ -203,6 +204,7 @@ scheme_OBJS = compfile.ostk    \
           srfi-189.ostk        \
           srfi-190.ostk        \
           srfi-196.ostk        \
+          srfi-207.ostk        \
           tar.ostk             \
           trace.ostk
 
--- a/lib/srfi-207.stk
+++ b/lib/srfi-207.stk
@@ -1,0 +1,858 @@
+(require "srfi-9")
+(require "srfi-1")
+(require "srfi-13")
+(require "srfi-151")
+(require "srfi-158")
+
+(define-module SRFI-207
+  (import SRFI-1)
+  (import SRFI-13)
+  (import SRFI-151)
+  (import SRFI-158)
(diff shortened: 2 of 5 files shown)


egallesio:
Hi @jpellegrini 

Happy new year!

I have merged this PR. I have rewritten some of the  C code to separate the reading of the strings and the #u8 strings. It should be more simple to read (at least I hope so :wink:)

BTW, I have also corrected a remaining bug similar to the one you signaled in PR #153.

jpellegrini:
> Happy new year!

Happy new year to you too!

Sorry if the code was not that great...
I have been working on some other SRFIs, but found stumbling blocks in all of them. I think I'll eventually get them done, but it may take a while. I'll try to rework the SRFI-41 PR ( #56 ) so it merges cleanly again (as I remember, it was failing the tests).
Then -- I'll be away for some weeks because I'll be operated on (spine surgery, not risky but absolutely necessary).

Then I'll be back to the other SRFIs!

egallesio:
> Sorry if the code was not that great...

Don't worry: I probably would have done the same thing if I had written it myself. But then I had to reread this code that I hadn't written, and I found it easier to rewrite it. It's easier when you arrive in second :wink:

> I have been working on some other SRFIs, but found stumbling blocks in all of them. I think I'll eventually get them done, but it may take a while. I'll try to rework the SRFI-41 PR ( #56 ) so it merges cleanly again (as I remember, it was failing the tests).

Great. 

> Then -- I'll be away for some weeks because I'll be operated on (spine surgery, not risky but absolutely necessary).
> Then I'll be back to the other SRFIs!

Ouch. Take care of yourself, it is the more important. Take time to fully recover.

PR #167, opened 2021-02-18
Use $(STKLOS_BINARY) instead of hardcoded path

jpellegrini:
This makes cross-compilation much easier.


egallesio:
Oops, my bad. 
Merged.

PR #171, opened 2021-03-09
Allow second argument to log; add tests to exp, log, sin, cos, tan...

### LINKED ISSUE #170: log does not accept a second argument
According to R7RS, it should accept an optional base as 2nd argument (the "z2" argument, on page 38).

jpellegrini:
R7RS allows log to be called with a second argument, the base of the logarithm:

```
(log 16 2) => 4.0
```

Also includes some simple tests for transcendental functions.

This fixes #170


egallesio:
Thanks for the fix.
Merged.

PR #174, opened 2021-03-14
Add SRFI 170 -- POSIX API

jpellegrini:
Hello @egallesio !

I think it's mostly done, there are just a couple of questions and small issues to be dealt with.

Tell me what you think about the proposed changes in core -- I can change this PR to include them.

# Comments

* We need to raise errors reporting `errno`, so:
  * the FFI was not used, since we cannot check `errno` with it;
  * some procedures, like `file-directories`, were re-written
    in srfi-170-impl.c.

* I wrote a new    `&posix-error` condition with slots for several data as recommended per the SRFI. It can be creted in Scheme as

```
(make-condition &posix-error
                           'procedure procedure
                           'args args
                           'errname (%get-posix-error-name errno)  ; a symbol with the same name as the C constant
                                                                   ; describing the error (for instance, `'EINVAL`)
                           'message (posix-get-message-for-errno errno)
                           'errno errno)
```

# Issues

* There's `error_posix` in `system.c` and  `STk_error_posix` in SRFI-170, which uses the newcondition.
  Proposal:
  * remove `error_posix` from `system.c`;
  * include `STk_error_posix` and the `%posix-error` condition in STklos core and update
    calls to `error_posix` to use this new one instead;
  This way we can even remove some code duplication, moving for example `directory-files` and a few others from the SRFI into the core.

* We ignore buffering modes in the SRFI. Is this OK?

* `open-directory` needs review, I think.

* I used `STk_eq` and `STk_intern` to compare symbols. Is that the correct  way?

* should we change `%time` to `time` in STklos core?
  * structure `%time` only has 2 slots, `second` and `nanoescond`, but SRFI-19 requires `time` to contain a slot with its type (utc, monotonic etc)
  * the SRFI explicitly says that the name of the structure should be `time-utc`, probably a subtype of `time`, or `time` with a slot saying it's UTC.
  * the SRFI exports a time-creating procedure `%make-time`, ugly. What about having  a `make-time` procedure in STklos core?

* If `p` is a port, but not a *file* port, should `(terminal? p)` return `#f`  or should it raise an error? The SRFI wasn't really clear.
  Gauche and Loko return `#f`

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -86,6 +86,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-158: Generators and Accumulators
     - SRFI-161: Unifiable Boxes
     - SRFI-169: Underscores in numbers
+    - SRFI-170: POSIX API
     - SRFI-171: Transducers
     - SRFI-173: Hooks
     - SRFI-174: POSIX Timespecs
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -126,6 +126,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-156.stk        \
           srfi-158.stk        \
           srfi-161.stk        \
+          srfi-170.stk        \
           srfi-171.stk        \
           srfi-173.stk        \
           srfi-174.stk        \
@@ -140,7 +141,7 @@ SRC_STK = bigloo-support.stk  \
           tar.stk             \
           trace.stk
 
-SRC_C = srfi-175-impl.c
+SRC_C = srfi-170-impl.c srfi-175-impl.c
 
 scheme_OBJS = compfile.ostk     \
           bigmatch.ostk         \
@@ -202,6 +203,7 @@ scheme_OBJS = compfile.ostk     \
           srfi-156.ostk         \
           srfi-158.ostk         \
           srfi-161.ostk         \
+          srfi-170.ostk         \
           srfi-171.ostk         \
           srfi-173.ostk         \
           srfi-174.ostk         \
@@ -216,7 +218,7 @@ scheme_OBJS = compfile.ostk     \
           tar.ostk              \
           trace.ostk
 
-scheme_SOLIBS = srfi-175-impl.@SH_SUFFIX@
+scheme_SOLIBS =  srfi-170-impl.@SH_SUFFIX@ srfi-175-impl.@SH_SUFFIX@
 
 DOCDB	    = DOCDB
 
--- a/lib/srfi-170-impl.c
+++ b/lib/srfi-170-impl.c
@@ -1,0 +1,1127 @@
+/*
+ * srfi-170-impl.c   -- C part of SRFI-170: POSIX API
+ *
+ * Copyright © 2021 Jeronimo Pellegrini <j_p@aleph0.info>
+ *
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307,
+ * USA.
+ *
+ *           Author: Jeronimo Pellegrini [j_p@aleph0.info]
+ *    Creation date: 09-Jan-2021 11:54
+ * Last file update: 13-Jan-2021 00:36 (jpellegrini)
+ */
+
+#include <limits.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/statvfs.h>
+#include <dirent.h>
+#include <fcntl.h>
+#include <string.h>
+#include <time.h>
+#include <grp.h>
+#include <pwd.h>
+
+#include "stklos.h"
+#include "struct.h"
+#include "fport.h"
+
+static SCM file_info_type, dir_info_type, user_info_type, group_info_type;
+static SCM posix_error;
+
+SCM time_type; /* FIXME */
+
+/* 3.1 Error handling */
+
+
+DEFINE_PRIMITIVE("posix-get-message-for-errno", posix_strerror, subr1, (SCM obj))
+{
+    if (!  ( INTP(obj) ||
+             (CONDP(obj) && 1)))
+        STk_error("bad integer or condition", obj);
+    int no;
+    char *errstr;
+    
+    
+    if (INTP(obj)) errstr = strerror(INT_VAL(obj));
+    else if (CONDP(obj)) {
+        /* don't presume that the message field in obj has been
+           set up properly. get the string from errno */
+        SCM e = STk_int_struct_ref(obj,STk_intern("errno"));
+        errstr = strerror(INT_VAL(e));
+    } else STk_error("impossible error!");
+    
+    return STk_Cstring2string(errstr);
+}
+
+
+DEFINE_PRIMITIVE("%get-posix-error-name", get_posix_error_name,subr1,(SCM n))
+{
+    /* FIXME: make names static, don't call STk_intern
+       every time a condition is rqaised... */
+
+    if (!INTP(n)) STk_error("bad integer ~S", n);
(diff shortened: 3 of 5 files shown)


jpellegrini:
I realized I was using `malloc` instead of the STklos wrapper. I have added a commit that fixes it.

egallesio:
I have just tarted to look at this code. I will answer you tomorrow. 
In the meanwhile, it is a **great** work. 
Thanks, for this great contribution. Having access to POSIX primitives is really important. Can't wait to merge your code.

egallesio:
Hi @jpellegrini ,

Whoa, this is really a great stuff you have done for this PR.

There is still a little bit of work. In particular, integration/adaptation of STklos existing functions to this SRFI as you said before.

## Globally 
I have made some changes to your code to avoid C compiler warnings

## 3.1

- Modification of the `&posix-error` condition to inherit from `&error-message`. This permits a better integration of Posix error with STk ones.

## 3.2

My understanding of constants such as `open/exclusive` must be initialized to an integer (O_EXCL) since it is said *Flags is an integer bitmask, composed by adding together any of the following constants: ...*
Their value is a symbol.
Consequently I have not tested yet `open-file` and `fd->port`

What do you think? 

About buffering, I suppose that we can ignore it for nowsince STklos IO are built on an abstract notion of port and buffering is not managed.


## 3.3

About directory-files, the STklos version has only one parameter and SRFI-170 two.
I propose to migrate it in C once the integration of &posix-error is done.


In `make-directory-files-generator` don't we need to have a `close-directory`?
In the reference implementation, we can find

```scheme
(define (make-directory-files-generator dir :optional dot-files?)
  (let ((dir-obj (open-directory dir dot-files?))
        (eof (eof-object))
        (done #f))
    (lambda ()
      (if done
          eof
          (let ((f (read-directory dir-obj)))
             (when (eq? f eof)
               (close-directory dir-obj)
               (set! done #t))
              f)))))
```

Moreover, this implementation doesn't need the ipmortation of SRFI-158. What do you think?

In `open-directory`, I have suppressed the strict control on the boolean since in Scheme, any non #f is considered as true. (BTW, I have done the same thing for `follow?` in `file-info`).

The current `real-path` systematically yield a reallocation error (it seems to
be my version of GC, never seen that before). I rewrote it. Hope it is correct
for you.

## 3.5

Your version of `current-directory` and `set-current-directory` should replace the `getcwd` and `chdir` when ported `&posix-error`

STklos `getpid` should be aliased to `pid`


## 3.11

OK with you: aliasing is needed for the environment variables.

egallesio:
Me again, 

> * I wrote a new    `&posix-error` condition with slots for several data as recommended per the SRFI. It can be creted in Scheme as
>
> ```
> (make-condition &posix-error
>                            'procedure procedure
>                            'args args
>                            'errname (%get-posix-error-name errno)  ; a symbol with the same name as the C constant
>                                                                    ; describing the error (for instance, `'EINVAL`)
>                            'message (posix-get-message-for-errno errno)
>                            'errno errno)
> ```
>

As you'll see, I have changed a bit this definition to allow a better integration with STklos errors by using inheritance (the repl shows now a normal message error) rather than an untrapped one.

> # Issues
> * There's `error_posix` in `system.c` and  `STk_error_posix` in SRFI-170, which uses the newcondition.
>   Proposal:
>
>   * remove `error_posix` from `system.c`;
>   * include `STk_error_posix` and the `%posix-error` condition in STklos core and update
>     calls to `error_posix` to use this new one instead;
>     This way we can even remove some code duplication, moving for example `directory-files` and a few others from the SRFI into the core.

Second solution is better in my opinion. I'll do it in the beginning of next week (I have already rewrote the code for that but not migrated in the core). I'll do that next week.

> * We ignore buffering modes in the SRFI. Is this OK?

Probably yes.

> * `open-directory` needs review, I think.

I have made some little changes, but this is a candidate to go in the core. The problem is that the one in STklos is already used elsewhere in the distribution (for ScmPkg). I'll see that.

> * I used `STk_eq` and `STk_intern` to compare symbols. Is that the correct  way?
Perfect.

> * should we change `%time` to `time` in STklos core?
>
>   * structure `%time` only has 2 slots, `second` and `nanoescond`, but SRFI-19 requires `time` to contain a slot with its type (utc, monotonic etc)
>   * the SRFI explicitly says that the name of the structure should be `time-utc`, probably a subtype of `time`, or `time` with a slot saying it's UTC.
>   * the SRFI exports a time-creating procedure `%make-time`, ugly. What about having  a `make-time` procedure in STklos core?

Probably a good idea (and having more fields is not a problem). If this does not bother you, I let you change it to be compatible with what you need for SRFI-19. Is it OK for you?

> * If `p` is a port, but not a _file_ port, should `(terminal? p)` return `#f`  or should it raise an error? The SRFI wasn't really clear.
>   Gauche and Loko return `#f`

I think that this is reasonable and in the spirit of this SRFI.

jpellegrini:
Hi,

In your redefinition of posix-error

```
  posix_error = STk_defcond_type("&posix-error",
                                 STk_err_mess_condition,
                                 LIST2(STk_intern("errname"), STk_intern("errno")),
                                 module);
```

there is no more `procedure` slot. Is that correct?   Should the procedure name be stored in the `location` slot?

jpellegrini:
> I have made some little changes, but this is a candidate to go in the core. The problem is that the one in STklos is already used elsewhere in the distribution (for ScmPkg). I'll see that.

We can use dynamic dispatch if necessary... No?

egallesio:
> there is no more `procedure` slot. Is that correct? Should the procedure name be stored in the `location` slot?

Yep, this is the name that is used for indicating the name of the function where the error is detected.

egallesio:
> We can use dynamic dispatch if necessary... No?

Yes, eventually. I have seen how it is managed in your code; we can keep it.

jpellegrini:
> The current real-path systematically yield a reallocation error (it seems to be my version of GC, never seen that before). I rewrote it. Hope it is correct for you.

It seems to work fine here... Although I had problems implementing it (as per my comment in the code).

jpellegrini:
>  Consequently I have not tested yet `open-file` and `fd->port`

Those are tested in the test suite; they seem to work.

jpellegrini:
Ok - I have pushed all corrections that do not depend on your rewrite of `&posix-error`. I'll wait for that then.

jpellegrini:
It looks like this implementation of SRFI-170 does not work on Android. When I  `(require "srfi-170")`, it complains about symbol `clock_gettime` not being defined...

jpellegrini:
@egallesio I'm a little lost -- besides the Android issue (`clock_gettime` not defined), are there any other problems pending for this SRFI?

egallesio:
> When I (require "srfi-170"), it complains about symbol clock_gettime not being defined... 
Does it work before, I touch your files? 
> I'm a little lost -- besides the Android issue (clock_gettime not defined), are there any other problems pending for this SRFI?
No. It is OK for me :+1:

jpellegrini:
> Does it work before, I touch your files?
No, it seems to be some Android-specific thing.,.

jpellegrini:
```
stklos> (require "srfi-170")
**** Error:
try-load: cannot open object file /data/data/com.termux/files/usr/lib/stklos/1.60/srfi-170-impl.so (dlopen failed: cannot locate symbol "clock_gettime" referenced by "/data/data/com.termux/files/usr/lib/stklos/1.60/srfi-170-impl.so"...)
stklos> 
```

The strange thing is that the tests do pass, no problem. But it fails when I try to load it from the REPL...

jpellegrini:
@egallesio -- I have only one Android device to test, so I'm not sure if it's even reproducible. Did you also test that?

egallesio:
I have tested yet. I'm rewriting parts of this SRFI. I'm moving some code form the SRFi to the base to avoid to have functions which are nearly identical in base and in the SRFI. `&posix-error` is now in the base, system functions will use posix-errors instead of "normal" errors.

Anyway, I suppose that the problem, is with the second parameter of `dlopen`, a flag which is perhaps different. I will have a look at this point as soon I   I have finished to intricate posix and base functions.

egallesio:
Still not tested on Android. :disappointed: 

I have integrated as much of the functions of SRFI-170 in the base. In fact, I have integrated most of the functions which are similar to STklos base and changed some names to be SRFI-170 compliant  (for instance the make-directory in base has been replaced by the create-directory with its optional parameter).  I still have some functions to integrate, and will finish that soon. 

BTW, I have changed the `create-temp-file`  which is now in the base. It returns now two values: a string and a port. I used your implementation, but I have seen that it leaks a file descriptor (the one created by `mkstemp`)→ I have converted it to a Scheme port, so that it can be deleted by the GC (hence the 2 values result). Finally, it simplifies the usage of this function. Still need to see when it is used to avoid reopening a new port on temporaries file names in existing code.

So there is still some work on this SRFI, but  the integration of common operation between STklos and the SRFI should make things simpler to use.

To b continued

jpellegrini:
Strange.
It passes tests on Android, but does not load `srfi-170-impl.so` from the REPL...

PR #178, opened 2021-03-19
SRFI-170 -- missing bits

jpellegrini:
@egallesio  -- I'll send individual commits for each remaining issue. Is this ok?

### DIFF
--- a/lib/date.stk
+++ b/lib/date.stk
@@ -54,8 +54,8 @@ doc>
 |#
 (define (time->seconds time)
   (if (time? time)
-      (exact->inexact (+ (%fast-struct-ref time %time 'second 0)
-                         (/ (%fast-struct-ref time %time 'nanosecond 1)
+      (exact->inexact (+ (%fast-struct-ref time %time 'second 1)
+                         (/ (%fast-struct-ref time %time 'nanosecond 2)
                             #e1e9)))
       (error 'time-seconds "bad time ~S" time)))
 
@@ -77,7 +77,7 @@ doc>
       (cond
         ((real? x)
          (let ((n (inexact->exact (round (* x #e1e9)))))
-           (make-struct %time (quotient n #e1e9) (remainder n #e1e9))))
+           (make-struct %time time-utc (quotient n #e1e9) (remainder n #e1e9))))
         ((integer? x)
          (make-struct %time  x 0))
         (else
--- a/lib/srfi-170-impl.c
+++ b/lib/srfi-170-impl.c
@@ -41,6 +41,10 @@
 #include "struct.h"
 #include "fport.h"
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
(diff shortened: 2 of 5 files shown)


jpellegrini:
Added a check for `PATH_MAX`, which may not necessarily be defined.
That said --
* I don't yet have STklos working on the Hurd, but I think it will work (I can compile and run it, but generating shared libs is not yet working)
* This SRFI seems to compile with the `PATH_MAX` patch, and it would be interesting to see if it tests fine on the Hurd, which does not aim for full POSIX compatibility: https://www.gnu.org/software/hurd/faq/posix_compatibility.html
More on that later...

egallesio:
Hi @jpellegrini ,

I'm back. Sorry for the long delay to merge this PR.

> I don't yet have STklos working on the Hurd, but I think it will work (I can compile and run it, but generating shared libs is not yet working)

Impressing. Great to know that STklos is nearly ready for GNU Hurd!

However, after merging it, I will delete the code you added at  the end of `system.c`:

```c
 STk_define_variable(STk_intern("time-tai"),       STk_intern("time-tai"), current_module);
  STk_define_variable(STk_intern("time-utc"),       STk_intern("time-utc"), current_module);
  STk_define_variable(STk_intern("time-monotonic"), STk_intern("time-monotonic"), current_module);
  STk_define_variable(STk_intern("time-thread"),    STk_intern("time-thread"), current_module);
  STk_define_variable(STk_intern("time-process"),   STk_intern("time-process"), current_module);
  STk_define_variable(STk_intern("time-duration"),  STk_intern("time-duration"), current_module);
```

The reason is that for now all the global symbols in STklos are functions, macros or types (structures or conditions). I think that there are no global which contains simple values (should verify that this claim is true, btw)

If I understand, you want to create symbols whose value is the symbol itself. Keywords can do that

```scheme
stklos> :time-tai
:time-tai
stklos>
```

So I think that we can keep the time field "type" in the `%time` object as a keyword. After that, SRFI-19 can starts with definitions such as

```scheme
;; SRFI-19 time type names
(define time-tai :time-tai)
(define time-utc :time-utc)
...
```

Hence, programs which do not use SRFI-19 will not have their space name polluted by SRFI-19 time types names.

I will modify also the rest of the patch accordingly.

Hope this will not complicate your work on SRFI-19.

What do you think?

jpellegrini:
> Impressing. Great to know that STklos is nearly ready for GNU Hurd!

I just found that the Hurd seems to have an issue with libgc, so we may need to wait for them to get that sorted out. Other than that, there are only minor problems.

jpellegrini:
> ```scheme
> ```scheme
> 
> ```
> 
> 
> Hence, programs which do not use SRFI-19 will not have their space name polluted by SRFI-19 time types names.
> I will modify also the rest of the patch accordingly.
> Hope this will not complicate your work on SRFI-19.
> What do you think?
> ```

Perfect! :)

PR #181, opened 2021-04-04
Add SRFI-25: Multi-dimensional Array Primitives

jpellegrini:
# SRFI-25 proposal

Hi @egallesio !
I have tried to make this one as efficient as possible, while keeping the code clean.
And I have implemented several extra procedures. See below...

## commit message

Also included are the following procedures from the array library
proposed by the SRFi author, Jussi Piitulainen:

```
;;; (array-shape arr)
;;; (array-length arr dim)
;;; (array-size arr)
;;; (array-equal? arr1 arr2)
;;; (shape-for-each shp proc [ind])
;;; (array-for-each-index arr proc [ind])
;;; (tabulate-array shp proc)
;;; (tabulate-array! shp proc ind)
;;; (array-map [shp] proc arr0 arr1 ...)
;;; (array-map! arr [shp] proc arr0 arr1 ...)
;;; (array->vector arr) (array->list arr)
```

And some others:

```
shared-array?
array-share-count
array-copy+share
```

## documentation

I did not document the SRFI procedures, but there are extra 
procedures that implementors usually add to the SRFI (they
are in the SRFI git repository, but not in the SRFI text).

I did document those, but I am not sure where they'll end up in
the documentation.

Also, how do I generate the documentation? It seems that `make`
is not enough.

## equal?

I have changed boolean.c, but it is not enough -- I'm not sure
where else to look.

PENDING: in bookean.c I didn't yet check array shape compatibility

## efficiency

I have tried as much as possible to make a fast
implementation -- however, there are lots of verifications
necessary for each element access. For example,
the element position may be specifed as arguments, as
elements in a vector, or as elements in an array.
This requires one more "if"...

## shared arrays

The SRFI requires it to be possible to share array 
elements through a procedure. Since our implementation
is in C, I have not used the reference code.

I have implmented an extra procedure, shared-array?,
which checks if an array is shared.

An array remembers if it points to some other array's data,
or if someone else points to its data. This is kept in a
counter, protected by a mutex.
The `shared-array?` and `share-array-count` procedures 
expose that to the user.

I did *not* implement a bookkeeping method to tell which
vectores share data with which ones, because it didn't
seem necessary, although it would be fun to imeplemnt.

## empty arrays and default values

The SRFI requires support for empty arrays (or at least
it is implied by the reference implementation and mailing
list archives). But the reference implementation is not
consistent, but I have adopter the same behavior:

* arrays are created with a default value, so
  (make-array (shape)) has a default value of #void
  
  (array-ref (make-array (shape)))  =>  #void
  
* from the reference implementation, from Kawa and Gauche,
  (shape) is an array of rank 2, but there are no indices that
  can be used on it, so referencing (shape) is an error:
  
  (array-ref (shape) i j) => out of bounds, no matter what i, j are!

* from the reference implementation:

 ` (define a (make-array (shape) -2))  ; empty array, default value -2`
  `(array-ref a) => -2`
  
  `(define b (make-array (shape 1 1) -2))  ; empty array, not referenceable:`
  `(array-ref b) => error`

  So...
  
## printing arrays

I uesd a somewhat simple representation,

```
stklos> (define a (array (shape 0 2 0 3 0 4)
                ;; (0 i j)
                -1  -2  -3  -4
                -5  -6  -7  -8
                -9 -10 -11 -12
                
                ;; (1 k j)
                -13 -14 -15 -16
                -17 -18 -19 -20
                -21 -22 -23 -24))

stklos> a
#3A((0 2 0 3 0 4)(-1 -2 -3 -4)(-5 -6 -7 -8)(-9 -10 -11 -12)(-13 -14 -15 -16)(-17 -18 -19 -20)(-21 -22 -23 -24))
```

I considered Kawa's system, but it's too complex. Maybe in the future...

When an array has a default value, it is shown:

```
stklos> (make-array (shape) 'whatever)
#0A(() whatever)
```

## EXTRA

(briefly mentioned in the commit message; expanded here)

Besides the reference implementation, there is an "array lib" in the SRFI
git repository. Different Schemes vary in what they implement of it.

```
;;; (array-shape arr)
;;; (array-length arr dim) 
;;; (array-size arr)
;;; (array-equal? arr1 arr2)
;;; (shape-for-each shp proc [ind])
;;; (array-for-each-index arr proc [ind])
;;; (tabulate-array shp proc)
;;; (tabulate-array! shp proc ind)
;;; (array-retabulate! arr shp proc [ind])
;;; (array-map [shp] proc arr0 arr1 ...)
;;; (array-map! arr [shp] proc arr0 arr1 ...)
;;; (array->vector arr) (array->list arr)
;;; (share-array/prefix arr k ...) (share-row arr k) (share-column arr k)
;;; (share-array/origin arr k ...) (share-array/origin arr ind)
;;; (array-append dim arr0 arr1 ...)
;;; (transpose arr dim ...)
;;; (share-nths arr dim n)
```

* Chicken and Larceny only implement what is in the SRFI document
* Kawa *only* implements `array-shape` and `array->vector`
* Gauche implements all *except* 
  `tablate-array!`
  `share-array/prefix`
  `share-array/origin`
  `array-append`
  `transpose`
  `share-nths`
* Racket implements all of them *except*
  `tabulate-array!`

We have the same as Gauche, plus `tabulate-array!`.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -34,6 +34,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-18: Multithreading support
     - SRFI-22: Running Scheme Scripts on Unix
     - SRFI-23: Error reporting mechanism
+    - SRFI-25: Multi-dimensional Arrays
     - SRFI-26: Notation for Specializing Parameters without Currying
     - SRFI-27: Source of random bits
     - SRFI-28: Basic Format Strings
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -90,6 +90,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-13.stk         \
           srfi-14.stk         \
           srfi-17.stk         \
+          srfi-25.stk         \
           srfi-26.stk         \
           srfi-27.stk         \
           srfi-31.stk         \
@@ -143,7 +144,8 @@ SRC_STK = bigloo-support.stk  \
           tar.stk             \
           trace.stk
 
-SRC_C = srfi-133-impl.c \
+SRC_C = srfi-25-impl.c \
+        srfi-133-impl.c \
         srfi-170-impl.c \
         srfi-175-impl.c
 
@@ -172,6 +174,7 @@ scheme_OBJS = compfile.ostk     \
           srfi-13.ostk          \
           srfi-14.ostk          \
           srfi-17.ostk          \
+          srfi-25.ostk          \
           srfi-26.ostk          \
           srfi-27.ostk          \
           srfi-31.ostk          \
@@ -225,7 +228,8 @@ scheme_OBJS = compfile.ostk     \
           tar.ostk              \
           trace.ostk
 
-scheme_SOLIBS = srfi-133-impl.@SH_SUFFIX@ \
+scheme_SOLIBS = srfi-25-impl.@SH_SUFFIX@ \
+                srfi-133-impl.@SH_SUFFIX@ \
                 srfi-170-impl.@SH_SUFFIX@ \
                 srfi-175-impl.@SH_SUFFIX@ 
 
--- a/lib/describe.stk
+++ b/lib/describe.stk
@@ -57,6 +57,14 @@
      ((struct-type? x)  (format port "the type structure named ~A"
 				(struct-type-name x)))
      ((module? x)	(format port "a module named ~A" (module-name x)))
+
+     ;; for srfi-25. arrays are a native type, but 'array?' is only available
+     ;; when the SRFI is loaded.
+     ((and (provided? "srfi-25")
+           (array? x))            (format port "an array of rank ~A and size ~A"
+                                          (array-rank x)
+                                          (array-size x)))
+
      (else	  	(format port "an an instance of class ~A"
 				(class-name (class-of x)))))
   (format port ".~%")
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -384,6 +384,7 @@ SCM STk_make_array(int rank, long *shape, SCM init)
                    sizeof(struct array_obj) +
                    sizeof(int) +                  /* shared */
                    sizeof(int*) +                 /* orig_share_count */
+                   ARRAY_MUTEX_SIZE +             /* share_cnt_lock */
                    sizeof(long) +                 /* size */
                    sizeof(long) +                 /* length */
                    sizeof(int) +                  /* rank */
@@ -403,6 +404,8 @@ SCM STk_make_array(int rank, long *shape, SCM init)
   s->rank             = rank;
   s->offset           = 0L;     /* will be updated, see below */
 
+  MUT_INIT(s->share_cnt_lock);
+  
   /* data begins right after the data pointer: */
   s->data_ptr = (SCM *)(&(s->data_ptr)) + 1;
 
@@ -840,6 +843,7 @@ DEFINE_PRIMITIVE("share-array", srfi_25_share_array, subr3, (SCM old_array, SCM 
                    sizeof(struct array_obj) +
                    sizeof(int) +                  /* shared */
                    sizeof(int*) +                 /* orig_share_count */
+                   ARRAY_MUTEX_SIZE +             /* share_cnt_lock */
                    sizeof(long) +                 /* size */
                    sizeof(long) +                 /* length */
                    sizeof(int) +                  /* rank */
@@ -848,6 +852,7 @@ DEFINE_PRIMITIVE("share-array", srfi_25_share_array, subr3, (SCM old_array, SCM 
                    sizeof(long*) +                /* multipliers */
                    sizeof(SCM*));                 /* data_ptr */
                    /* don't include data elements, we'll use the othre array's */
+
 
     /* multipliers */
     long *old_mult = ARRAY_MULTS(old_array);
@@ -879,6 +884,8 @@ DEFINE_PRIMITIVE("share-array", srfi_25_share_array, subr3, (SCM old_array, SCM 
     s->shape            = cshape;
     s->multipliers      = new_mult;
     s->data_ptr         = ARRAY_DATA(old_array);
+
+    MUT_INIT(s->share_cnt_lock);
 
     STk_register_finalizer(a, shared_array_dec_count);
 
@@ -987,6 +994,8 @@ DEFINE_PRIMITIVE("array-copy+share",srfi_25_array_copy_share,subr1,(SCM old_arra
     long total_size =
         sizeof(struct array_obj) +
         sizeof(int) +                  /* shared */
+        sizeof(int*) +                 /* orig_share_count */
+        ARRAY_MUTEX_SIZE +             /* share_cnt_lock */
         sizeof(long) +                 /* size */
         sizeof(long) +                 /* length */
         sizeof(int) +                  /* rank */
@@ -1003,6 +1012,8 @@ DEFINE_PRIMITIVE("array-copy+share",srfi_25_array_copy_share,subr1,(SCM old_arra
 
     memcpy(n,o, total_size);
 
+    MUT_INIT(n->share_cnt_lock);
+    
     /* If old_array was not share-created, then we copied its data vector,
        and nobody points to the new copy. We created a fresh, non-shared array. */
     if (*ARRAY_SHARE_COUNT(old_array) >= 0)  *ARRAY_SHARE_COUNT(new_array)=0;
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -516,7 +516,7 @@ DEFINE_PRIMITIVE("make-array",srfi_25_make_array,subr12,(SCM shape, SCM obj))
 
 DEFINE_PRIMITIVE("shape",srfi_25_shape,vsubr,(int argc, SCM *argv))
 {
-  if (argc % 2) STk_error("bad array shape numracketber or args ~S", argc);
+  if (argc % 2) STk_error("odd number of arguments (~S) given for shape", argc);
 
   /* shape of a shape is 0 d 0 2 */
   long *meta_shape = STk_must_malloc(4*sizeof(long));
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -543,7 +543,7 @@ DEFINE_PRIMITIVE("array", srfi_25_array, vsubr, (int argc, SCM *argv))
     if (argc < 1) STk_error("not enough arguments");
     
     SCM shape = *argv--;
-    if (!STk_srfi_25_shapep(shape)) STk_error("bad array shape ~S",shape);
+    if (STk_srfi_25_shapep(shape) == STk_false) STk_error("bad array shape ~S",shape);
     
     long *cshape = shapetoCshape(shape);
     long len = ARRAY_LENGTH(shape);
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -513,7 +513,7 @@ DEFINE_PRIMITIVE("make-array",srfi_25_make_array,subr12,(SCM shape, SCM obj))
 
 DEFINE_PRIMITIVE("shape",srfi_25_shape,vsubr,(int argc, SCM *argv))
 {
-  if (argc % 2) STk_error("bad array shape nummber or args ~S", argc);
+  if (argc % 2) STk_error("bad array shape number or args ~S", argc);
 
   /* shape of a shape is 0 d 0 2 */
   long *meta_shape = STk_must_malloc(4*sizeof(long));
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -513,7 +513,7 @@ DEFINE_PRIMITIVE("make-array",srfi_25_make_array,subr12,(SCM shape, SCM obj))
 
 DEFINE_PRIMITIVE("shape",srfi_25_shape,vsubr,(int argc, SCM *argv))
 {
-  if (argc % 2) STk_error("bad array shape number or args ~S", argc);
+  if (argc % 2) STk_error("bad array shape numracketber or args ~S", argc);
 
   /* shape of a shape is 0 d 0 2 */
   long *meta_shape = STk_must_malloc(4*sizeof(long));
@@ -637,7 +637,7 @@ DEFINE_PRIMITIVE("array-ref", srfi_25_array_ref,vsubr, (int argc, SCM *argv))
         index = get_index_from_array(array, *argv);
         
     } else STk_error("Index must be vector, array or sequence of integers");
-    
+
     return ARRAY_DATA(array)[index];
 }
 
@@ -897,7 +897,7 @@ DEFINE_PRIMITIVE("share-array", srfi_25_share_array, subr3, (SCM old_array, SCM 
  * (shared-array? array)
  *
  * Will return true when the array has its data shared with other
- * arrays.
+ * arrays, and false otherwise.
 doc>
  */
 DEFINE_PRIMITIVE("shared-array?",srfi_25_shared_arrayp,subr1,(SCM array))
@@ -914,8 +914,9 @@ DEFINE_PRIMITIVE("shared-array?",srfi_25_shared_arrayp,subr1,(SCM array))
  * elements through |(share-array array shape proc)|, and that were not
  * yet garbage collected.
  * Note that it may take a long time for an object to be garbage
- * collected automatically, but it is possible to force a garbage
- * collection pass with |gc|
+ * collected automatically. It is possible to force a garbage
+ * collection pass by calling |(gc)|, but even that does not guarantee that
+ * a specific object will be collected.
 doc>
  */
 DEFINE_PRIMITIVE("array-share-count",srfi_25_array_share_count,subr1,(SCM array))
@@ -1022,7 +1023,7 @@ DEFINE_PRIMITIVE("array-copy+share",srfi_25_array_copy_share,subr1,(SCM old_arra
  * This procedure will apply proc to all valid sequences of
  * indices in |shape|, in row-major order.
  *
- * If |index-object| is not provided, then proc must accept
+ * If |index-object| is not provided, then |proc| must accept
  * as many arguments as the number of dimensions that the shape
  * describes.
  * @lisp
@@ -1035,8 +1036,8 @@ DEFINE_PRIMITIVE("array-copy+share",srfi_25_array_copy_share,subr1,(SCM old_arra
  * [2 11]
  * @end lisp
  * If |index-object| is provided, it is used as a place to store the
- * indices, so proc must accept a vector (this is to avoid pushing
- * and popping too many values when calling proc).
+ * indices, so proc must accept either a vector or an array (this is to
+ * avoid pushing and popping too many values when calling |proc|).
  * |index-object|, when present, must be aither a vector or array.
  * @lisp
  * (let ((vec (make-vector 2 #f)))
@@ -1196,7 +1197,7 @@ DEFINE_PRIMITIVE("shape-for-each", srfi_25_shape_for_each, vsubr, (int argc, SCM
 
 static void print_array(SCM array, SCM port, int mode)
 {
-   /* This should probably use shape-for-each, but thta function expects us
+   /* This should probably use shape-for-each, but that function expects us
       to pass it a Scheme procedure to be called in each step, so I left this
       as a hard-coded loop (actually I first wrote this function then used its
       loop as a template for shape-for-each). -- jpellegrini */
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -1216,15 +1216,9 @@ static void print_array(SCM array, SCM port, int mode)
    int rank = ARRAY_RANK(array);
 
    char buffer[100];
-   sprintf(buffer,"#%d",rank);
-   STk_puts(buffer, port);
-
-   /* if the array is shared, we prepend "S" */
-   if (ARRAY_SHARED(array)) STk_putc('S',port);
+   STk_puts("#,(array (",port);
 
    /* write the array shape */
-   STk_puts("A((",port);
-   
    for (int i =0; i< rank; i++) {
      sprintf(buffer,"%ld %ld",
              ARRAY_SHAPE(array)[2*i],    /* lower bound */
@@ -1232,7 +1226,7 @@ static void print_array(SCM array, SCM port, int mode)
      STk_puts(buffer, port);
      if (i != rank-1 ) STk_putc(' ', port);
    }
-   STk_puts(")", port);
+   STk_putc(')', port);
 
    /* now, the tricky part. we could easily just run through ARRAY_DATA(array),
       printing the elements and including ")(" to separate strides, however
@@ -1270,7 +1264,7 @@ static void print_array(SCM array, SCM port, int mode)
    if ( (rank==0) || empty) {
        /* Empty arrays may have a default value. We show it here. */
        if (ARRAY_DATA(array)[0]) {
-           STk_putc(' ', port);
+           STk_putc(' ',port);
            STk_print(ARRAY_DATA(array)[0], port, mode);
        }
    } else {
@@ -1278,12 +1272,7 @@ static void print_array(SCM array, SCM port, int mode)
        int updated = 1;
        while(updated) {
            updated = 0;
-           
-           /* last dimension index is the lowest. open parens */
-           if (INT_VAL(VECTOR_DATA(idx)[rank-1]) == ARRAY_SHAPE(array)[(rank-1)*2])
-               STk_putc('(', port);
-           
-           
+
            /* now we print one element: */
            SCM x = STk_srfi_25_array_ref(2, &args[1]);
            if (x)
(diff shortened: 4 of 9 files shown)


jpellegrini:
This felt like it was quite large, but The `.so` file is smaller than that of SRFI-170.  :)

```
$  ll -h lib/srfi-*-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 36K Mar 15 14:37 lib/srfi-132-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 23K Apr  4 00:34 lib/srfi-133-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 61K Apr  4 00:34 lib/srfi-170-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 38K Apr  4 00:34 lib/srfi-175-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 29K Mar 16 21:13 lib/srfi-208-impl.so
-rwxr-xr-x 1 jeronimo jeronimo 45K Apr  4 19:52 lib/srfi-25-impl.so
```

jpellegrini:
I wanted to check if it was really worth implementing this in C, so I wrote this little program. Not a real benchmark, but helps see how faster it is than the refrence implementation in Scheme:

```
;;; 1) require SRFI-25  if you want to test the C version
;; (require "srfi-25")

;;; 2) Load these if you want to test the reference implementation.
;;;     These files are in the SRFI-25 github repository.
;; (load "srfi-25-reference.scm")
;; (load "arlib.scm")

(require "srfi-27")

(define-syntax push!
  (syntax-rules ()
    ((_ e l)
     (set! l (cons e l)))))

(define (prod v)
  (let ((p 1))
    (do ((i 0 (+ i 1)))
        ((= i (vector-length v)))
      (set! p (* p (vector-ref v i))))
    p))

(define res #f)

(let ((ss '()))
  (do ((i 0 (+ 1 i)))
      ((= i 10))
    (let ((x (random-integer 2))
          (y (random-integer 2)))
      (push! (+ i x y 3) ss)
      (push! (+ i x) ss)))
  (let ((s (apply shape ss)))
    (let ((a (make-array s #t)))
      (let ((idx (make-vector 10 #f)))
        (time (shape-for-each s
                              (lambda (idx)
                                (array-set! a idx
                                            (prod idx)))
                              idx)))
      (set! res a)
      #t)))
```

On my machine, the Scheme implementation takes 6 times longer to finish. The C version was really worth implementing!   :smiley: 

The runinng time for this is somewhat comparable to Gauche and Racket.

I didn't test Chicken, because it has no `shape-for-each` in its SRFi-25.

jpellegrini:
One issue:

This is a test I was willing to add, for threaded creation of shared arrays.

```
(let ((thread-num 50)
      (arrays-per-thread 20)
      (t '())
      (orig (make-array (shape 0 2 0 3) 'unused)))
  (dotimes (i thread-num)
    (push! (make-thread (lambda ()
                          (let ((a '()))
                            
                            (dotimes (j arrays-per-thread)
                              (push! (share-array orig
                                                  (shape 0 2 0 3)
                                                  (lambda (x y) (values x y)))
                                     a))
                            
                            (let Loop ((l a))
                              (test "srfi 25 jp 25.threads.3" #t (shared-array? (car l)))
                              (test "srfi 25 jp 25.threads.4" -1 (array-share-count (car l)))
                              (when (cdr l) (Loop (cdr l)))))))
           t))
  (for-each thread-start! t)
  (for-each thread-join! t) 
  (test "srfi 25 jp 25.threads.1"
        (* thread-num arrays-per-thread)
        (array-share-count orig))
  (test "srfi 25 jp 25.threads.2" #t (shared-array? orig)))
```

It exits with an "`EXIT`" message from STklos, without any more information.
If I comment out the `Loop`, then it passes. 
I'm not sure what I did wrong here.

(I suppose I can use `test` inside the threads I create... No?)

jpellegrini:
Never mind. It seems to be working actually. Wrong parenthesis grouping was causing it to fail.
Pushed this new test in one more commit in this PR.
If you'd like the full PR in a single commit, tell me.

egallesio:
Hi @jpellegrini,

I have just rapidly look your messages , (not found  the time to actually read the code). It is very impressive.
Performances are promising. That's great. 

About, printing , perhaps it could be a good thing to use SRFI-10 sharp comma to represent an array. It would permit to read it back with the standard reader. Defining in the SRFI something like 

```scheme
(define-reader-ctor 'array array)
``` 
and producing `#,(array ...)` when printing an array should permit to ease the reading of  it.
I don't know if it is a goo idea and/or if it implies a lot of modifications in your code. What do you think?

jpellegrini:
> and producing `#,(array ...)` when printing an array should permit to ease the reading of it.
> I don't know if it is a goo idea and/or if it implies a lot of modifications in your code. What do you think?

Ooh, of corse! I had completely forgotten about SRFI-10.
I'll change it and push later.

Now, about reading it:   do you think it's necesary to also implement the reader for this syntax now, or could we do that part later?

jpellegrini:
So. That was actually easy, although not a one-liner in the SRFI...

```
(define-reader-ctor 'array
  (lambda args
    (apply array (apply shape (car args)) (flatten (cdr args)))))
```

jpellegrini:
Ok, I think it's good now.

But @egallesio  -- could you later look into the tests? I tried testing something that needed the GC to run, but `(gc)` doesn't seem to deterministically trigger a gc pass (or did I do something wrong?) The finalizers for shared arrays are not running in the tests.

jpellegrini:
I think I used `inline` instead of `Inline`.   Do I add a new commit fixing that?

egallesio:
## documentation

> I did not document the SRFI procedures, but there are extra
> procedures that implementors usually add to the SRFI (they
> are in the SRFI git repository, but not in the SRFI text).
>
> I did document those, but I am not sure where they'll end up in
> the documentation.

Yes, this is the first time that we have this. Perhaps, We can add
them in the 14.3 section (Misc Information in the SRFI section).


> Also, how do I generate the documentation? It seems that `make`
> is not enough.

No; Generating of the documentation is really epic now. It runs with
personal, and not distributed version of Skribe. This program cannot
run on recent versions of STklos (mostly because symbols are case
significant now I think, but I didn't take the time to investigate).
In fact, I should make a dockerized version of Skribe just for that It
is on my TODO-list since a long time.

Anyway, I will add them to this place if you want.


## equal?
>
> I have changed boolean.c, but it is not enough -- I'm not sure
> where else to look.
>
> PENDING: in bookean.c I didn't yet check array shape compatibility
>

I have modified the base and it is no more necessary to modify `boolean.c`
(and `stklos.h` btw). I have also added a way to avoid the modification of
`describe.stk`.  The new base permits the writing of a SRFI without modifying
the base (and more generally some code which adds some new type).

I have rewrote the code of your SRFI to use the new base, and put your equal
code in the `lib/srfi-25.imp.c` file. It seems to be correct now (not added
the array shape compatibility).


## efficiency

> I have tried as much as possible to make a fast
> implementation -- however, there are lots of verifications
> necessary for each element access. For example,
> the element position may be specifed as arguments, as
> elements in a vector, or as elements in an array.
> This requires one more "if"...

Yes, the way arrays are specified in this SRFI implies a lot of
controls. I think that nothing can be done to avoid that.

## shared arrays
>
> The SRFI requires it to be possible to share array
> elements through a procedure. Since our implementation
> is in C, I have not used the reference code.
>
> I have implmented an extra procedure, shared-array?,
> which checks if an array is shared.
>
> An array remembers if it points to some other array's data,
> or if someone else points to its data. This is kept in a
> counter, protected by a mutex.
> The `shared-array?` and `share-array-count` procedures
> expose that to the user.
>
> I did *not* implement a bookkeeping method to tell which
> vectores share data with which ones, because it didn't
> seem necessary, although it would be fun to imeplemnt.

Shared arrays are beyond my head. I must admit that I don't really
understand why we need them and how tho use them.

Didn't even found a way to effectively see that there is a share. Shouldn't
modifying an element should modify the array in several places? I have not
been able to prove this point with small tests. I think I have definitively
have not understood the shared array concept.

## empty arrays and default values

> The SRFI requires support for empty arrays (or at least
> it is implied by the reference implementation and mailing
> list archives). But the reference implementation is not
> consistent, but I have adopter the same behavior:
>
> * arrays are created with a default value, so
>   (make-array (shape)) has a default value of #void
>
>   (array-ref (make-array (shape)))  =>  #void
>
> * from the reference implementation, from Kawa and Gauche,
>   (shape) is an array of rank 2, but there are no indices that
>   can be used on it, so referencing (shape) is an error:
>
>   (array-ref (shape) i j) => out of bounds, no matter what i, j are!
> * from the reference implementation:
>
>  ` (define a (make-array (shape) -2))  ; empty array, default value -2`
>   `(array-ref a) => -2`
>
>   `(define b (make-array (shape 1 1) -2))  ; empty array, not referenceable:`
>   `(array-ref b) => error`
>
>   So...

Weird idea! Here again I don't understand the why it is so!


 ## EXTRA

> (briefly mentioned in the commit message; expanded here)
>
> Besides the reference implementation, there is an "array lib" in the SRFI
> git repository. Different Schemes vary in what they implement of it.
>
> [ ...]

Awesome

-----------------------------------------------------

> I had the impression that this was quite large, but The `.so` file is smaller than that of SRFI-170.  :)
>
> ```
> $  ll -h lib/srfi-*-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 36K Mar 15 14:37 lib/srfi-132-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 23K Apr  4 00:34 lib/srfi-133-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 61K Apr  4 00:34 lib/srfi-170-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 38K Apr  4 00:34 lib/srfi-175-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 29K Mar 16 21:13 lib/srfi-208-impl.so
> -rwxr-xr-x 1 jeronimo jeronimo 45K Apr  4 19:52 lib/srfi-25-impl.so
> ```

That's true!! And I see also that you have some SRFI implementation that I don't know off :smiley:

----------------------------------------------------------------------

> Never mind. It seems to be working actually. Wrong parenthesis grouping was causing it to fail.

Happy that you didn't find a bug in the thread part (I hate this kind of bug).

> If you'd like the full PR in a single commit, tell me.

Several commits are fine and easier to read.

----------------------------------------------------------------------

> Now, about reading it: do you think it's necessary to also implement the reader for this syntax now, or could we do that part later?

This syntax permits to avoid to have a special reader.

> I'll try to do that anyway. But evidently, shared arrays will become non-shared if they're written and later read back.

True. Do you think it is a problem?

About the external form of an array. I have changed you code. I hope you won't mind. In fact, the code you submitted has the values in a list, and the parenthesis needed to be deleted (I have understood that the fix was simple after having changed your code . :confused:)

The writing of `(array (shape 0 2) 'foo 'bar)` is now `#,(array #,(shape 0 2) foo bar))`. That means that `shape` is also a key for read. and we have

```scheme
stklos> #,(shape 0 2)
#,(array #,(shape 0 1 0 2) 0 2)
```

Probably not useful. Anyway, this writing is probably to much convoluted and I
can go back to your notation after having deleted the superfluous
parentheses). Tell me.


### Things I need to do

- Add your documentation
- Go back on the way to write arrays
- Other things?

egallesio:
> I think I used `inline` instead of `Inline`. Do I add a new commit fixing that?

No, the inline macro was because at the time of the original writing of STklos (pre-2000), the inline keyword was not accepted by all compilers. Using a macro permitted to replace it by an empty string. This is, hopefully, no more necessary today.

egallesio:
BTW, there are a lot of limitations that  could be avoided today (for instance, declaring a variable in a `for` loop, using the dot notation in structure aggregates at initialization time, ...). After all, all this point are in C-99 which was defined more than 20 years ago :smiley:

jpellegrini:
> ## documentation
> No; Generating of the documentation is really epic now. It runs with
> personal, and not distributed version of Skribe. This program cannot
> run on recent versions of STklos (mostly because symbols are case
> significant now I think, but I didn't take the time to investigate).
> In fact, I should make a dockerized version of Skribe just for that It
> is on my TODO-list since a long time.

Oh, I see. No problem then

> Anyway, I will add them to this place if you want.

That would be nice, if it's not too much trouble.

> ## equal?
> I have rewrote the code of your SRFI to use the new base, and put your equal
> code in the `lib/srfi-25.imp.c` file. It seems to be correct now (not added
> the array shape compatibility).

Great news!
 
> ## shared arrays
> > The SRFI requires it to be possible to share array
> > elements through a procedure. Since our implementation
> > is in C, I have not used the reference code.
> > I have implmented an extra procedure, shared-array?,
> > which checks if an array is shared.
> > An array remembers if it points to some other array's data,
> > or if someone else points to its data. This is kept in a
> > counter, protected by a mutex.
> > The `shared-array?` and `share-array-count` procedures
> > expose that to the user.
> > I did _not_ implement a bookkeeping method to tell which
> > vectores share data with which ones, because it didn't
> > seem necessary, although it would be fun to imeplemnt.
> 
> Shared arrays are beyond my head. I must admit that I don't really
> understand why we need them and how tho use them.

The SRFI assumes the array is stored as a contiguous vector, and that the array indices will be translated
into a single index.  `share-array` takes one array and produces another, using the same underlying vector,
but with a different indexing scheme. It may not even use all of the original array.
The translation is through an affine map. I think the SRFI doesn't document it properly for users.
There should be some documentation like this maybe:

When doing

```
(define new (share-array original shape proc))
```

`proc` *must* implement an affine procedure that maps the indices of `new` to the indices of `original` --
say new is given indices `r_1, r_2, ... r_n`.   Then `proc` runs the map

```
s = Ar + c
```

where A is a matrix and `c` is a constant vector, and `s` = `s_1, s_2, ..., s_k` is the resulting set of indices for the old array.

We only run `proc` during array creation, because it can be abstracted away, and shared arrays are as efficient as ordinary arrays.

```
> (define a (array (shape 0 2 0 2) -1 -2 -3 -4))    ;; a simple 2x2 matrix

;; array b will invert the index order, i <-> j, and change signal:
> (define b (share-array a 
                        (shape -1 1 -1 1)
                        (lambda (i j) (values (- j)
                                                (- i)))))

> (shape-for-each (shape -1 1 -1 1) (lambda (x y) (format #t "position ~a ~a -> ~a~%" x y (array-ref b x y))))
position -1 -1 -> -4
position -1 0 -> -2
position 0 -1 -> -3
position 0 0 -> -1
```
The new array will be something like this
```
     y
   -------
x | -4 -2
  | -3 -1
```
with `x` and `y` in the `[-1,1)` range.

> Didn't even found a way to effectively see that there is a share. Shouldn't
> modifying an element should modify the array in several places?

Yes, setting an element changes it in all arrays that share the same underlying store.

```
stklos> (define a (array (shape 0 2 0 2) -1 -2 -3 -4))    ;; a simple 2x2 matrix

;; for Pascal users, we have 1-based indices on array b:
stklos> (define b (share-array a (shape 1 3 1 3) (lambda (i j) (values (- i 1) (- j 1)))))

;; and it uss the underlying store of a:
stklos> (shape-for-each (shape 1 3 1 3)  (lambda (i j)  (format #t  "position ~a ~a -> ~a~%" i j (array-ref b i j))))
position 1 1 -> -1
position 1 2 -> -2
position 2 1 -> -3
position 2 2 -> -4

stklos> (array-set! b 1 2 'x)
stklos> (shape-for-each (shape 1 3 1 3)  (lambda (i j)  (format #t  "position ~a ~a -> ~a~%" i j (array-ref b i j))))
position 1 1 -> -1
position 1 2 -> x
position 2 1 -> -3
position 2 2 -> -4

;; a changed too! see:
stklos> (shape-for-each (shape 0 2 0 2)  (lambda (i j)  (format #t  "position ~a ~a -> ~a~%" i j (array-ref a i j))))
position 0 0 -> -1
position 0 1 -> x
position 1 0 -> -3
position 1 1 -> -4
```

```
> > $  ll -h lib/srfi-*-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 36K Mar 15 14:37 lib/srfi-132-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 23K Apr  4 00:34 lib/srfi-133-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 61K Apr  4 00:34 lib/srfi-170-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 38K Apr  4 00:34 lib/srfi-175-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 29K Mar 16 21:13 lib/srfi-208-impl.so
> > -rwxr-xr-x 1 jeronimo jeronimo 45K Apr  4 19:52 lib/srfi-25-impl.so
```
> That's true!! And I see also that you have some SRFI implementation that I don't know off 

Ah, I have been experimenting... :)

> > I'll try to do that anyway. But evidently, shared arrays will become non-shared if they're written and later read back.
> 
> True. Do you think it is a problem?

No... It would be too cumbersome to come up with reader syntax for reading arrays that share the same underlying store.

> About the external form of an array. I have changed you code. I hope you won't mind. In fact, the code you submitted has the values in a list, and the parenthesis needed to be deleted (I have understood that the fix was simple after having changed your code . )
> 
> The writing of `(array (shape 0 2) 'foo 'bar)` is now `#,(array #,(shape 0 2) foo bar))`. That means that `shape` is also a key for read. and we have
> 
> ```scheme
> stklos> #,(shape 0 2)
> #,(array #,(shape 0 1 0 2) 0 2)
> ```
> 
> Probably not useful. Anyway, this writing is probably to much convoluted and I
> can go back to your notation after having deleted the superfluous
> parentheses). Tell me.

I don't know which is better. When I wrote that I was trying to mimic the behavior of quote: `'(a (b c d))` only needs a first quote, and everything else inside it is quoted. So `#,(array` would imply the first argument is the shape. But I'm not sure it's a good idea. Since `shape` *is* a procedure, your way seems to reflect that better. But it's a bit longer to type if one has to.

> ### Things I need to do
> * Add your documentation

Ok

> * Go back on the way to write arrays

I'm not sure this is needed. Either way works!

> * Other things?

yes --
It seems that the current code is segfaulting when accessing out of bounds for shared arrays. In the weekend I'll try to write an efficient bounds-checking procedure that would run only when an array is created (for each valid index of the new array, calculate its position in the underlying store, and see if it's within bounds).

Also, I added a counter to each array. `0` if it's a standard array, `-1` if it uses some other array's store, and `k` if it is an array whose store is being used by `k` others.
I added a mutex to the array structure, and added a finalizer to decrease it when a shared array is collected. But it doesn't always run! Is this some characteristic of libgc, or did I get something wrong?

jpellegrini:
> Go back on the way to write arrays

We could look into how other Schemes do this. It would be nice to have an external representation that would be readable by one or two different implementations out there (or not?)

jpellegrini:
Look, this is interesting!

I ran this on different Schemes:

```
(define a (array (shape 0 2 0 2) 'w 'x 'y 'z))
(write a)
```

- Gauche: `#,(<array> (0 2 0 2) w x y z)#<undef>`
- Kawa: `#2a((w x) (y z))`  (Kawa has  really sophisticated code for reading/writing objects. It can also print arrays in fancy boxes)
- Chicken: does not have an external representation!
- Sagittarius: does not have an external representation! (it's a SRFI-9 record)
- Racket: does not have an external representation! (it's a SRFI-9 record)
- Larceny: does not have an external representation! (it's a SRFI-9 record)
- Scheme48: does not have an external representation!

Probably severalof the Schemes use the reference implementation, and hence use the SRFI-9 record definition.

jpellegrini:
Oh, there's one more thing. It seems that the same problem that happened with SRFi-170 happens with SRFI-25 on Android.
STklos can't load `srfi-25-impl.so` because it claims `memcpy` is not known.
Probably the same fix for SRFI-170 would work for SRFI-25...

jpellegrini:
> I think the SRFI doesn't document it properly for users.

Maybe also the document should say something like

"If an arraw has shape `(0 P 0 Q o R)`, then to map indices `(pp, qq,r r)` into a single index `i`, we usually do a linear map,
`Q*pp + R*qq+ rr`.
What `share-array` does is to make a different interface to the same array. This time, it will use a composition of a new affine map (given by the user as a procedure) and the original array linear map. Since the result of the composition is an affine map, we can use the map composition coefficients, and array access will be as fast for the shared array as it is for the original one."

Maybe I could write something to be added to the documentation. Better than what I put here, of course -- these are just disconnected ideas.

jpellegrini:
> I added a mutex to the array structure, and added a finalizer to decrease it when a shared array is collected. But it doesn't always run! Is this some characteristic of libgc, or did I get something wrong?

I think you have fixed this (it was the missing pointer dereference). It seems to pass the tests!

jpellegrini:
> I think you have fixed this (it was the missing pointer dereference). It seems to pass the tests!

Spoke too soon. It actually fails (I was looking at the wrong terminal!)

```
test srfi-25 jp 17.share.22 on (array-share-count orig) expected 2 but got 3
test srfi-25 jp 17.share.27 on (shared-array? orig) expected #f but got #t
test srfi-25 jp 17.share.28 on (array-share-count orig) expected 0 but got 3
```

egallesio:
> The SRFI assumes the array is stored as a contiguous vector, and that the array indices will be translated
> into a single index.  `share-array` takes one array and produces another, using the same underlying vector,
> but with a different indexing scheme. It may not even use all of the original array.
> The translation is through an affine map. I think the SRFI doesn't document it properly for users.
> There should be some documentation like this maybe:
>
> When doing
>
> ```
> (define new (share-array original shape proc))
> ```
>
> `proc` *must* implement an affine procedure that maps the indices of `new` to the indices of `original` --
> say new is given indices `r_1, r_2, ... r_n`.   Then `proc` runs the map
>
> ```
> s = Ar + c
> ```
>
> where A is a matrix and `c` is a constant vector, and `s` = `s_1, s_2, ..., s_k` is the resulting set of indices for the old array.
>
> We only run `proc` during array creation, because it can be abstracted away, and shared arrays are as efficient as ordinary arrays.
>
>
> ```
> > (define a (array (shape 0 2 0 2) -1 -2 -3 -4))    ;; a simple 2x2 matrix
>
> ;; array b will invert the index order, i <-> j, and change signal:
> > (define b (share-array a
>                         (shape -1 1 -1 1)
>                         (lambda (i j) (values (- j)
>                                                 (- i)))))
>
> > (shape-for-each (shape -1 1 -1 1) (lambda (x y) (format #t "position ~a ~a -> ~a~%" x y (array-ref b x y))))
> position -1 -1 -> -4
> position -1 0 -> -2
> position 0 -1 -> -3
> position 0 0 -> -1
> ```
> The new array will be something like this
> ```
>      y
>    -------
> x | -4 -2
>   | -3 -1
> ```
> with `x` and `y` in the `[-1,1)` range.
>
>


Thank you @jpellegrini for the clear explanation.


>
> > Didn't even found a way to effectively see that there is a share. Shouldn't
> > modifying an element should modify the array in several places?
>
> Yes, setting an element changes it in all arrays that share the same underlying store.

I have seen it now.

> ```
> stklos> (define a (array (shape 0 2 0 2) -1 -2 -3 -4))    ;; a simple 2x2 matrix
>
> ;; for Pascal users, we have 1-based indices on array b:
> stklos> (define b (share-array a (shape 1 3 1 3) (lambda (i j) (values (- i 1) (- j 1)))))
> ```

Funny. I like this example.




> It seems that the current code is segfaulting when accessing out of bounds for shared arrays. In the weekend I'll try to write an efficient bounds-checking procedure that would run only when an array is created (for each valid index of the new array, calculate its position in the underlying store, and see if it's within bounds).
> 
> Also, I added a counter to each array. `0` if it's a standard array, `-1` if it uses some other array's store, and `k` if it is an array whose store is being used by `k` others.
> I added a mutex to the array structure, and added a finalizer to decrease it when a shared array is collected. But it doesn't always run! Is this some characteristic of libgc, or did I get something wrong?

I'll have a look at that.

Normally, finalizers should be run as soon as the object is collected, but you must be sure that there are no more reference on it and this is always the tricky part of this. I have seen also that all your allocations use `STk_malloc`, but probably some of them could be done using STklos_malloc_atomic (probably all the ones which allocates longs, rather than SCMs). Since Boehm GC is conservative, this permits to be sure that it will not mark by error a zone as used.


BTW, Since I just see it. I have forgot to as you if the modification
in line 926 of the C file was correct (I had a warning and I have
proposed this correction, but I'm not sure of what I have done here).

egallesio:
> Look, this is interesting!

Yes. Very few use an external representation. Thanks for testing. 

I have finally used the Gauche representation (and we have now two implementation which use the same external representation. Since Gauche use, the name `<array>`, I have also defined the class `<array>` in SRFI-25 and added support for defined a new class for new types without modifying the base.

jpellegrini:
> Normally, finalizers should be run as soon as the object is collected, but you must be sure that there are no more reference on it and this is always the tricky part of this. I have seen also that all your allocations use `STk_malloc`, but probably some of them could be done using STklos_malloc_atomic (probably all the ones which allocates longs, rather than SCMs). Since Boehm GC is conservative, this permits to be sure that it will not mark by error a zone as used.

Hmm, ok. I have changed all `STk_must_malloc` instances by their `_atomic` counterpart (the only place where I allocate `SCM` is actually for a vector of integers, so there should be no pointers inside it anyway). And the finalizer still doesn't run. Actually, I think it did run a few times from the REPL, but it's not deterministic, and seldom happens.
I have analyzed the `array_obj` structure, and it doesn't look like anything would point into an array that has `shared`=-1 (which is the one that should have its finalizer triggered). I'll look into this again during the weekend.

The code for checking if a shape can be given to `array-share` is almost ready. I'll make a new PR in the next days.

jpellegrini:
I have made a PR for the shape checking code.

PR #182, opened 2021-04-07
Add SRFI 154: First-class dynamic extents

jpellegrini:
Hi @egallesio !
One more SRFI. A small one, this time.


jpellegrini:
I'll make this one an R7RS lib too.

PR #185, opened 2021-04-08
Add SRFI-29: Localization

jpellegrini:
Hi @egallesio !
One more.

This one has a small problem: Since it needs to change the `format` directives, I added a new `format` procedure, and shadowed the old one:

```
(define %stklos-format format)

(define-module SRFI-29
...

(define (%srfi-29:format A . rest)

);; END OF DEFINE-MODULE

(select-module STklos)
(import SRFI-29)

(define format (in-module SRFI-29 %srfi-29:format))

(provide "srfi-29")
```

But then tests seem to break in a way that's not easy to debug. If I change `format` to `%format`, in the SRFI andin the tests,then all tests pass:

```
(select-module STklos)
(import SRFI-29)

(define %format (in-module SRFI-29 %srfi-29:format))   ;; here, we don't override STklos format, just for testing

(provide "srfi-29")
```

I'm not sure why that happens.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -38,6 +38,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-26: Notation for Specializing Parameters without Currying
     - SRFI-27: Source of random bits
     - SRFI-28: Basic Format Strings
+    - SRFI-29: Localization
     - SRFI-30: Nested Multi-line Comments
     - SRFI-31: A special form for recursive evaluation
     - SRFI-34: Exception Handling for Programs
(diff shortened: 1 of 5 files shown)


jpellegrini:
The last commit seems to fix the SRFI-29 tests, but then STklos fails when testing the next SRFI (37).
Still investigating.

jpellegrini:
> testing the next SRFI (37).

Fixed that. Now it fails on SRFI-48 and I think I've found a bug (see issue #203 )

jpellegrini:
Ok. I found the bug... I was using `list-copy` on a possibly circular list, and the tests for SRFI-48 did send a circular list to `format`.
It's fixed now: passes all tests, and redefines `format` properly. :)

jpellegrini:
(I mean, this would be ready to include in STklos, as far as I can see)

egallesio:
Another one merged :smiley: 
Thanks @jpellegrini

PR #191, opened 2021-04-10
Fix SRFI 170 build error on Mac

lassik:


lassik:
This could also be handled via the `./configure` script, but I don't know how to do that.

egallesio:
Hi @lassik ,

Weird that MacOS doesn't use the POSIX names.  

> This could also be handled via the ./configure script, but I don't know how to do that.

Probably but your correction seems  perfectly fine to me. Applied it.

Thanks for your help.

lassik:
Indeed, I don't know why they don't use the POSIX names, or whether they have any plans to add them.

PR #195, opened 2021-04-17
SRFI 25: add shape checking when creating shared arrays

jpellegrini:
Now the user cannot create shared arrays that are incompatible with the original array shape (leading to out of bound indices).

* I have made functions to show sequences of numbers and to show the mapping as strings. Then, I used `math.h` (and `-lm`, which STklos would already use anyway) to calculate the needed space for them. Is that OK?
* Added some tests for wrong shapes
* I have tried to make a reeeealy informative message. see:

```
stklos> (require "srfi-25")
"srfi-25"
stklos> (define a (make-array (shape 0 2 0 3) 'z))

stklos> (share-array a (shape 0 3) (lambda (x) (values x x)))
**** Error:
share-array: Shape (0 3) does not map to shape (0 2 0 3) under mapping x_0 -> 1y_0; x_1 -> 1y_0. Index (2) for the new array goes out of bounds in the original array.

stklos> (share-array a (shape 0 2 0 3) (lambda (x y) (values y x)))
**** Error:
share-array: Shape (0 2 0 3) does not map to shape (0 2 0 3) under mapping x_0 -> + 1y_1; x_1 -> 1y_0. Index (0 2) for the new array goes out of bounds in the original array.
```

Racket does something similar, but they don't tell the user the last bit (which index goes out of bounds).

Gauche does not check this on shared array creation; it does later during referencing.

### DIFF
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -25,7 +25,7 @@
  */
 
 #include "stklos.h"
-
+#include <math.h>
 
 /*
   ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jpellegrini:
I think this is pending:
* Check why the finalizers are not running. One thing I noticed is that I am copying the mutex instead of using its address. But this should not prevent the finalaizers from running. Still looking into it
* Find out why SRFIs 25 and 170 do not load in Android. I have found that PR #125 seems to help: the `-lc` when compiling shared libs makes the error change: now it does not complain about things from `libc`, but from `libgc`.

jpellegrini:
Aaaand it seems  that pasing `-lc -lgc` fixes SRFI-25 and SRFI-170 on Android!!! (See PR #125 -- there I only add `-lc` only. with `-lgc` it works). I'll add more to that PR.

egallesio:
Great . Merged.

PR #198, opened 2021-04-22
"equal?" arrays should have the same shape

jpellegrini:
Otherwise, we'd be considering 
```
[ a b c d ] 
```
the same as
```
[ a b ]
[ c d ]
```

### DIFF
--- a/INSTALL-android.md
+++ b/INSTALL-android.md
@@ -62,8 +62,14 @@ to forward TCP connections so we can ssh into the device.
 
 ## Set up sshd in Termux
 
-Open Termux in the Android device and prepare it to accept
-ssh connections:
+Open Termux in the Android device and update its packages.
+
+```
+pkg update
+pkg upgrade
+```
+
+Now prepare it to accept ssh connections:
 
 ```
 pkg install openssh
@@ -121,7 +127,11 @@ lrwxrwxrwx 1 u0_a196 u0_a196 28 Jun 28 07:33 pictures -> /storage/emulated/0/Pic
 lrwxrwxrwx 1 u0_a196 u0_a196 19 Jun 28 07:33 shared -> /storage/emulated/0
 ```
 
-Then `cd storage/external-1` and you will be in the SD card directory (you won't
+If you get a "permission denied" error, or if the `storage/` directory was not
+created, please try revoking and granting again permissions to the Termux app
+in your Android configuration (see https://wiki.termux.com/wiki/Termux-setup-storage).
+
+Now `cd storage/external-1` and you will be in the SD card directory (you won't
 see the contents of the SD card there, Android doesn't let you do that -- but
 you are in a directory specifically created for Termux in your extenal SD card).
 
@@ -133,7 +143,13 @@ After ssh-ing into the device, install the build tools. Termux does not have gcc
 apt install libgc libpcreposix make clang git autoconf automake
 ```
 
-Configure: in order to install STklos for usage within Termuc, we use the prefix as below.
+Of course, clone the repository (or you can download a release tarball and extract it):
+
+```
+git clone https://github.com/egallesio/STklos
+```
+
+Configure: in order to install STklos for usage within Termux, we use the prefix as below.
 
 ```
 ./configure --prefix=/data/data/com.termux/files/usr/
--- a/lib/srfi-25-impl.c
+++ b/lib/srfi-25-impl.c
@@ -1523,11 +1523,12 @@ static void print_array(SCM array, SCM port, int mode)
 
 static SCM test_equal_array(SCM x, SCM y)
 {
-  long lx, ly, rx, ry, i;
+  long ix, iy, lx, ly, rx, ry, i;
   SCM *dx, *dy;
   long *sx, *sy;
+  SCM data_x, data_y;
 
-  lx = ARRAY_SIZE(x); ly = ARRAY_SIZE(y);
+  lx = ARRAY_LENGTH(x); ly = ARRAY_LENGTH(y);
   rx = ARRAY_RANK(x); ry = ARRAY_RANK(y);
   if (lx == ly && rx == ry) {
     dx = ARRAY_DATA(x);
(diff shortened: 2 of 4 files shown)


jpellegrini:
There is still a need for `array-equal?`, I think. I'll write more about it later.

jpellegrini:
The added tests show why `array-equal?` is necessary.

```
a: [ -1 -2 -3 -4 -5 -6 -7 -8 ]
b: [ -1 -3 -5 -7 ]
c: share a, using only even indices
   c has addressable elements [ -1 -3 -5 -7 ], although it gets them jumping over a's data.
```

Now, `b` and `c` do _NOT_ have the same `ARRAY_DATA`. But they have the same addressable contents! So, they are not `equal?`, but `array-equal?` does see that and returns `#t`.

jpellegrini:
`array-equal?` can still be optimized (it uses `array->list` on the shape, which is usually small -- we could avoid that)

jpellegrini:
But doesn't it feel like `equal?` should do what `array-equal?` is doing?

jpellegrini:
> Otherwise, we'd be considering

I mean, for consistency with `equal?` on lists: `(equal? '(a b c d)  '((a b) (c d)))` is false...

egallesio:
`array-equal?` can still be optimized (it uses `array->list` on the shape, which is usually small -- we could avoid that)
Yep, but this probably not a problem in practice. 

> But doesn't it feel like `equal?` should do what `array-equal?` is doing?
> Otherwise, we'd be considering
> 
> I mean, for consistency with `equal?` on lists: `(equal? '(a b c d) '((a b) (c d)))` is false...

In fact, we can say that they are `array-equal?` if they print the same.  What do you think?

jpellegrini:
> In fact, we can say that they are `array-equal?` if they print the same. What do you think?

I'm not sure we need two equality predicates. If they print the same, then they have the same shape and addressable contents. That would be what `equal?` checks...
If the arrays are shared or not, should we be dealing  with that?

egallesio:
Hello @jpellegrini,

Ok for keeping only the `equal?` predicate for arrays. As a consequence, I have applied only the two first commits of this PR. I have not tried to make the `equal?` work for shared array. I let you do it in C as you kindly proposed :wink:
For now, I let this PR open.  

BTW, I have integrated your SRFI-25 documentation in the STklos ref. manual.
Thanks again for all.

jpellegrini:
Hi @egallesio !

So, what I meant with my other message was this:

```
stklos> (define a (array (shape 0 8) -1 -2 -3 -4 -5 -6 -7 -8))
;; a
stklos> (define b (array (shape 0 4) -1 -3 -5 -7))
;; b
stklos> (define c (share-array a (shape 0 4) (lambda (x) (* x 2))))
;; c
stklos> b
#,(<array> (0 4) -1 -3 -5 -7)
stklos> c
#,(<array> (0 4) -1 -3 -5 -7)
stklos> (equal? b c)
#f
```

Shouldn't this be `#t`?

I have changed this PR so it does that. `test_equal_array` will fetch each element by its index (and not just traverse `ARRAY_DATA`), so now arrays that print the same would be `equal?`.
What do you think?

PS: I see you have added a release tag, but you did as if it was 1.60 being released? Wouldn't that be 1.70?
Also, maybe this change (array equality) could be included in 1.71, if 1.70 is released already? Or do you think the change is not good?

jpellegrini:
Ops. Wait, something is not yet right.

jpellegrini:
Ok, I think it's almost done, but I don't know why this happens:

```
(let ((a (array (shape 0 2 0 2 ) -1 -2 -3 -4)))
(equal? #,(<array> (1 3 1 3) -1 -2 -3 -4)
        (array-copy (share-array/origin  a 1 1))))
**** Error:
%equal-try: array index `0' < `1' for dimension `0'
```

`%equal-try` checks cycles, right? Why is it failing?

jpellegrini:
> `%equal-try` checks cycles, right? Why is it failing?

Sorry, that was my mistake. It all seems to work now. I have included more tests for `equal?`

The only thing missing for SRFI-25 would be the share counter. It's not in the SRFI spec, but I thought it would be a nice addition. But it we don't get to make it work we can remove it...

(Although it would be rally nice to have the counter)

egallesio:
> Sorry, that was my mistake. It all seems to work now. I have included more tests for equal?

No problem, but since the git history was not clear, I have not applied directly the PR but have  patched the files in `lib` and `tests` directories with your code.  Seems to be OK. 

Tests which uses array-copy to avoid previous behavior of `equal?` have been reverted to not use it anymore.

> PS: I see you have added a release tag, but you did as if it was 1.60 being released? Wouldn't that be 1.70?
Also, maybe this change (array equality) could be included in 1.71, if 1.70 is released already?

No I have added a release tag on 1.60 because someone asked me. 
1.70 is not ready yet :smile: 

Can we close this PR?

egallesio:
OOPS. It seems that we have worked on the same file at the same time, and the last version of `srfi-25-impl.c` I have committed doesn't integrate your last version. I'll commit a new version in a couple of minutes.

egallesio:
No files seems to be identical (some indentations only differ). Forge my previous message.

jpellegrini:
> No files seem to be identical (some indentations only differ). Forget my previous message.

Are you working on merging it? Or would you like me to do that?

egallesio:
> > No files seem to be identical (some indentations only differ). Forget my previous message.
> 
> Are you working on merging it? Or would you like me to do that?

I think that everything has been pushed now.  
Normally, the GitHub tree should be consistent with your tree (I hope so).

jpellegrini:
That's great! It passes the tests, and now there's a progress indicator for SRFIs during the tests. :)

egallesio:
> That's great! It passes the tests, and now there's a progress indicator for SRFIs during the tests. :)

Displaying this indicator was incorrect on Android. So it also permitted to spot a bug !

PR #204, opened 2021-04-28
Add SRFI-46: Basic Syntax-rules Extensions (actually part of R7RS small)

jpellegrini:
Hi @egallesio !

I noticed this was actually easy to implement, so here it is! SRFi-46.
This one is nice because several reference implementations of later SRFIs use the configurable ellipsis, which this one brings.

The commit message:

STklos already has half of this SRFI implemented: mbe.stk
does understand the tail patterns defined in the SRFI text.
The only missing part was the configurable ellipsis symbol,
which this commit brings.

Since STklos does not recognize symbols beginning with ":",
we check for symbols or keywords in front of the SYNTAX-RULES
symbol. This seems to work, and does not break any tests.

This also includes some basic tests.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -52,6 +52,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-41: Streams
     - SRFI-43: Vector library
     - SRFI-45: Primitives for Expressing Iterative Lazy Algorithms
+    - SRFI-46: Basic Syntax-rules Extensions
     - SRFI-48: Intermediate Format Strings
     - SRFI-51: Handling rest list
     - SRFI-54: Formatting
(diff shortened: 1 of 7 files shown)


lassik:
IIRC, the SRFI 46 extensions are included as a standard part of R7RS-small.

jpellegrini:
> IIRC, the SRFI 46 extensions are included as a standard part of R7RS-small.

Oh yes, you're right! It's in section 4.3.2, page 23 of the PDF! :)

Anyway, the macro system is being re-written, so probably this PR is not really important anymore, as a new implementation of `syntax-rules` is likely to be added! :smiley:

jpellegrini:
Hi @egallesio !
I have worked a bit on this one, so now it cleanly applies. I thought that even if STklos will have a different `syntax-rules` implementation in the future, this PR would make sense since it brings some R7RS features.
It passes all tests (and there are some more tests for it included)

egallesio:
Great! I'll merge this PR as soon as possible.
Thanks.

egallesio:
Hi @jpellegrini, 

I don't know why thi PR is still open, since it  has already been merged.
I'm closing it.

PR #212, opened 2021-05-10
Rewrite SRFI-27  (it's up to 10x faster)

jpellegrini:
Hi @egallesio !

So... This is a complete rewrite of SRFI-27!

### Why?

I noticed that tests for SRFI-194 ran very slow, so I did some
measurements and noticed that STklos timing on this SRFI could
be improved. The SRFI, as it is today, is completely implemented
in Scheme, so I thought I'd implement it in C.

I have a special interest in this SRFI, because I do use it to run
some simple simulations.

### What's in this new implementation?

* It is now possible to plug new PRNGs;
* By default we use the Mersenne Twister, which has good statistical 
  properties, very large period and is fast. The implementation is in C;
* Random states are nicely printed;
* `random-source-make-reals` now returns an exact rational when its
  `unit` argument is exact rational (the SRFI actually requires that!),
  and has been written in a careful way so as to do what the SRFI 
  prescribes when generating rationals;
* Added basic tests.

### Timings

I have made some rudimentary measurements on CPU time.

The table below shows that the new code brings a 10x
improvement on running time for generation of integers,
and a 5x improvement for reals, and with this new implementation
STklos would be close to some of the other Schemes.

Measurements are in seconds. This is not a real benchmark
(I did only a couple of runs for each value, for example), but
it shows that there was a great improvement.

|            | bignum | fixnum | fixnum fx+ | real   | real fx+ fl+ |
|------------|--------|--------|------------|--------|--------------|
|STklos OLD  | 224.13 | 27.79  | 27.54      | 27.95  |              |
|STklos NEW  |  19.86 |  2.25  |  2.26      |  5.28  |              |
Chibi            |      9.78   | 9.60  |  4.8        |  7.84 | 7.82 |
|Chicken csi |  26.60 |  9.12  |  9.07      |  9.90  | 9.05         |
|Chicken csc |  15.42 |  1.26  |  2.65      |  1.08  | 2.38         |
|Gambit      |  18.85 |  4.55  |  4.56      |  4.50  | 4.28         |
|Gauche      |  66.16 |  2.34  |  2.55      |  2.87  | 7.3          |
|Guile       | 393.91 |  0.60  |            |  5.35  |              |
|Racket      |  14.74 |  1.50  |            |  0.814 |              |
| | | | | | |
| ABCL      |  17.18           | 11.70 |          | 10.10          | |
| Clasp      |  9.44              | 0.5 |              |    9.2          | |
| Clisp        |  26.58    | 17.82 |                 | 27.39 | |
| Clisp .fas |    7.52 | 2.16      |                 | 12.11 | |
|ECL          |  27.71  | 13.48  |                | 10.55| |
|SBCL        |   3.44 | 1.34 |               |   1.8 | |

* Chibi was very fast, but the result was somewhat imprecise 
  (always answers something around _4.65e+18_, should be closer to _5.0e+18_)

The good performance of STklos with bignums is likely related to STklos
using the GMP, which is *very* fast.

The version of Racket used is recent, so it is already implemented on
top of Chez Scheme.

The columns with `fx+` and `fl+` show timings  using those,
to see if they were relevant for these specific experiments. 
For the Schemes that have them they did not seem to make much 
of a difference. Some even ran slower with them!

I did not find out if Guile and Racket have specialized procedures
like `fx+` and `fl+`.

I could not easily find a way to measure execution time of S-expression
evaluation in MIT Scheme so  I left it out.

The timings for Common Lisp were with type declarations (included because
I thought it'd be interesting). "Clisp .fas" means Clisp copiled file (which is way
faster than interpreted Clisp)

### Issues

* The procedures `random-real` and `random-integer` are not available. For some
reason, when SRFI-27 is autoloaded, those are not imported (but all other
procedures in the module are!)

* I had to replicate parts of the declarations from `bignum.c` in `prng-mt.c`. I'm not sure if 
  it would be be better to factor out definitions into `bignum.h`.

* I have used the 64-bit version of the generator only. It would be possible to use the 32-bit when the
   underlying platform has 32-bit words, and there is a speedup when we do that, but that would complicate 
  the architecture:
  - we'd need to check at runtime if we're on 32 or 64 bits (I don't know how to do that portably at compilation time), and
  - load one of 2 different `.so` files, so we'd have 3 C files instead of one `mersenne-twister.c`.

* I did not include randomness tests in the test suite, I only manually verified that the generated 
  numbers seem to be uniformly distributed.

### Code used for timings

Please tell me if I did something silly here that may have affected the measurements.

```scheme
;; This is for STklos. Should be changed for other Schemes.
;; Some Schemes need SRFI-143 and SRFI-144 for fx+ and fl+
;;
;; Also, Guile doesn't have a `time` macro. Instead, it has a REPL
;; command `,time`.

(require "srfi-27")

(define s (make-random-source))


(time
(let ((sum 0)
      (gen (random-source-make-integers s)))
  (do ((i 0 (+ 1 i)))
      ((>= i 10000000))
    (set! sum (+ 1 (+ sum (gen 10000000000000000000)))))
  (display "----------------------------------X make-integers, bignum")
  (newline)
  (/ sum 10000000.0))
)


(time
(let ((sum 0)
      (gen (random-source-make-integers s)))
  (do ((i 0 (+ 1 i)))
      ((>= i 10000000))
    (set! sum (+ 1 (+ sum (gen 1000)))))
  (display "----------------------------------X make-integers, fixnum")
  (newline)
  (/ sum 10000000.0))
)

(time
 (let ((sum 0)
       (gen (random-source-make-integers s)))
   (do ((i 0 (fx+ 1 i)))
       ((>= i 10000000))
     
     (set! sum (fx+ 1 (fx+ sum (gen 100)))))
   (display "----------------------------------X make-integers, fixnum fx+")
   (newline)
   (/ sum 10000000.0))
 )

(time
(let ((sum 0.0)
      (gen (random-source-make-reals s)))
  (do ((i 0 (+ 1 i)))
      ((>= i 10000000))
    (let ((x (gen)))
      (set! sum (+ sum x))))
  (display "----------------------------------X make-reals")
  (newline)
  (/ sum 10000000.0))
)


(time
(let ((sum 0.0)
      (gen (random-source-make-reals s)))
  (do ((i 0 (fx+ 1 i)))
      ((>= i 10000000))
    (let ((x (gen)))
      (set! sum (fl+ sum x))))
  (display "----------------------------------X make-reals fl+")
  (newline)
  (/ sum 10000000.0))
)
```


jpellegrini:
> * The procedures `random-real` and `random-integer` are not available. For some
>   reason, when SRFI-27 is autoloaded, those are not imported (but all other
>   procedures in the module are!)

Ah yes -- could you help with that? I can't see why those aren't being loaded.

jpellegrini:
The tests for SRFI-216 will fail, but that is only because of the two procedures that are not being made available.

egallesio:
Very impressive! 

> Ah yes -- could you help with that? I can't see why those aren't being loaded.

I suppose it has to do with the autoloads which are broken (I broke them when I rewrote the compiler for integrating SRFI-5). I have already seen that autoload have this weir behavior.   I'll correct that. In the meantime, perhaps, you can try to delete the autoload directive appearing in `lib/boot.stk`. Tell me. 

I'll try to fix the problem.

jpellegrini:
> you can try to delete the autoload directive appearing in `lib/boot.stk`

Yes -- but then it breaks because http.stk uses it.
Do I include a `require` for SRFI-27 in `http.stk`?

jpellegrini:
> Yes -- but then it breaks because http.stk uses it.

And the tests for SRFI-219 also

egallesio:
My bad. The problem seems to be with macros. I don't understand why it does not work in this case. :confused:

jpellegrini:
I have included `(require "srfi-27")` in `tests/srfis/216.stk`, but it stil doesn't seem to work:

```
stklos> (random 10)
**** Error:
random: variable `random-integer' unbound
stklos> random-integer
#[closure random-integer]
stklos> (random-integer 10)
2
```

jpellegrini:
> I have included `(require "srfi-27")` in `tests/srfis/216.stk`, but it stil doesn't seem to work:

The example above is without the autoload, and loading SRFI-216 manually.  That is strange.
Maybe it has to do with SRFI-216 and how it uses `random-integer`?

jpellegrini:
Ah!

SRFI-216 counts on autoload to load `random-integer`, but it refers to the symbol `random-integer` when it's already *inside* the module.
So, SRFI-216 causes SRFI-27 to be loaded, but is `SRFI-27` `import`ed into SRFI-216?

(See, I have used `SRFI-27` as module name. You had another name before...)

jpellegrini:
To adapt to the new way of creating mixed modules I'll need to make a force push here and rewrite the history of this PR.  More later.

egallesio:
> To adapt to the new way of creating mixed modules I'll need to make a force push here and rewrite the history of this PR. More later.

Hi @jpellegrini,

I apologize for the modification. But if you prefer, you can do as usual and I'll adapt your code to the new way to create mixed module (I have let the old targetq in `lib/Makefile.am`.

Not have the time to read all your remarks on the make tests lock (I have seen it also sometimes). I'l look at it asap.

jpellegrini:
> I apologize for the modification.

Oh, I was not complaining! The changes are good.

I have already made some changes, but the module doesn't load correctly.

* I wanted to have pluggable random number generators (a maybe slower but with better quality for crypto, one faster for ordinary things etc). So I implemented one of them in `prng-mt.c`. I did not call it `srfi-27.c` because it really is only the implementation of one of the generators;
* So I used `prng-mt` instead of `srfi-27` in `Makefile.am` and in `srfis-data.stk`, but it seems that I made STklos confused:

```
stklos> (require "srfi-27")
**** Error:
find-module: module with name `SRFI-27' does not exist
stklos> (load "prng-mt")
stklos> (make-random-source)
#[<random-source-mt> 7f44878d7d80]
stklos> 
```

So loading `prng-mt` works, but `require`ing "srfi-27" doesn't... Why?

jpellegrini:
> So loading `prng-mt` works, but `require`ing "srfi-27" doesn't... Why?

Of course, that also break the tests... But I did run them manually and they seem to be ok.

jpellegrini:
> So loading `prng-mt` works, but `require`ing "srfi-27" doesn't... Why?

Hm, is this a bug or am I abusing the system?

```scheme
stklos> (trace %load )
stklos> (require "srfi-27")
.. -> %load with args = ("srfi-27" ("." "/home/jeronimo/.stklos/pkg/lib" "/home/jeronimo/.stklos/ext" "/usr/local/lib/stklos/1.60" "/usr/local/share/stklos/1.60") ("so" "ostk" "spi" "stk" "scm"))
**** Error:
find-module: module with name `SRFI-27' does not exist
stklos> (require "prng-mt")
.. -> %load with args = ("prng-mt" ("." "/home/jeronimo/.stklos/pkg/lib" "/home/jeronimo/.stklos/ext" "/usr/local/lib/stklos/1.60" "/usr/local/share/stklos/1.60") ("so" "ostk" "spi" "stk" "scm"))
.. <- %load returns #void
WARNING: "prng-mt" was not provided
"prng-mt"
```

This works:

```scheme
stklos> (load "prng-mt")
stklos> 
```

and it's what I expected, since I have this in `srfis.stk`:

```scheme
     (27 "Source of random bits" (random) "prng-mt")
```

egallesio:
Hi @jpellegrini, 

We have in fact two problems. Since,  `require` is file based (and not SRFI based),  it tries to load the file `srfi-27.stk` which exists and starts with a `(select-module SRFI-27)` whereas  the `define-module`  is in the C file. In fact, we should use a `(require-feature  "srfi-27")` here, since it looks at the SRI table to find the file to load. 

I propose the following changes for mixed modules
  - change the name of the Scheme file in mixed modules from `srfi-xxx.stk` to `srfi-www-incl.stk`. So, a require of `srfi-xxx` will not try to load a file which contains only a part of the implementation.
  - use `require-feature` in make test to load the files
  - do not install the  `srfi-www-incl.stk` in the `install-sources` target since, this files are not directly loadable, without the C code.  
 
I have tested on SRFI-27, and it works,  except the `autoload` for random-integer and I don't yet understand why.

I'll try to fix all that stuff.

jpellegrini:
Hi @egallesio !
I've tried what you suggested, and it didn't work. I have committed what I have tried so far, but I'm not sure if I have missed something.
I'll get back to this later

egallesio:
Hi @jpellegrini,

I have a version which is completely OK (library is correctly loaded , autoloaad are correct, all tests passed.
However,  since `(random-integer 10)` starts with a great number of 0, I tried to see the if the distributions seems correct.
I started with 

```scheme
stklos> (load "./prng-mt.so")
stklos> (dotimes (i 10) (print (random-integer 10)))
0
0
0
9
0
0
0
9
0
3
```
OK seven 0 on  10 numbers, it's possible, why not? 
So I tried to use the generator on a bigger sample with

```scheme
(let ((ht (make-hash-table)))
  (dotimes (i 1_000_000)
           (let ((r (random-integer 10)))
             (hash-table-set! ht r (+ 1 (hash-table-get ht r 0)))))
  (hash-table->list ht))
```
and it segfaults (not always with the same value of `i`).

Using a very big stack is OK (I used `--stack-size 100000000`)
It seems that this code use a lot of stack frames.  Two hours trying to find where the problem without success :disapointed:
As a consequence, I have not yet merged your code. Of course, I have no problem with this loop and the current implementation of SRFI-27.

Please, can you try to see where the comes from, because I'm out of idea....

jpellegrini:
Oh, I'm sorry!
I'll check that and get back to this.

jpellegrini:
I found it. instead of passing a random STATE to `%random-integer-from-source-mt`I was passing a random SOURCE (the procedure name is also misleading...)

jpellegrini:
I have just pushed another commit.

I think it's easier to keep the `srfi-27.stk` name for the file. We can make that better in the future.
This one should pass the tests and produce actually pseudo random numbers (the other was reading part of the code as if it were the state of the generator!) Anyway, only `random-integer` and `random-real` were broken. The closure-generating procedures,`random-source-make-{integers, reals}` was fine.

jpellegrini:
> ```scheme
> (let ((ht (make-hash-table)))
>   (dotimes (i 1_000_000)
>            (let ((r (random-integer 10)))
>              (hash-table-set! ht r (+ 1 (hash-table-get ht r 0)))))
>   (hash-table->list ht))
> ```

The result now is

```
((8 . 100221) (2 . 99673) (0 . 100197) (3 . 100310) (9 . 99462) (7 . 99861) (5 . 100010) (4 . 100023) (1 . 100395) (6 . 99848))
```
 :)

jpellegrini:
I have reverted to using the standard naming scheme... But now, while I can use SRFI-27 perfectly, it is not loaded during tests.
I even tried including `(require "srfi-27")` in teh test file, but it doesn't work! :confused:

jpellegrini:
What is happening?

Calling the REPL from command line:

```
$ stklos
  \    STklos version 1.60  (Id: 83d31b1)
   \   Copyright (C) 1999-2021 Erick Gallesio <eg@unice.fr>
  / \  Université Côte d'Azur
 /   \ [Linux-5.10.0-6-amd64-x86_64/pthread/readline/utf8]
stklos> (require "srfi-27")
"srfi-27"
stklos>  (find-module 'SRFI-27)
#[module SRFI-27]
stklos> 
```

Ok. So I made the test file for SRFI-27 print the current module, and `(find-module `SRFI-27)`:

```
(require "srfi-27")

(format #t "module ~a~%" (current-module))
(format #t "module 2 ~a~%" (find-module 'SRFI-27))

(test "srfi-27 random-source? 1"
      #t
      (random-source? default-random-source))

...
```

Now, during tests:

```
$ cd tests
$  ../src/stklos --no-init-file --utf8-encoding=yes
  \    STklos version 1.60  (Id: 83d31b1)
   \   Copyright (C) 1999-2021 Erick Gallesio <eg@unice.fr>
  / \  Université Côte d'Azur
 /   \ [Linux-5.10.0-6-amd64-x86_64/pthread/readline/utf8]
stklos> (load-path `("../lib"
             ,@(glob "../lib/srfi-[0-9]*/")
             ,@(load-path)))

(load "./test.stk")
stklos>   (test-init "TEST.LOG")
stklos>   (load "test-number.stk")
  (load "test-keyword.stk")
  (load "test-callcc.stk")
  (load "test-sport.stk")
  (load "test-fport.stk")
  (load "test-threads.stk")
  (load "test-srfi.stk")
==== Testing Numbers  ...                                        passed
==== Testing Keywords  ...                                       passed
==== Testing call/cc & dynamic-wind  ...                         passed
==== Testing String Ports (SRFI 6)  ...                          passed
==== Testing File Ports  ...                                     passed
==== Testing Threads  ...                                        passed
==== Testing SRFIs  ...                                          
      1    2    4    5   13   25   27module #[module stklos]
**** Error:
find-module: module with name `SRFI-27' does not exist
```

How come? The two module printing lines are right after the `require`, and  I have verified from the REPL that the require works!

egallesio:
All problems resolved I think except a new one. 
In your code you use the function  mpz_import which is not implemented in the gmp lignt version included in STklos. That means that we cannot compile STklos if we don't have gmp installed. Is it possible to avoid the usage of this function?

jpellegrini:
> In your code you use the function mpz_import which is not implemented in the gmp lignt version included in STklos. That means that we cannot compile STklos if we don't have gmp installed. Is it possible to avoid the usage of this function?

Oops! I didn't realize it was the light version that was included. I'll get that fixed!

jpellegrini:
> gmp lignt version included in STklos

@egallesio -- I see that this is a version extracted from MPI. But the links to the original library are broken:

* `README` points to http://spinning-yarns.org/michael/mpi which seems to redirect to a list of statistics books (the author didn't renew and hence lost the domain, maybe?)
* `mpi.h` points to  http://www.dartmouth.edu/~sting/ says page not found.

What is the URL for the library?

Also, the version of `mpi.h` included seems to have a date of 2004. Is that correct?

jpellegrini:
> All problems resolved I think except a new one.

Hi @egallesio !
I'm not sure what happens. I still get the errors I mentioned during tests.
I'll try a new checkout from git to see if it helps

egallesio:
I suppose this because you have an old srfi-27.ostk in youe path (I had this problem, with all the different rewritings).
The file that must be loaded is "srfi-27.so".

jpellegrini:
> I suppose this because you have an old srfi-27.ostk in youe path (I had this problem, with all the different rewritings).

YES! :) :) Now it passes all tests.

jpellegrini:
About the GMP, I see that in `gmp/README` there is a note:

The rewriting was necessary because old GMPlite was hard to maintain
(it embeds asm code and is hard to understand). On the contrary MPI,
is simpler (it is basically one source file).

There is a "mini-gmp" that is shipped with libgmp that is a single C file, and as far as I can see, uses no assembly. It has more features than the one which is being used, and is likely much faster also, and it would probably make the maintenance of a wrapper unnecessary. What about using it?

I ask because with the current version we'd need to adapt a lot (we need to use bignum exponentiation for example), and also, the included GMP wrapper has a limite on the string buffer size.

jpellegrini:
> There is a "mini-gmp" that is shipped with libgmp that is a single C file

In the "mini-gmp" sbdirectory of their repository:

https://gmplib.org/devel/repo-usage

They explicitly say you shouldn't even "build" it; just include the C file where you need it... Looks nice.

What do you think?

jpellegrini:
I have tested the mini-gmp and it seems to work fine. But I still need to work on it before making a PR -- assuming you are willing to change (I think it'd be nice not having to maintain a wrapper, and get good code from the GMP people!)

jpellegrini:
Oh. Here's something I wasn't expecting:

|                    | size | lib .so size | time |
|-----------------|-------|---------------|-------|
|system gmp | 713K | 513K      |  6.9s |
|mini-gmp  (which I suggested)    | 769K | 0        |  27s |
|mpi (currently included with STklos in `gmp/`)   | 737K | 0             |    8s |

So it seems that the mimi-gmp I have been talking about is not really that great of an alternative... I already knew that mini-gmp was slower than the original GMP, but... The already provided MPI is smaller and faster!

I'll then see if I can modify the provided bignum library that comes with STklos. Is that ok?


This is the code I used for the time measurement. Not really a benchmark, but it does help...
```scheme
(let ((s (expt 2 1000)))
  (time
   (dotimes (i 50_000)
     (set! s (+ s (expt 2 i)))))
  s)
```

egallesio:
Hi @jpellegrini ,

Sorry for my latency. I had a big exam to prepare... 

Using mini-gmp is a good idea. I didn't know that GMP provides now it. Thanks for the information 

I've seen your benchmark, and yes it is a bit disjointing. Anyway, I think it is a better solution that my wrapper around MPI
    -  it is slow, but it covers a bigger surface than the previous wrapper. It will be probably less boring to implement  code with bignums in C (as you have done recently and for the codes you already before such as new hashes. It will also permit to avoid some of the  bypassing you had to do to be compatible with GMP/MPI.
    - if someone need speed, she/he can install the real GMP. And GMP, is now really heavily used (this was not the case in 2009, when I wrote the MPI wrapper) It is even now a dependency of `coreutils`. So it is probably installed on all the reasonable GNU/Linux  "big" installations (OK, perhaps  not on routers :wink:)
 

> PS: If you'd like to see what I have done to integrate the mini-gmp (which doesn't seem to be worth the effort after all), it's in the mini-gmp branch in my Github fork of STklos. I checked that CFLAGS included -O2 when compiling libgmp.a. Without it it was even slower (92s).

Too bad, I have also done something on may side :-1: before I saw this message. I'll commit it, just afer sending this message.

PR #214, opened 2021-05-12
Fix non-threaded build of SRFI 25

jpellegrini:
Oops... The non-threaded build was broken. Sorry.


jpellegrini:
@egallesio already applied manually

PR #218, opened 2021-05-15
Fixes to documentation

jpellegrini:
Remove duplication in `HACKING.md`;
Mention MELPA in `README.md`


egallesio:
Thanks for these corrections.

BTW, I will try to have a look to Geiser now that it Geiser-STklos is in MELPA For now,  I still work with my own mode that I have written a long time ago and which is very minimal.

jpellegrini:
> BTW, I will try to have a look to Geiser now that it Geiser-STklos is in MELPA For now, I still work with my own mode that I have written a long time ago and which is very minimal.

Geiser was designed to feel as much as possible as SLIME... But of course, it's not by any means as featureful as SLIME!

PR #221, opened 2021-05-17
NaN flonums are not finite [ R7RS ]

jpellegrini:
Hi @egallesio !
Currently STklos returns `#t` for `(finite? +nan.0)`, but R7RS says that

_"The finite? procedure returns #t on all real numbers except +inf.0, -inf.0, and +nan.0, ..."_

This PR changes `FINITE_REALP` so as to be R7RS compliant!

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -49,7 +49,7 @@ static int use_srfi_169 = 1; /* do we allow the use of underscores in numbers? *
 static int real_precision = REAL_FORMAT_SIZE;
 static unsigned int log10_maxint;
 
-#define FINITE_REALP(n) ((REAL_VAL(n) != minus_inf) && (REAL_VAL(n) != plus_inf))
+#define FINITE_REALP(n) isfinite(REAL_VAL(n))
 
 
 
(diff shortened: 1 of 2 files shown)


lassik:
May be even better to use [`isfinite()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/isfinite.html) from from the C library.

lassik:
`isnormal()` is like `isfinite()` but also checks that the number is not subnormal. R7RS doesn't say anything about subnormal numbers.

jpellegrini:
> `isnormal()` is like `isfinite()` but also checks that the number is not subnormal.

Yes, but checking for normaliity would exclude zero. `isfinite` is the correct one.

jpellegrini:
Hm, there is still some more to check.
NaN is neither finite nor infinite, so `(infinite? +nan.0)` should return `#f`, and with this PR it returns `#t`...
I'll get this fixed.
More later.

jpellegrini:
Ok, it seems to be correct now, at the cost of some more code (`finite?` and `infinite?` cannot be boolean complementary).

Maybe there is a way to rewrite that in a more elegant way.

lassik:
Wow, you're right. 0.0 is subnormal.

Your changes look good to me. I can't think of a way to write them better.

egallesio:
Thanks for the correction. 
Merged.

jpellegrini:
Hi @egallesio !
I think you just closed, and didn't merge (?)

egallesio:
Yep. I was trying to see why i don't see it on my local tree.... :rofl:

egallesio:
Yep/ This explains why I don't see it on my local tree when doing a git pull :rofl:

Issue #220, opened 2021-05-17
Should numerator always be non-negative?


jpellegrini:
R7RS says that `denominator` should always return a non-negative number. _It says nothing about `numerator`._

For `(numerator -2/3)`: Chez, Chibi, Chicken, Gambit, Gauche, Guile, Kawa, Loko, MIT Scheme and Racket  all return -2.

Should STklos perhaps also return -2? (Currently it returns 2).

lassik:
If all other major Scheme implementations return a negative number, that's a pretty strong precedent that STklos should as well.

egallesio:
> If all other major Scheme implementations return a negative number, that's a pretty strong precedent that STklos should as well.

and one can expect that dividing one by the other produces a result with a  good sign!

I pushed a commit that correct this. Thanks @jpellegrini for signaling it.

PR #222, opened 2021-05-17
Add SRFI-144: Flonums

jpellegrini:
@egallesio  here's SRFI 144, Flonums!

I have made some simple tests, and it doesn't seem that having basic math flonum operators like `fl+` and such are worthwile -- there is no speedup (probably the flonum implementation in STklos is already efficient enough that the cost of a procedure call is large compared to the speedup that we'd have).

However, this one also brings several mathematical constants and other functions, and it's part of R7RS large, so I thought it would be worth it.

Issues:
* Two tests fail because `numerator` returns a positive number for negative arguments (see #220). If this behaviour is not supposed to change, I can change this PR to take that into account;
* One test fails because `(finite? +nan.0)` currently returns true (easy to fix -- see PR #221)
* Two tests fail because `(* 0.5 (expt 2 -1073))` is zero, and it should be `fl-least`  ( `4.94065645841247e-324` ).

And that's it. The tests are by Will Clinger (so there is a ton of them!), and we know exactly why those 5 fail.
I know how to fix the first two items; the third is a mystery to me (the "`(* 0.5 (expt 2 -1073))`" is zero problem).


jpellegrini:
I forgot one thing!
This SRFI is using the same configure option to control parameters that is used for fixnums. We should probably change that...

jpellegrini:
> This SRFI is using the same configure option to control parameters that is used for fixnums. We should probably change that...

I pushed one more commit that unifies the control of fixnum/flonum parameters in one single option... Tell me if this is OK @egallesio !

jpellegrini:
> * Two tests fail because `(* 0.5 (expt 2 -1073))` is zero, and it should be `fl-least`  ( `4.94065645841247e-324` ).

* Chibi and Loko also retrun 0.0
* Bigloo, Chicken, Cyclone, Gambit, Gauche, Guile, Kawa, MIT Scheme, Racket, Sagittarius return 4e-324 or 4.9406564584125e-324

jpellegrini:
Maybe `rational2double` in `number.c` could be improved?

jpellegrini:
It works. More precise, but slower... I'll make a PR, and you decide if it's worth it!

jpellegrini:
> It works

I mean, `(* -0.5 (expt 2 -1073))` gives the correct result, but `(* -0.5 (expt 2.0 -1073))` (with inexact 2) does not.

Even after changing `rational2double`,

```
stklos>  (expt 2 -1073)
1/101201126653655309176247673359458653524778324882071059178450679013715169783997673445980191850718562247593538932158405955694904368692896738433506699970369254960758712138283180682233453871046608170619883839236372534281003741712346349309051677824579778170405028256179384776166707307615251266093163754323003131653853870546747392
stklos> (expt 2.0 -1073)
0.0
```

jpellegrini:
I did make a pr for the first case (`exact->inexact` for rationals) -- #223 , but not the second (`expt`). @egallesio what's the preferred course of action here?

jpellegrini:
Sorry @egallesio I force-pushed, but I had made a small mess here. I hope you were not yet using this brnach.

Anyway, we'd need to decide what to do with the `expt` issue before this could be considered for inclusion... Tell me later.

egallesio:
Hi, 

I think that this should be OK now. I have tried to merge it but I have not seen the tests you talked about (there is no `tests/srfi/144.stk` file. Have you one?

jpellegrini:
Ops... I'll add a commit.

jpellegrini:
Ok. I had some problems here and had to force-push again, but it seems to work.

There are only two tests failing:

* `test srfi-144  on (map (lambda (x) #f) weird) expected (#f #f #t) but got (#f #f #f)`
  This one would be fixed by #221 
* `test srfi-144  on ints expected (-20.0 -8.0 -2.0 ... but got (-20.0 -8.0 -2.0 ...`
  This should be fixed now that numerator and denominator have the same sign. And the two lists shown in the tests do look the same... So there's still some little detail there to check.

egallesio:
Both errors seem to remain with the current version:disappointed:

jpellegrini:
> Both errors seem to remain with the current version

Hm, I was expecting the first one to disappear. I'll investigate.

jpellegrini:
> Both errors seem to remain with the current version

Ah, yes. I sent a PR fixing it in STklos, but not in SRFI-144, which was the original motivation...
I have pushed again, this version passes the first test. I'll check the second

jpellegrini:
Ah,this is why the second test fails:

```
stklos> (numerator 1e+23)
1e+23
stklos> (= (numerator 1e+23) 1e+23)
#f                                      ;; <== wasn't expecting this!!!
stklos> (= 1e23 1e23)
#t
stklos> (= 1e+23 1e+23)
#t
```

Hm, very strange. I'll look into it.

jpellegrini:
That is, precision is being lost here:
```
    case tc_real:     return
                        exact2inexact(STk_numerator(inexact2exact(q)));
```
in primitive `numerator`, in `number.c`.

jpellegrini:
But eve fixing that, the problem persists... I have changed `numerator` to do this:

```
  static double unused_argument;
  switch (TYPEOF(q)) {
    case tc_real:     {
        double frac = modf(REAL_VAL(q), &unused_argument);
                        if (FP_ZERO == fpclassify(frac)) { fprintf(stderr,"GOT IT %g\n", frac); return q; }
                        else return exact2inexact(STk_numerator(inexact2exact(q)));
                      }
```

And it seems that it would work:

```
stklos> (= (numerator 1e+23) 1e+23)
GOT IT 0
#t
```

However,
```
stklos>
   (define posints (map flonum '(1 2 3 4 5 10 65536 1e23)))
   (define nats (cons (flonum 0) posints))
   (define ints (append (map flonum '(-20 -8 -2 -1)) nats))
   (define posfracs (map flonum '(1/1000 1/10 1/3 1/2)))
stklos> (map flnumerator ints)
(-20.0 -8.0 -2.0 -1.0 0.0 1.0 2.0 3.0 4.0 5.0 10.0 65536.0 1e+23)
stklos> (map = (map flnumerator ints) ints)
(#t #t #t #t #t #t #t #t #t #t #t #t #f)
stklos> ints
(-20.0 -8.0 -2.0 -1.0 0.0 1.0 2.0 3.0 4.0 5.0 10.0 65536.0 1e+23)
stklos> (car (reverse ints))
1e+23
stklos> (flnumerator (car (reverse ints)))
1e+23
stklos> (= (flnumerator (car (reverse ints))) 1e+23)
#f
```

Hm, so something is still wrong.

jpellegrini:
Ok, so it was easier than I thought. I fixed `flnumerator` and it works!  All tests pass!!!
(At least on my desktop machine here)

egallesio:
I have made small adaptations for the new way to make mixed-modules. I hope that every thing is OK (at least the tests pass, and they are numerous :wink: )
Thanks.

PR #223, opened 2021-05-17
Be more precise when turning rationals into doubles

jpellegrini:
`(inexact (expt 2 -1073))` should be 4.94065645841247e-324,
but STklos returns zero. This patch makes the process of
turning rationals into doubles more precise, and STklos
will now return 4.94065645841247e-324.

This helps with the tests for SRFI-144 (PR #222 ), but I think it's an enhancement per se.
Unless we're willling to sacrifice some precision for speed (and I'd say this patch probably makes the whole process a lot slower, since we'll be using bignum division (`mpq_div`) instead of native double division).

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -472,7 +472,15 @@ static double rational2double(SCM r)
 
   switch (convert(&num, &den)) {
     case tc_integer: return ((double) INT_VAL(num)) / ((double) INT_VAL(den));
-    case tc_bignum:  return scheme_bignum2double(num) / scheme_bignum2double(den);
+    case tc_bignum: {
+        mpq_t a, b;
+        mpq_init(a);
+        mpq_init(b);
+        mpq_set_z(a,BIGNUM_VAL(num));
+        mpq_set_z(b,BIGNUM_VAL(den));
+        mpq_div(a,a,b);
+        return mpq_get_d(a);
+    }
     default:         STk_panic("bad rational ~S", r);
   }
   return 0.0; /* never reached */


jpellegrini:
But it uses `mpq`, which is not present in the provided gmp. I'll work on that.

jpellegrini:
> But it uses `mpq`, which is not present in the provided gmp. I'll work on that.

Done.

jpellegrini:
@egallesio now that the mini-gmp is used, we may not need the `#ifdef`... I'll fix that

jpellegrini:
Ok, removed the ifdef

jpellegrini:
Ops... I was wrong! It doesn't compile. I thought the mini-gmp also had a mpq feature (in another file?)

jpellegrini:
> Ops... I was wrong! It doesn't compile. I thought the mini-gmp also had a mpq feature (in another file?)

(I know I said it was nice that it had a single C file, but I think it has rationals in a separate file)

If not, we could leave it as it is (not do the extra precision). Chibi doesn't seem to do that either.
Then we may adapt the tests of SRFI 144 to exclude this one, and include a note there explaining the reason.

jpellegrini:
> If not, we could leave it as it is (not do the extra precision). Chibi doesn't seem to do that either.
> Then we may adapt the tests of SRFI 144 to exclude this one, and include a note there explaining the reason.

I meant for the mini-GMP case only. When system GMP is available, we can use mpq types

egallesio:
Hi @jpellegrini,
I was  a bit reluctant to use the mpq library in STklos (it is redundant with the code in `number.c`, which was written before GMP included rationals). Furthermore, having a different result depending on the fact we use or not the real  or mini gmp is not suitable imho (we could of course  add the `mini-mpq.c` in the  `./gmp/libgmp.a` library, but this add a lot of code for just one function). 

So I adapted the code of `mpq_get_d` found in `mini-mpq.c` (only a few lines of code) and :tada: 

```scheme
stklos> (* 0.5 (expt 2 -1073))
4.94065645841247e-324
stklos> 
```
both with standard and provided GMP.

jpellegrini:
That's great @egallesio !
Is it a change that could be useful to mini-gmp per se? We could send it upstream.

jpellegrini:
> We could send it upstream.

Oh, sorry! I didn't understand what you meant. You didn't change mini-libc, you got code from it and incorporated.
Next time I'll read *after* coffee.

PR #226, opened 2021-05-18
Do not use mpz_import when using provided bignums

jpellegrini:
Hi @egallesio !

It's fixed!

This patch works around that problem...

As to the mini-gmp I had told you about... Well, depending on the usage, it may be faster or slower.

* For generating random integers, mini-gmp is faster (almost as fast as system gmp)
* For usual bignum arithmetic, the STklos-included MPI-MGP is faster

I'm not sure I have a preference. mini-gmp would free us from maintaining a wrapper, but may be generally slow.

Anyway, there are two alternatives:

* This PR has a patch to SRFI-27 *only* (does not change STklos' lib for bignums)
* My "mini-gmp" branch does *not* touch SRFI-27, but replaces the bignum-lib with mini-gmp. If you want I may make a PR for that one too, then you pick one.

### DIFF
--- a/gmp/gmp.c
+++ b/gmp/gmp.c
@@ -38,7 +38,7 @@
   with 32 bits (conversion to/from string). This is slow, but it works.
 
   We know that we have a 64 bits value when the value is ouside the
-  range [MIN32_SI, MAX32_SI]. In this case, we pass by the deciaml
+  range [MIN32_SI, MAX32_SI]. In this case, we pass by the decimal
   representation of the number in a string
  */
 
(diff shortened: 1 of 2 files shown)


egallesio:
Finally, we will not need this workaround with new GMP. I apologize for the work you had done ad which is no more necessary. 
But, this is a good illustration, that it is preferable to be as compatible as possible with GMP, thanks to **mini-gmp**, even if it is not so efficient.

jpellegrini:
Yay!
Problem solved!!! :)

PR #229, opened 2021-05-26
Don't exit with nonzero status for "-h" option

jpellegrini:
If the user asked for help, we don't need to exit with -1 status. The intended action (help) was performed successfully.

### DIFF
--- a/utils/stklos-compile.stk
+++ b/utils/stklos-compile.stk
@@ -163,7 +163,7 @@
      (("no-time" :help "Don't display compilation time")
         (compiler:time-display #f))
      (("help" :alternate "h" :help "This help")
-        (arg-usage (current-error-port))
+        (arg-usage (current-output-port))
         (exit 0))
 
      (else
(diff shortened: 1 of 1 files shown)


lassik:
`-h` and `--help` should also write the usage message to stdout, so that it can be grepped for example.

When the usage is shown due to invalid command line syntax, then it should be written to stderr (as it currently is).

The usual pattern is something like this:

```Scheme
(define (usage status port)
  (write-string ... port)
  (exit status))

(usage 0 (current-output-port))  ; -h or --help
(usage 1 (current-error-port))   ; invalid syntax (exit code is not standardized, but should be non-zero)
```

jpellegrini:
> `-h` and `--help` should also write the usage message to stdout, so that it can be grepped for example.

Yes, good catch!
I'e added another commit. Output goes to stderr when a wrong argument is passed, but to stdout if the user requested help.

egallesio:
Thanks to both of you.
Merged.

PR #230, opened 2021-05-26
Fix stdout/stderr and exit code 0/1 in usage

lassik:
Fixes the main `stklos` command in the same way as #229 fixes other commands.


egallesio:
Thanks @lassik.
Merged.

PR #233, opened 2021-05-29
Add SRFI-221: Generator/accumulator sub-library

jpellegrini:
Hello @egallesio !
SRFI-221 was finalized today, so I ported it.

However, something isn't working, I can't tell why:

SRFi-221 uses SRFI-41. So this is in thebeginning of srfi-221.stk:

```
(require "srfi-41")
(require "srfi-158")

(define-module SRFI-221
  (import SRFI-41)
  (import SRFI-158)
  
  (export
    accumulate-generated-values
    genumerate
    gcompose-right
    gcompose-left
    gchoice
    generator->stream
    stream->generator)

```

But then,

```
stklos> (require "srfi-221")
"srfi-221"
stklos> (define g-expect (circular-generator 1 2 3))
;; g-expect
stklos> (define g-actual (stream->generator (stream-constant 1 2 3)))
;; g-actual
stklos> (generator->list (gtake g-expect 20))
(1 2 3 1 2 3 1 2 3 1 2 3 1 2 3 1 2 3 1 2)
stklos> (generator->list (gtake g-actual 20))
**** Error:
g-actual: variable `stream-null?' unbound
stklos> stream-null?
#[closure stream-null?]            ;; see, srfi-41 does export that! but srfi-221 procedures can't see it, why?
stklos> (stream-null? 1)
#f
```

What could be happening?

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -110,4 +110,5 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-216: SICP Prerequisites (Portable)
     - SRFI-217: Integer Sets
     - SRFI-219: Define higher-order lambda
+    - SRFI-221: Generator/accumulator sub-library
     - SRFI-223: Generalized binary search procedures
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -138,6 +138,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-207.stk        \
           srfi-216.stk        \
           srfi-217.stk        \
+          srfi-221.stk        \
           srfi-223.stk        \
           trie.stk            \
           tar.stk             \
@@ -216,6 +217,7 @@ scheme_OBJS = compfile.ostk     \
           srfi-207.ostk         \
           srfi-216.ostk         \
           srfi-217.ostk         \
+          srfi-221.ostk         \
           srfi-223.ostk         \
           trie.ostk             \
           tar.ostk              \
--- a/lib/srfi-221.stk
+++ b/lib/srfi-221.stk
@@ -1,0 +1,178 @@
+;;;;
+;;;; srfi-221.stk	        -- Implementation of SRFI-221
+;;;;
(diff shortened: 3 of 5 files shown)


jpellegrini:
Is SRFI-41 implemented the correct way?

```
(require "streams-primitive")
(require "streams-derived")

(define-module SRFI-41
  (import STREAMS-PRIMITIVE STREAMS-DERIVED)
  (export stream-null stream-cons stream? stream-null? stream-pair? stream-car
          stream-cdr stream-lambda define-stream list->stream port->stream stream
          stream->list stream-append stream-concat stream-constant stream-drop
          stream-drop-while stream-filter stream-fold stream-for-each stream-from
          stream-iterate stream-length stream-let stream-map stream-match _
          stream-of stream-range stream-ref stream-reverse stream-scan stream-take
          stream-take-while stream-unfold stream-unfolds stream-zip)
  (export-syntax stream-cons stream-lambda stream-delay stream-lazy
                 define-stream stream stream-let stream-match stream-of)
) ;; END OF DEFINE-MODULE
;;;; ======================================================================
(select-module STklos)

(import STREAMS-PRIMITIVE STREAMS-DERIVED SRFI-41)
(provide "srfi-41")
```

jpellegrini:
Just a thought: `stream-null?` uses `stream-delay`, which is a macro; maybe I didn't do the right thing to export the macros from `STREAM-PRIMITIVE`?

jpellegrini:
I understand what happens. If a module imports names, and then exports it, the procedure bindings are lost?

```scheme
(define-module A
  (export x)
  (define x 3.1415))

(define-module B
  (import A)
  (export x y)
  (define y (sqrt 2)))

(import B)

y  => 1.4142135623731

x
**** Error:
%execute: variable `x' unbound
```

That's what happened.

There's no "reexport" I suppose... :)

egallesio:
Hi @jpellegrini ,

Thanks for this SRFI. Sorry for being so silent and so long to merge this one.

jpellegrini:
> Sorry for being so silent and so long to merge this one.

No problem -- I am sorry for not contributing as much as I'd like.

PR #235, opened 2021-06-07
Resolve conflict in Makefile.am

jpellegrini:
Some tests were mysteriously failing... Because Makefile.am had an unresolved conflict!


jpellegrini:
This was also preventing autoreconf to run...

jpellegrini:
Ops. I had included SRFi-27 in the wrong makefile. Fixed.

egallesio:
Opps I also came across it and I  corrected it in a previous commit.

PR #237, opened 2021-06-14
Add instructions to install on Android emulator

jpellegrini:
Hi @egallesio !

It is also possible to run STklos on an Android emulator on x86 architectures!
I've written instructions -- not sure if they're worth including in STklos, but anyway, here's the PR!


egallesio:
Hi @jpellegrini,

I have merged this text into `INSTALL-Android.md` (to avoid having too many files at the top of the directory tree). 
I have also added a note about using  an x86 ISO of Android OS. 

So, yours text is merged, even if this specific PR is not. 
Thanks.

PR #240, opened 2021-07-27
Add docs to define-class, find-class and is-a?

jpellegrini:
Add online documentation for these, so they'll be available on the online help.

Will do `define-method` later.


egallesio:
Cool. It is a good idea to make it available to the online help. 
Merged.

PR #242, opened 2021-07-28
Numerics: don't try to call double2rational on NaNs; use long to represent fixnums

jpellegrini:
In number.c, NaNs would enter the tc_real case, and double2rational would then enter an infinite loop. This commit adds a guard for that.


jpellegrini:
This would need to check for infinities also. Will do later

jpellegrini:
Okay, it's correct now. Protects against calling `exact` in nan's and infinities.

jpellegrini:
Also fixing a second problem (`sqrt` would return nan because it used `ints` to represent fixnums).

egallesio:
Thanks for this PR. 
Merged.

PR #243, opened 2021-07-28
Add SRFI-223: Generalized binary search procedures

jpellegrini:
Hi @egallesio ! One more SRFI...

The tests for this one depend on PR #242 (otherwise they'll enter a loop and never finish).


egallesio:
One more SRFI! Thanks @jpellegrini
Merged!

PR #244, opened 2021-07-28
Adds SRFI-222: Compound objects

jpellegrini:
Here's SRFI-222! :)


lassik:
SRFI 222 is more abstract or generic than most SRFIs. Currently the main use case is that a Scheme implementation can be compatible with the R6RS exception hierarchy. Does STklos have such compatibility as a goal?

jpellegrini:
> Does STklos have such compatibility as a goal?

I don't know if @egallesio has such plans but I'd guess not (conditions are a native Scheme type implemented in C, in `src/cond.c`)
I made the PR because it may perhaps be useful in other situations, and because it was easy to add :)

PR #246, opened 2021-08-06
Clean up bits of SRFI-25

jpellegrini:
No big deal; there was a macro which was defined twice, and a variable that served no purpose.
Passes the tests.


egallesio:
Merged.

PR #248, opened 2021-08-11
Optimize do loops

jpellegrini:
# Description and remaining issue

Hi @egallesio --
I thought I'd try to optimize the translation of `do` forms, so they would not include a procedure call (and use goto instead).
So... It seems to work, it generates much shorter and faster code, but... It seems to have some issues that I don't understand -- it doesn't pass the tests, there seems to be some interaction with threads:

```
==== Testing Threads  ...                                        failed
Errors found in this section:
test thread-join! on (mt-fib 35) expected 14930352 but got #(fail)
==== Testing SRFIs  ...                                          
      1    2    4    5   13   25**** Error while executing file "do-test.stk"
```

If I enter the `mt-fib` procedure from `tests/test-threads.stk` and then call `(mt-fib 35)`, I get this:

```
stklos>  (mt-fib 35)
**** Error (in thread "thread189"):
thread-join!: cannot join on myself (deadlock will occur)
    (this error may be signaled again later)
**** Error:
thread-join!: cannot join on myself (deadlock will occur)
```

What did I miss?

Below is the explanation of what I've done.

# Optimize 'do' loops

Do not translate do into letrec, use goto instead

```
 (do inits test body)

 --- is rewritten to --->

 (let inits
   (%%label L)
   (if test
       result
       (begin body
              (set! var-1 update-1)
              (set! var-2 update-2)
              ...
              (%%goto L))))
```

This saves a procedure call in each iteration, and generates less
opcodes. The speedup is considerable:

```
(time
 (let ((sum 0))
   (do ((i 0 (+ i 1)))
       ((= i 100_000_000) sum)
     (set! sum (+ sum 1)))))
```

From 7221 with `LETREC` to 5614 with `GOTO`.
(NOT a real benchmark; just a quick test -- I've ran 10 times that
code with each do-translator, and those numbers are the average.

The number of opcodes also went from 32 to 20.

This commit also adds sanity checks to the do form:

* too many arguments in each binding (only <2 was checked before)

* number of elements in the test-and-result form

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -1437,9 +1437,9 @@ doc>
 
   ;; rewrite the body
   (let ((loop-name (new-label)))
-    `(let ,(map (lambda (init) (if (= (length init) 2)
-                              init
-                              (reverse (cdr (reverse init))))) inits)
+    `(letrec* ,(map (lambda (init) (if (= (length init) 2)
+                               init
+                               (reverse (cdr (reverse init))))) inits)
        (%%label ,loop-name)
        (if ,(car test)
            (begin ,@(if (null? (cdr test))
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -1395,27 +1395,64 @@ doc>
 ;;
 ;; DO
 ;;
-(define (rewrite-do inits test body)
-  (let ((loop-name (gensym)))
-    `(letrec ((,loop-name
-       (lambda ,(map car inits)
-         (if ,(car test)
-             (begin ,@(if (null? (cdr test))
-                          (list (void))
-                          (cdr test)))
-             (begin ,@body
-                    (,loop-name ,@(map (lambda (init)
-                                         (if (< (length init) 2)
-                                             (compiler-error 'do
-                                                              init
-                                                              "bad binding ~S"
-                                                              init)
-                                             (if (null? (cddr init))
-                                                 (car init)
-                                                 (caddr init))))
-                                       inits)))))))
-       (,loop-name ,@(map cadr inits)))))
 
+;; Rewrite do loops
+;;
+;; (do inits test body)
+;; ->
+;; (let inits
+;;   (%%label L)
+;;   (if test
+;;       result
+;;       (begin body
+;;              (set! var-1 update-1)
+;;              ...
+;;              (%%goto L))))
+;;
+;; * 
+;; Vars with no update form are not updated.
+;;
+(define (rewrite-do inits test body)
+  ;; check inits
+  (for-each (lambda (init)
+              (when (< (length init) 2)
+                (compiler-error 'do
+                                init
+                                "not enough forms in binding ~S"
+                                init))
+              (when (> (length init) 3)
+                (compiler-error 'do
+                                init
+                                "too many forms in binding ~S"
+                                init)))
+            inits)
+
+  ;; check test-and-result form
+  (when (or (null? test)
+            (> (length test) 2))
+    (compiler-error 'do
+                    test
+                    "bad test-and-result form"
+                    test))
+
+  ;; rewrite the body
+  (let ((loop-name (new-label)))
+    `(let ,(map (lambda (init) (if (= (length init) 2)
+                              init
+                              (reverse (cdr (reverse init))))) inits)
+       (%%label ,loop-name)
+       (if ,(car test)
+           (begin ,@(if (null? (cdr test))
+                        (list (void))
+                        (cdr test)))
+           (begin ,@body
+                  ,@(filter (lambda (x) x) ; remove the useless '#f's
+                            (map (lambda (init)
+                                   (if (= (length init) 3)
+                                       (list 'set! (car init) (caddr init))
+                                       #f)) ;; will be removed
+                                 inits))
+                  (%%goto ,loop-name))))))
 
 (define (compile-do e env tail?)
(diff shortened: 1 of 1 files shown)


jpellegrini:
It would also be possible to optimize named lets, but it's trickier. For example,
```
(let Loop ((x 0))
  (if (= x 10)
       Loop                     ;;  using Loop as first class entity
      (Loop (+ x 1))))
```

```
(let Loop ((x 0))
  (if (= x 10)
      x
      (Loop (Loop (+ x 1)))))  ;; nested Loop requires us to rename x
```

But... It's still possible. I'll work on that later.

jpellegrini:
Also, I should have used `let*`+`set!` in the initialization  instead of `let`, otherwise this (valid) expression won't work:

```
(do ( (f (lambda (x) (if (> x 0) x 0))) )
  (#t (f -3))
  4)
```

I'll update the PR later

jpellegrini:
I'm getting closer to understand why this `do` expansion doesn't work. 

See: this code,

```scheme
(define-syntax push!
  (syntax-rules ()
    ((_ e lst)
     (set! lst (cons e lst)))))

(let* ((thread-num 10)
       (t '())
       (a (make-vector thread-num -2)))
  (dotimes (i thread-num)
    (format #t "thread ~a~%" i)(flush (current-output-port))
    (push! (make-thread
            (lambda ()
              (format #t "my vector is ~a, i is ~a~%" a i)))
              ;(vector-set! a i (* 2 i))))
           t))
  (for-each thread-start! t)
  (for-each thread-join! t)
  a)
```

works fine in current STklos, producing this output:

```
thread 0
thread 1
thread 2
thread 3
thread 4
thread 5
thread 6
thread 7
thread 8
thread 9
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 9
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 8
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 7
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 6
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 5
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 4
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 3
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 2
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 1
my vector is #(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2), i is 0
#(-2 -2 -2 -2 -2 -2 -2 -2 -2 -2)
```

But with this PR, the variable `i` is always `10`. This is this expanded version of it (which (which is wrong):

```scheme
(let ((a (make-vector 10))
      (t '()))
  (letrec* ((i 0))
    (%%label 20)
    (if (>= i 10)
        (begin (void))
        (begin (format #t "thread ~a~%" i)
               (flush (current-output-port))
               (push! (make-thread
                       (lambda ()
                         (format #t "my vector is ~a, i is ~a~%" a i)))   ;; oops -  i here will be 10 when the thread starts!
                      t)
               (set! i (fx+ i 1))
               (%%goto 20))))
    (for-each thread-start! t)
  (for-each thread-join! t)
  a)
```

It will output:

```
thread 0
thread 1
thread 2
thread 3
thread 4
thread 5
thread 6
thread 7
thread 8
thread 9
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
my vector is #(#void #void #void #void #void #void #void #void #void #void), i is 10
#(#void #void #void #void #void #void #void #void #void #void)
```

So the problem is that the variable references in `push!` macro inside the loop are being expanded in the wrong way/time with this new `do` expansion.

jpellegrini:
Found it: R7RS says,

> variables, and then the iteration phase begins.
> Each iteration begins by evaluating `test`; if the result is
> false (see section 6.3), then the `command` expressions are
> evaluated in order for effect, the `step` expressions are 
> evaluated in some unspecified order,
>    **_the `variable`s are bound to fresh locations_**,
>  the results of the `step`s are stored in the
> bindings of the `variable`s, and the next iteration begins.

My transformation does not bind the variables to fresh locations - so it's not valid.

But it still feels like this could be optimized somehow...

jpellegrini:
If there are ways to optimize better, it's not with this PR, s I'm closing it.

egallesio:
I was working on `dotimes` and have optimized it. A 500_000_000  `dotimes` passes from 15s to 11s on my machine. I have not applied the same optimization for `do` loop. I'll do it asap.

PR #250, opened 2021-08-17
Print signature with procedure object

jpellegrini:
So, when compiler:generate-signature is true:

```
stklos> ((in-module STKLOS-COMPILER compiler:generate-signature) #t)
```

This would happen:

```
stklos> (define f (lambda (a b) (+ a b)))
stklos> f
 #[closure f (a b)]        ;; fixed arity

stklos> (define g (lambda args (apply + args)))
stklos> g
 #[closure g args]         ;; var arity

stklos> (define h (lambda (a b . c) c))
 stklos> h
 #[closure h (a b . c)]    ;; var arity II

stklos> abs
 #[primitive abs]          ;; signature not available
```

### DIFF
--- a/src/proc.c
+++ b/src/proc.c
@@ -62,12 +62,19 @@ SCM STk_make_closure(STk_instr *code, int size, int arity, SCM *cst, SCM env)
 }
 
 
-static void print_lambda(SCM closure, SCM port, int _UNUSED(mode))
+static void print_lambda(SCM closure, SCM port, int mode)
 {
   if (CLOSURE_NAME(closure) != STk_false)
-    STk_fprintf(port, "#[closure %s]", SYMBOL_PNAME(CLOSURE_NAME(closure)));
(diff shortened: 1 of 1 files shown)


lassik:
Great!

egallesio:
Merged. Thanks for this PR @jpellegrini 
BTW, fixed the test against `STk_nil` that should be against `STk_false` instead.

jpellegrini:
>  BTW, fixed the test against `STk_nil` that should be against `STk_false` instead.

Oops! Sorry about that. The funny thing is that... It worked for the tests I did :)

egallesio:
No problem. 
In fact, you can see the problem when loading a compiled file  without  integrated  formals, you had

```scheme
stklos> (require "srfi-1")
"srfi-1"
stklos> iota
#[closure iota #f]
stklos> 
```
Now it prints `#[closure iota]`

PR #251, opened 2021-08-18
Use + insted of fx+ in dotimes

jpellegrini:
Hi @egallesio !

`dotimes` uses `fx+`, but the documentation says the index is "an integer", and `dotimes` also uses `>=` later to check the exit condition, so I suppose it should be `+` instead of `fx+`.
If not, then I can make another PR changing `>=` to `fx>=?`, and changing the documentation.

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -871,7 +871,7 @@ doc>
            (let ((limit  (gensym))
                  (result (if (null? result) (list '(void)) result)))
              `(let ((,limit ,count))
-                (do ((,var 0 (fx+ ,var 1)))
+                (do ((,var 0 (+ ,var 1)))
                     ((>= ,var ,limit) ,@result)
                   ,@body))))
          bindings))


egallesio:
The problem with `fx>=' is that an expression like `(dotimes (i 9.9) (print i))` will yield an error. 
But that is true, that the documentation says that the expression must be an integer. :thinking:

jpellegrini:
Hmm. Ok - 
I know nobody would write this, but
```
(dotimes (i (inexact (expt 2 64))) 'whatever)
```
would fail with `fx+` complaining that `i` is not a fixnum, but only after trillions of iterations...

jpellegrini:
Maybe:
- document that the expression may be inexact, but should fix in a fixnum for comparison, and keep `fx+`?
- test the limit with `fixnum?`, and then use one of two loops, with either `fx+` or `+`?
- drop the fixnum approach entirely and just use `+`?

egallesio:
Yep, but On 32 bits machines, this is not too long to exhaust all the values.....

I have tried several versions of  `dotimes`:
```scheme
(define-macro (dotimes1 bindings . body)
  (apply (lambda (var count . result)
           (let ((limit  (gensym))
                 (result (if (null? result) (list '(void)) result)))
             `(let ((,limit ,count))
                (do ((,var 0 (+ ,var 1)))
                    ((>= ,var ,limit) ,@result)
                  ,@body))))
         bindings))


(define-macro (dotimes2 bindings . body)
  (apply (lambda (var count . result)
           (let ((limit   (gensym))
                 (fxlimit (gensym))
                 (result (if (null? result) (list '(void)) result)))
             `(let* ((,limit   ,count)
                     (,fxlimit (if (not (integer? ,limit))
                                   (error "count must be an integer")
                                   (if (real? ,limit)
                                       (exact (truncate ,limit))
                                       ,limit))))
                (do ((,var 0 (fx+ ,var 1)))
                    ((fx>=? ,var ,fxlimit) ,@result)
                  ,@body))))
         bindings))


(define-macro (dotimes3 bindings . body)
  (apply (lambda (var count . result)
           (let ((limit   (gensym))
                 (fxlimit (gensym))
                 (fx+     (gensym))
                 (fx>=?   (gensym))
                 (result (if (null? result) (list '(void)) result)))
             `(let* ((,limit   ,count)
                     (,fxlimit (if (not (integer? ,limit))
                                   (error "count must be an integer")
                                   (if (real? ,limit)
                                       (exact (truncate ,limit))
                                       ,limit)))
                     (,fx+      (if (bignum? ,fxlimit) fx+ +))
                     (,fx>=?    (if (bignum? ,fxlimit) fx>=? >=)))
                (do ((,var 0 (,fx+ ,var 1)))
                    ((,fx>=? ,var ,fxlimit) ,@result)
                  ,@body))))
         bindings))

(let ((count 500_000_000))
  (print (time (dotimes (i count) #void)))   ;; current implementation fx+ & >
  (print (time (dotimes1 (i count) #void)))  ;; use + and >=
  (print (time (dotimes2 (i count) #void)))  ;; use fx+ and fx>=? (don't test bignum)
  (print (time (dotimes3 (i count) #void)))) ;; use fx+ and fx>=? or + and >= if bignum
```

And the  interesting result is: **the more you try to be clever, the less you are efficient** (note that 500_000_000 is not bignum).

The most efficient is the current version, but this is very close to the version you propose.  The third version is slow because, `fx>=` is far less efficient than `>=`` which is optimized by the pephole optimizer when calld with two parameters only (something like 20% faster).  With 3 parameters, this is no more true.

The last version takes cares of  the upper limit that could be a bignum (it could have a meaning on 32 bits architectures). It is slow, because (I suppose the access) to local is slower than accessing a global. 

Wa could have two paths too, but this is not realistic, since body can be very big and would be duplcated if we have 2 paths. We could put the body in a thunk, but this will imply calling a function.... 

Interesting no? 

I suppose that using `+` and `>=` is probably a good solution since it is more general and nearly as efficient as the current version 

Note: I have changed the implementation of `dotimes` on Apr 26.

jpellegrini:
Ok, I'll change the PR then!

jpellegrini:
> Ok, I'll change the PR then!

Uh. Actually, it's already the way you proposed!

egallesio:
Yep, See my comment in PR #248.

jpellegrini:
> less efficient than `>=`` which is optimized by the pephole optimizer when calld with two parameters only (something like 20% faster). With 3 parameters, this is no more true.

I was thinking about that -- several inlined arithmetic operations could benefit from that, right? +, -, *, /, fx+, etc all have binary inlined versions and n-ary procedure-call versions. (I actually had this idea looking into the optimizations made by the Emacs Lisp bytecode compiler. They don't do it with a peephole optimizer, though -- they do code transformation, which is tricky in Scheme since you'd need to check if `+` is really bound to the original `+` procedure).

egallesio:
Hi @jeronimo,

I was wrong in my comment on PR #248 :disappointed: . In fact, I felt in the same problem you had with your optimization of do loops.  Without, calling a functions closures capture the wrong environment, as you seen.

jpellegrini:
But `dotimes` does not need to create new bindings for the variable, so I suppose it could be optimized (not `do`, only `dotimes`)...

egallesio:
Merged. 

I need to add some test for the `tagboody` implementation, and I'll commit it (with the  `repeat` form)

PR #253, opened 2021-08-26
Advertise SRFI-15 (fluid-let)

jpellegrini:
Although withdrawn, it is supported and even used by the STklos compiler!
Some other implementations do support it also (Chez, Loko, Chicken, Iron)...

But I see you may not want to advertise it since it's been withdrawn.


lassik:
See also: https://github.com/pre-srfi/fluids

egallesio:
Hi, 
I have merged this PR. At first, I thought it should not be in the list of supported SRFIs, since it was withdrawn, However, you are right: `fluid-let` is sufficiently important to be in this list.

PR #255, opened 2021-09-01
Add SRFI-132: Sort Libraries

jpellegrini:
Hello @egallesio !

Here's SRFI-132.

Some highlights:

* core of the SRFI (merging and sorting) written in C
* vector sorting uses [timsort](https://github.com/python/cpython/blob/main/Objects/listsort.txt)
  - Seems more efficient than "quicksort reverting to insertion when n<16", for both random and totally ordered vectors (I have done some basic experiments)
* destructive list sorting uses almost no extra memory (only one tiny vector and a small stack);
  Naive merge sort is too slow to be practical
* vector merging uses galloping (explained in the link about timsort above)

Remarks:

* `vector-stable-sort!` uses timsort , with most of the small optimizations.
* `list-stable-sort` runs `list->vector`, then `vector-stable-sort!`, then `vector->list`. Easy, and it's very fast.
* `list-stable-sort!` -- this one was tricky. I thought that if someone is using an in-place sorting procedure, then maybe we shouldn't waste memory, so I did not do the "copy to vector, sort, then copy to a new list" thing. I have implemented timsort for lists (works quite well, since timsort is a variant of mergesort) but it still uses O(n) extra memory in the worst case.
* for the unstable variants, I didn't use anything different. I have tried, for example, quicksort for vectors (reverting to shellsort when n<16, n<32), but it isn't much better than timsort, at least for the data sets I used. So the "unstable" procedures actually perform a stable sort!
* The SRFI lists some procedures related to partitioning, selection etc at the end. Since these are not used in our implementation of sorting and there wasn't anything that would really need to be very fast, I adapted Scheme code from Chibi.

Some timings:

I have done experiments with almost-sorted, strangely-distributed, sorted, and  random vectors and lists.
For the random vector case, I have some timings here, that I collected just to make sure that what I was doing was not too slow, and it seems that the result isn't bad.

(I already knew heapsort would perform bad, but I wanted to know how much).

sort a random VECTOR of length 10_000_000:

| system/algorithm | time (s) | remark |
| --- | --- | --- |
| STklos heapsort  (pure)       | 25.1 | |
| STklos quicksort (pure)           | 14.7 | |
| STklos quicksort (+shell when <16)  | 12.3 | |
| STklos quicksort (+shell when <32)  | 12.1 | |
| | | |
| STklos timsort (destructive)    | 12.5 |   |
| | | |
| Chibi  (destructive)      | 3.5 | fastest! |
| Chibi  (non-destructive)      | 4.4 | fastest! |
| Chicken (destructive)         | 27.5  | (REPL, interpreted) |
| Chicken (destructive)         | 22.8  | (compiled) |
| Chicken (non-destructive)         | 26.7  | (REPL, interpreted) |
| Chicken (non-destructive)         | 29.2  | (compiled) |
| Gambit  (destructive)      | 25.6 | |
| Gambit  (non-destructive)      | 25.6 | |
| Gauche  (destructive)  | 71.6 | |
| Gauche  (non-destructive)  | 70.2 | |
| Racket | 10.1 | `vector-sort`, not SRFI-132 |
| Sagittarius (destructive) |  56.7 | |
| Sagittarius (non-destructive) | 58.3 | |
| SBCL `sort` | 11.2|  REPL, not optimized |
| SBCL `stable-sort` | 4.2 | REPL, not optimized. stable faster than unstable sort! (for a random vector) |

The Chibi timing is impressive - it's not that much faster than others with complex comparison predicates like `(lambda (a b) ...)`, but even then it's faster than all others.

sort a random LIST of length 10_000_000:

| system/algorithm | time (s) | remark |
| --- | --- | --- |
| STklos mergesort   (find all runs first, then merge) | >1h | |
| | |
| STklos `list-stable-sort` (non-destructive) |       13 | |
| STklos timsort  (destructive)                          | 19 |  |
| | | |
| Chibi  (destructive)      | 4.1  | fastest! |
| Chibi  (non-destructive)      | 7.3  | fastest! |
| Chicken (destructive)         | 20.4  | (REPL, interpreted) |
| Chicken (non-destructive)         | 178.1  | (REPL, interpreted) |
| Chicken (destructive)         | 19.1  | (compiled) |
| Chicken (non-destructive)         | 157  | (compiled) |
| Gambit (descructive)    | 31.4 | |
| Gambit (non-descructive)    | 31.2 | |
| Gauche                             | 65 | |
| Gauche  (non-destructive)  | 67.7 | |
| Racket | 11.3 | `sort` |
| Sagittarius (destructive) | 53.1 | |
| Sagittarius (non-destructive) | 57.2 | |
| SBCL `sort` | 8.7 | REPL, not optimized |
| SBCL `stable-sort` | 8.6 | REPL, not optimized |

And this SRFI comes with extensive tests (and passes all of them!)

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -78,6 +78,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-128: Comparators (reduced)
     - SRFI-129: Titlecase procedures
     - SRFI-130: Cursor-based string library
+    - SRFI-132: Sort Libraries
     - SRFI-133: Vector Library (R7RS-compatible)
     - SRFI-134: Immutable Deques
     - SRFI-135: Immutable Texts
--- a/configure.ac
+++ b/configure.ac
@@ -596,6 +596,7 @@ AC_CONFIG_FILES([Makefile src/Makefile src/extraconf.h doc/Makefile
 
           lib/srfi-25/Makefile
           lib/srfi-27/Makefile
+          lib/srfi-132/Makefile
           lib/srfi-133/Makefile
           lib/srfi-144/Makefile
           lib/srfi-170/Makefile
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -23,7 +23,8 @@
 # Last file update:  8-Jun-2021 09:02 (eg)
 
 SUBDIRS = Match.d SILex.d Lalr.d ScmPkg.d \
-		  srfi-25 srfi-27 srfi-133 srfi-144 srfi-170 srfi-175
+		  srfi-25 srfi-27 srfi-132 srfi-133 \
+		  srfi-144 srfi-170 srfi-175
 
 COMP ?= ../utils/tmpcomp
 STKLOS_BINARY ?= ../src/stklos
--- a/lib/srfi-132/Makefile.am
+++ b/lib/srfi-132/Makefile.am
@@ -1,0 +1,73 @@
+# Makefile for STklos srfi-132 library
+#
+# Copyright © 2021 Jeronimo Pellegrini - j_p@aleph0.info
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307,
+# USA.
+#
+#           Author: Jeronimo Pellegrini [j_p@aleph0.info]
+#    Creation date: 18-Aug-2021 17:24 (jpellegrini)
+# Last file update: 23-Aug-2021 16:15 (jpellegrini)
+
+SRFI=srfi-132
+
+#======================================================================
+OUT            = $(SRFI).@SH_SUFFIX@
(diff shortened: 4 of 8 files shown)


jpellegrini:
@egallesio - I was sleepy when I first wrote the PR description and mixed up one thing - I said something about the memory used by Timsort while thinking of its stack only..  Anyway, it's corrected now (edited the description).

There is one thing that would need to be decided:
`vector-sort!`, which is allowed to not be stable, would use O(n) memory in the worst case (as implemented now), but runs in O(n) time in the best case. If we revert to heapsort (which uses O(1) memory), it would run in O(n lg n) time best case.

Do we consider `vector-sort!` as something someone would use in memory-critical situations/devices, and use heapsort?
Or do we implement the faster alternative (timsort)?

jpellegrini:
> Do we consider `vector-sort!` as something someone would use in memory-critical situations/devices, and use heapsort?
> Or do we implement the faster alternative (timsort)?

Or quicksort, which is also fast, but uses O(log n) memory?

lassik:
Wow, incredibly thorough work!

egallesio:
Hi @jpellegrini,
@lassik is correct: it is  a very impressive work. 

I hope to merge this PR by the end of the week. 

>  Do we consider `vector-sort!` as something someone would use in memory-critical situations/devices, and use heapsort? Or do we implement the faster alternative (timsort)?

Perhaps we can stay as is for now and  use later a parameter object  (like `vector-sort-algorithm`) to let the user choose the algorithm she/he want to be used.

jpellegrini:
> Perhaps we can stay as is for now and use later a parameter object (like `vector-sort-algorithm`) to let the user choose the algorithm she/he want to be used.

Agreed! :)

egallesio:
Merged

egallesio:
Hi @jpellegrini,
That is weird: code didn't compile (a variable was used before it definition). I have modified it a bit. Tests pass.

jpellegrini:
Great @egallesio !
I don't know why it compiled here and not there. But if it's fixed, that 's great! :)

PR #256, opened 2021-09-02
Try to optimize procedure calls

jpellegrini:
Hi @egallesio !

I was trying to optimize the vm a bit (I don't expect to really get great results, but why not try?) -- and had this idea: in `listify_top`, instead of consing for each argument, allocate an array of cons objects and fill it so it  will be a list.

But for some reason, when compiling STklos with this change, things work until it's time to compile  lang-mzscheme.

```
../utils/tmpgenlex lang-mzscheme.l lang-mzscheme.inc lang-mzscheme
Segmentation fault
```

This is the stack trace:

```
Thread 1 "stklos" received signal SIGSEGV, Segmentation fault.
0x00005555555827f3 in STk_car (x=0x0) at list.c:153
153	  if (CONSP(x)) return CAR(x);
(gdb) where
#0  0x00005555555827f3 in STk_car (x=0x0) at list.c:153
#1  0x00005555555add08 in run_vm (vm=0x7ffff7f98f00) at vm.c:1617
#2  0x00005555555b0067 in STk_load_bcode_file (f=0x7ffff7f99c60) at vm.c:2209
#3  0x00005555555b0102 in STk_load_boot (filename=0x7fffffffe243 "../src/boot.img") at vm.c:2229
#4  0x000055555559e023 in main (argc=3, argv=0x7fffffffdeb8) at stklos.c:236
```
Why would that happen? As far as I can see, the returned value is the same Scheme list that would be returned by the old version of `listify_top`.

This is the idea I had:

`listify_top` gets the arguments from the stack and creates a
list. This is done by consing each element, and that means
an allocation call for each.

We can allocate the memory for all arguments at once.

Advantages:
- A single call to `STk_must_malloc` (and to libgc) instead of several calls
- The arguments will stay close to each other, good for caching.
  I guess they would already if libgc was called sequentially to
  allocate them, but "what if" we have two threads making procedure
  calls?

Drawback:
- so long as there is a reference to *one* of these objects, the
whole list will stay in memory and won't be garbage collected.

We allcoate an array res, of cons cell, each pointing to the next:

```
    +------------+-------------+         +------------+
    |  _____ ___ |   _____ ___ |         |  _____ ___ |
    | | car | ----->| car | ------> ...  | | car | -------> nil
    |  ----- --- |   ----- --- |         |  ----- --- |
    +------------+-------------+         +------------+
         res[0]      res[1]         ...     res[n-1]
```

And return a pointer to the first one ( `&res[0]` ).

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -332,11 +332,50 @@ void STk_print_vm_registers(char *msg, STk_instr *code)
 
 static Inline SCM listify_top(int n, vm_thread_t *vm)
 {
-  SCM *p, res = STk_nil;
+  SCM *p;
+  struct cons_obj *res = STk_must_malloc(n * sizeof(struct cons_obj));
 
-  for (p = vm->sp, vm->sp+=n; p < vm->sp; p++)
-    res = STk_cons(*p, res);
-  return res;
+  /* We first check for n==0, because in this case we need to return
+     STk_nil, and *not* the address of the beginning of a vector whose
+     first cell is nil... */
+  if (n == 0) return STk_nil;
+
+  /*
+    res is a C array of cons cell, each pointing to the next.
+
+    +------------+-------------+         +------------+
+    |  _____ ___ |   _____ ___ |         |  _____ ___ |
+    | | car | ----->| car | ------> ...  | | car | -------> nil
+    |  ----- --- |   ----- --- |         |  ----- --- |
+    +------------+-------------+         +------------+
+         res[0]      res[1]         ...     res[n-1]
+
+    We first allocate it, then fill it so it will be a list, and then
+    return a pointer to the first one ( &res[0] ).
+
+    Why?
+    So we can allocate in a single call instead of calling STk_cons
+    for each value.
+    The drawback: so long as there is a reference to *one* of these
+    objects, the whole list will stay in memory and won't be garbage
+    collected.
+   */
+
+  int i=n-1;
+  for (p = vm->sp, vm->sp+=n; p < vm->sp; p++) {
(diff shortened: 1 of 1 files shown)


jpellegrini:
Another idea: in order to avoid keeping long arrays like that in memory (as described in "Drawback") in case someone creates a procedure with 10000 arguments, we could split the arguments array in a series of arrays.

jpellegrini:
I'd also chance `adjust_arity`, which seems to do something similar...

jpellegrini:
I've found the problem. I needed to return `STk_nil` when n is zero.

But testing, I see that it brings no noticeable enhancement in processing time. At least for the two benchmarks I tried (fib and nqueens, from https://github.com/ecraven/r7rs-benchmarks.git -- so feel free to close this PR if you want.

(Maybe it does make a difference when there are procedures with a large amount of parameters?)

Anyway -- this approach (lists as vectors) could perhaps be used for all lists in STklos? Create them (lists as vectors) with a configurable size, (default 10, maybe) and use the vector from right to left for consing (and create  anew one when it's full).
That could (maybe) speed up the handling of lists.

jpellegrini:
> no noticeable enhancement in processing time

To be more specific, fib went from 1m45.147s to 1m44.644s; nqueens from 3m23.062s to 3m22.202s.

jpellegrini:
I tried also changing `adjust_arity`, but it does not help at all (and the fib benchmark get slightly worse).

jpellegrini:
I think this would only be worth for procedure calls with a really long argument list, so not that interesting. Closing.

Issue #257, opened 2021-09-02
let-values formals: cannot enter single variable to receive values in a list


jpellegrini:
Hi @egallesio !

From what I understand of R7RS, just as one can do

```scheme
(let-values (( (a . b) (values 1 2 3 4) )) b)
 => (2 3 4)
```

It should be possible to do

```scheme
(let-values ((x (values 1 2 3))) x)
```
which should bind `x` to `(1 2 3)`, similar to what we have in lambda formals, ` ((lambda x x) 1 2 3) => (1 2 3)`.

But STklos doesn't allow that:
```
stklos> (let-values ((x (values 1 2 3))) x)
**** Error:
let-values: incorrect binding (x (values 1 2 3))
```
Anyway - most other Schemes accept `(x (values ...))` and return a list.

So I guess STklos could be changed to do that too. I know this is in `r7rs.stk` - I'm a bit busy now but will try to change that if you think it's OK.

(This way of using `let-values` is used a lot in the reference implementation of SRFI-224)

jpellegrini:
It works now! :)

PR #258, opened 2021-09-02
let-values: accept single variable in binding

jpellegrini:
Just as one can do

(let-values (( (a . b) (values 1 2 3 4) )) b)
 => (2 3 4)

It should be possible to do

(let-values ((x (values 1 2 3))) x)

which should bind x to (1 2 3), similar to what we have in lambda formals, ((lambda x x) 1 2 3) => (1 2 3).

This PR seems to fix it, and doesn't break anything (tests still pass).

### DIFF
--- a/lib/r7rs.stk
+++ b/lib/r7rs.stk
@@ -139,7 +139,8 @@ is
       ;; Verify that the binding seem correct
       (unless (and (list? binding)
                    (= (length binding) 2)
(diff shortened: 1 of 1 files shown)


jpellegrini:
This would (I think) be a solution to #257 .

egallesio:
I should have seen this one. Thanks, @jpellegrini  for the fix.

PR #259, opened 2021-09-02
Add SRFI-162: Comparators sublibrary

jpellegrini:
Here's SRFi-162, because I thought we could have one more, and this one was easy...

As per suggestion in the SRFI text, I have not created a new file, but included this in the previous comparator library, `srfi-128.stk` so
* `srfis.stk` says you need to load `srfi-128.stk` in order to get SRFI-162;
* `srfi-128.stk` provides both libraries

Passes all tests.


jpellegrini:
Hi @egallesio - I've force-pushed another version of the PR that includes the code in `scheme/comparator.stk` instead of `srfi/128.stk`...

PR #263, opened 2021-09-04
Split stklos into lib and binary

jpellegrini:
So we can have an embedable `libstklos` in the near future. A small patch, as you can see!

I'll work on an example program embedding STklos.

Pending:
* Don't clutter `$PREFIX/include` with `socket.h`, `extraconf.h` etc.
* Maybe have three targets, `install-base`, `install-dev` and `install`? (Not sure if this is needed)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -133,6 +133,8 @@ case $OS in
       OS_NAME_VERSION=$OS\_$v;
 esac
 
+# We use libtool because we need to build libstklos separate from the stklos binary:
+LT_INIT
 
 ###
 ### Checks for libraries.
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -59,14 +59,6 @@ if NO_THREAD
   THREAD_FILES = thread-none.c mutex-none.c
 endif
 
-stklos_SOURCES = base64.c boolean.c boot.c box.c char.c \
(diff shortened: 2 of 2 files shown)


jpellegrini:
One more thing: the binary can't find the library and I had to set LD_LIBRARY_PATH manually. I think there is a way to fix this.

jpellegrini:
An I'd have to study the STkos boot process. I though I knew how it worked, but it's more complicated than my initial understanding.
The idea would be to have something similar to a boot image, but without the REPL, just so the user can load the basic libraries.

jpellegrini:
I haven't given up on this :)  I'll try to work on it later... The tricky part is to adjust the boot process.

ryandesign:
> So we can have an embedable `libstklos` in the near future.

I'm surprised because:

https://github.com/egallesio/STklos/blob/590c727038b2790ff047008035072f031eb399d3/README.md?plain=1#L14-L15

Is the README wrong?

jpellegrini:
> Is the README wrong?

I think @egallesio forgot to remove that part of the README. As he mentioned in this PR, STkos used to be embeddable.

By the way, he explained what the problem is, and if there is a solution then STklos cold be embeddable again...

egallesio:
> Is the README wrong?

For now, it is. I have suppressed this sentence from the README. I'll reintegrate it when this will be again possible.  
This will be done after v2.00.

PR #264, opened 2021-09-05
Add SRFI-214: Flexvectors

jpellegrini:
Hello @egallesio !
This is SRFI-214, Flexvectors.

It depends on PR #258 because the implementation uses the `(let-values ((x ...`
way of using `let-values`.

About performance: access to individual vector elements is done via
`flexvector-ref` and `flexvector-set!`, and they only seem slower than
their `vector-*` counterparts because they check their arguments.
The rest of the SRFI rerference implementation, although could be
optimized, seems to not have any parts that are much slower than
native Scheme vectors, so we may leave it as it is for now.

Maybe we should remove the bounds check, which is done by the `vector-*`
procedures anyway? Or maybe not? :)

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -107,6 +107,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-196: Range Objects
     - SRFI-207: String-notated bytevectors
     - SRFI-208: NaN procedures
+    - SRFI-214: Flexvectors
     - SRFI-216: SICP Prerequisites (Portable)
     - SRFI-217: Integer Sets
     - SRFI-219: Define higher-order lambda
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -136,6 +136,7 @@ SRC_STK = bigloo-support.stk  \
           srfi-190.stk        \
           srfi-196.stk        \
           srfi-207.stk        \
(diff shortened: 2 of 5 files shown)


jpellegrini:
I forgot one thing: I implemented flexvectors as a struct. Maybe I should have done the same as in SRFI-25, and have a printed/readable representation?

Currently we have
```
stklos> (define x (flexvector 1 2 3))
;; x
stklos> x
#[struct <flexvector> 140715222696032]
```

But it could print `#,(flexvector 3 #(1 2 3))`

jpellegrini:
```
stklos> (define x (make-flexvector 3 #f))
;; x
stklos> x
#,(<flexvector> 3 #(#f #f #f #f))
```

egallesio:
Thanks, @jpellegrini, for this PR.

I have merged it, but I stepped back on your first implementation, because creation of *struct* and access to slots is really faster than with objects (since we don't have the Meta Object Protocol overhead). Reading your two versions, it seems that you used objects to customize *flexvector* writing, but printing of structs can be changed too, thanks to the `struct-type-change-writer!`.  Tell me if I have misunderstood something.

jpellegrini:
Hi @egallesio !

Well,  ok that makes sense. The new `-ref` procedure for flexvectors needs to access the internal `vector` object, and that's a method call...

PR #265, opened 2021-09-11
Prettify describe on structs a tiny bit more

jpellegrini:
Hi @egallesio !
Do you thing this is good?
Maybe `describe-parts-header` could be `list-describe-elements` or something. This looks like it could also be useful elsewhere.

Use "There are no slots", "The only slot is" and "The slots are", depending on the number of slots.

```
stklos> (describe z)
#[struct-type z 139710076780608] is a structure type whose name is z.
Parent structure type: #f
There are no slots.
stklos> (define-struct z name)
;; z
stklos> (describe z)
#[struct-type z 139710077482048] is a structure type whose name is z.
Parent structure type: #f
The only slots is:
	name
stklos> (define-struct z name color)
;; z
stklos> (define-struct z name type)
;; z
stklos> (describe z)
#[struct-type z 139710075052160] is a structure type whose name is z.
Parent structure type: #f
Slots are:
	name
	type
```

The same for conditions and instances of structs.

### DIFF
--- a/lib/describe.stk
+++ b/lib/describe.stk
@@ -74,6 +74,13 @@
   (void))
 
 
+(define (describe-parts-header part-list zero one many port)
+  (let ((msg (case (length part-list)
+               ((0) zero)
+               ((1) one)
+               (else many))))
+    (format port msg)))
+
 ;;;
 ;;; describe for STklos instances
 ;;;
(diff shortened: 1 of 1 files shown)


egallesio:
I have merged it, with small modifications. 

Thanks, @jepellegrini, for this PR.

PR #274, opened 2021-11-09
Fix typos in library.stk and add some tests

jpellegrini:
First commit - nothing essential, just some typos!
Second commit - added some tests


egallesio:
Merged. Thanks @jpellegrini .

PR #275, opened 2021-11-09
More R7RS library tests

jpellegrini:
One more test section...  :)


egallesio:
Thanks for these tests @jpellegrini

PR #276, opened 2021-11-10
SRFI-214: change to R7RS library

jpellegrini:
Hello @egallesio !
I have migrated one SRFI to the new R7RS `define-library` implementation. Tell me if this is OK, and I'll try to migrate others.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -127,7 +127,6 @@ SRC_STK = bigloo-support.stk  \
           srfi-190.stk        \
           srfi-196.stk        \
           srfi-207.stk        \
-          srfi-214.stk        \
           srfi-216.stk        \
           srfi-217.stk        \
           srfi-223.stk        \
@@ -194,7 +193,6 @@ scheme_OBJS = compfile.ostk     \
           srfi-190.ostk         \
           srfi-196.ostk         \
           srfi-207.ostk         \
-          srfi-214.ostk         \
           srfi-216.ostk         \
           srfi-217.ostk         \
           srfi-223.ostk         \
--- a/lib/srfi/214.stk
+++ b/lib/srfi/214.stk
@@ -1,5 +1,5 @@
 ;;;;
-;;;; srfi-214.stk         -- SRFI-214: Flexvectors
+;;;; 214.stk         -- SRFI-214: Flexvectors
 ;;;;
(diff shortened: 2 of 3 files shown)


egallesio:
Hi @jpellegrini,

Thanks a lot for your help. Your implementation is absolutely correct. 

However, wait for a couple of days: I'm finishing a new (cheating) implementation of require which  will implement `(require "srfi-nn"))` as
```scheme
(in-module STklos) 
(import (srfi nn))  ;;  which will eventualky load a file in lib/srfi
;; <== eventually add code here for a particular value of nn
(provide "srfi-nn")
```

This is a gross hack, but STklos modules pre-dates SRFIs, and the idea at this time was to hide the SRFI internal functions to the user. So the idea was to have all (and only these) useful symbols available to the user when it requires the SRFI file. Now, we can imagine that this hack (looking if the required file is a srfi) will be deleted later. Now that the Scheme has a library/module system, this is  probably no more necessary. 
  
The advantage of this version of `require` is that the actual SRFIs can all be located in the `lib/srfi` directory, and we will not need anymore to have the relay file `lib/srfi-nn.stk`. 

I have still some problems to resolve (such as extract the macros of an imported file at compile time, rather than at execution time). 

This should make the adaptation of the numerous SRFIS (this is your fault :grinning:)  for `define-library` simpler.

jpellegrini:
Ah, that's a nice thing! Ok, I'll wait then!

egallesio:
Hi @jpellegrini, 

This is harder than I thought, and it is difficult to bootstrap. I thought that the code I had was correct, but I saw that in certain cases, it loads the installed files (in `/usr/local` :disappointed:). Sorry for being so long.

jpellegrini:
Hi @egallesio - no problem!
I'm also a bit slow here due to some personal problems. But we'll get everything ready soon!

egallesio:
I hope everything goes well for you soon.

jpellegrini:
Hi @egallesio !

Is it correct now? :)

egallesio:
Not only correct, it's perfect!
Merged

jpellegrini:
Hm, it compiles fine on my system, but seems to fail Github CI... Why?

jpellegrini:
Ah, I didn't commit Makefile.in.

egallesio:
Great. I have merged your new PR. Thanks

PR #278, opened 2021-11-20
Implement some algebraic simplification

jpellegrini:
I know I said I'd be working on the macros, but... This one is so simple that I had to implement it :)

Don't operate on identities in two-args + and *

```
(+   0 a) => a
(fx+ 0 a) => a
(*   1 a) => a
(fx* 1 a) => a
```

The n-args case would need to be handled in code generation, though...

Without the optimization:

```
stklos> (disassemble-expr ' (+ s (+ a 0)))

000:  GLOBAL-REF-PUSH      0
002:  GLOBAL-REF           1
004:  IN-SINT-ADD2         0
006:  IN-ADD2             
007:
```

With the optimization:
```
stklos> (disassemble-expr ' (+ s (+ a 0)))

000:  GLOBAL-REF-PUSH      0
002:  GLOBAL-REF           1
004:  IN-ADD2             
005:
```

And:

```
stklos> (time
 (let ((s 0)
       (a 2))
   (dotimes (i 100_000_000)
     (set! s (+ s (+ a 0))))
   s))
Elapsed time: 9165.562 ms                                 <=== before optimization
200000000
```

```
stklos> (time
 (let ((s 0)
       (a 2))
   (dotimes (i 100_000_000)
     (set! s (+ s (+ a 0))))
   s))
Elapsed time: 7703.148 ms                                 <=== after optimization
200000000
```

I've repeated the timing 10 times, and it consistently gives something close to 1.5s difference.

Passes all tests.

### DIFF
--- a/lib/peephole.stk
+++ b/lib/peephole.stk
@@ -203,7 +203,7 @@
                                           (this-arg1 code)
                                           (next-arg1 code))))
 
-             ;; [GLOBAL-REF, TAIL-INVOKE] => GREF-TAIL(INVOKE
+             ;; [GLOBAL-REF, TAIL-INVOKE] => GREF-TAIL-INVOKE
              ((and (eq? i1 'GLOBAL-REF) (eq? i2 'TAIL-INVOKE))
               (replace-2-instr code (list 'GREF-TAIL-INVOKE
                                           (this-arg1 code)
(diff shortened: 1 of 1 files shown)


jpellegrini:
I know it sounds silly, but code not hand-written, but generated by macros or whatever, could generate `(* 1 x)` or `(+ 0 x)`...

Later I'd like to simplify the general case, but that would need to be done more carefully in the code generator.

jpellegrini:
> to simplify the general case

And then probably this one, in the peephole optimizer, could be deleted, since it would be handled by the general case.

jpellegrini:
Ok, so I even had another commit that would do this:

```
             ;; (null? '()) => #t
             ;; (not #f)              => #t
             ;; (not everything else) => #f
```

But I decided to log the used optimizations, and these algebraic simplifications seem to be rarely used: during the tests, there was
* ONE occurrence of remove `IN-SINT-ADD2`
* ONE occurrence of `NULLP -> NIL`

I think this kind of optimization may be more interesting with constant folding and other tricks when dealing with boolean and math expressions, but maybe not necessarily interesting by themselves?

Anyway, the code works... If you want I may add the other commit (for `null?` and `not`). Or you can close the issue if this doesn't seem too useful (after all, it's more code into the STklos core).

jpellegrini:
> maybe not necessarily interesting

Actually I think that currently, many optimizations would not be interesting because they won't compensate the cost of procedure calls, which seems a bit high in STklos. I suppose that some form of inlining (for some cases - short procedures) and enhancements to the VM could help, perhaps?
I've considered loop unrolling, but although it's easy with `dotimes`, it requires a large effort in the case of `do` and named `let`, which are more common.

egallesio:
In fact, I think we cannot do such simplifications, since we are no more able to verify that the remaining argument is a valid number. With such an optimization, `(+ #f 0)` will be replaced by `#f`, whereas the user code is invalid.

jpellegrini:
> With such an optimization, `(+ #f 0)` will be replaced by `#f`, whereas the user code is invalid.

Yes... But that case would be caught during constant folding, I suppose?

Anyway, the optimization (in the case it was interesting) would need to be
```
(+ a 0) => (check-number a)
```
instead of
```
(+ a 0) => a
```
I'm not sure it is worth it.

jpellegrini:
@egallesio - I think this PR really isn't good, so I'm closing it...

PR #282, opened 2021-11-23
Small aesthetic changes

jpellegrini:
I'm not sure about the empty line in `repl.stk` (is it intentional?)

Also, I'm not sure the HTTP download procedure would be documented like that; does this make it go into the manual? If so, in which section?

If I got anything wrong tell me!

### DIFF
--- a/lib/help.stk
+++ b/lib/help.stk
@@ -81,7 +81,7 @@
         (string-append
           "Symbols beginning with '%' are intended for low-level use \n"
           "and have therefore been left undocumented.\n"
-          "Furthermore, they can be changed without notice. Dont use them.\n"))
+          "Furthermore, they can be changed without notice. Don't use them.\n"))
       (else
          ;; no documentation
          #f))))
--- a/lib/http.stk
+++ b/lib/http.stk
@@ -270,6 +270,14 @@
 ;; ----------------------------------------------------------------------
 ;;	http-download ...
 ;; ----------------------------------------------------------------------
+#|
+<doc EXT http-download
+ * (http-download url dest)
+ *
+ * Downloads the contents at |url| using HTTP, and sends the result to
+ * |dest|, which must be an output port.
+doc>
+|#
 (define (http-download url dest)
 
   (define (get-username str)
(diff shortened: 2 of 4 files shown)


egallesio:
Thanks, @jpellegrini for this PR. 
I will merge it.

About, `http-download`,  I will postpone the `(insertdoc 'http-downlod)` because I think that this solution is no more appropriate (most of http accesses are now using https,  and it would be probably a better solution to have an extension using `libcurl` which is more général, and always updated). It seems to very simple to implement it with the `easy` interface. It's on my TODO list since a long time, 

Having `http-download` (and the `lex` and `yacc`) is only needed for *ScmPkg* which I think should  be deprecated (no new package since several years, a package system which is different of STklos modules and R7RS libraries).  Thus, `libcurl,`, `lex` and `yacc`  could be extensions. 

BTW, I think it is time to have a new release, to freeze things. I would really like to merge the `deflib` branch in master after that to start on a 2.00 release (bumping the version number is necessary since, it will bring some incompatbilities  with modules, such as immutable symbols, import order résolution different, ...)

jpellegrini:
> Having http-download (and the lex and yacc) is only needed for ScmPkg which I think should be deprecated (no new package since several years, a package system which is different of STklos modules and R7RS libraries). Thus, libcurl,, lex and yacc could be extensions.

Ok, I agree thta this makes sense. Now, having HTTP/lex/lalr as extensions will be really nice!

> BTW, I think it is time to have a new release, to freeze things. I would really like to merge the deflib branch in master after that to start on a 2.00 release (bumping the version number is necessary since, it will bring some incompatbilities with modules, such as immutable symbols, import order résolution different, ...)

Yay, that's great! 8)
We need to get all SRFIs ported to the new system, then... I'll work on some of them -- those with higher number, and which do not depend on smaller ones (because if they do, we'd have to go patching again later)

I'm sorry that I didn't get the new macros for 2.00 -- I've been doing small things, because I'm having some problems that are preventing me from focusing for long hours. I think things will be better soon.

PR #283, opened 2021-11-23
SRFI-132: now an R7RS library

jpellegrini:
Hi @egallesio !
Now I've tried porting a mixed SRFI. I hope I got everything right...


egallesio:
Everything is alright. 
Thannks

PR #284, opened 2021-11-23
SRFI-60: now an R7RS library

jpellegrini:
@egallesio -- this converts SRFI 60 into a  R7RS library.

One question: other libraries use SRFI 60, and I changed `(require "srfi-60")` to `(import (srfi 60))`, but I see you have used the module name in other places, like `(import srfi/4)`. Probably we should use only one of the styles (parentheses or slash). Which one is preferred?


egallesio:
@jpellegrini, I think that we should use now the R7RS notation `(import (srfi N))` 
The form `(import srfi/4)` should not be used anymore. It was necessary during the bootstrap of the new import system, since the parenthesized form was not accepted at this time. There is no more reason to use it now.
I have modified all (I think) these old forms, when merging the PR.

PR #287, opened 2021-11-23
SRFI-174: now an R7RS library

jpellegrini:


jpellegrini:
@egallesio -- having the SRFIs in the same line in `Makefile.am` seems to be a little troublesome - we keep generating git conflicts.
Maybe we can change that to one-per-line?

PR #288, opened 2021-11-23
SRFI-173: now an R7RS library

jpellegrini:


egallesio:
Done by hand since it was simpler than resolving conflicts.

PR #292, opened 2021-11-27
Use append on two lists, not cons (stklos-compile command line parsing)

jpellegrini:


egallesio:
Oops. Thanks, @jpellegrini

PR #296, opened 2021-11-27
SRFIs 66 and 74: now R7RS libraries

jpellegrini:
74 uses 66, so I converted both at once. Sorry, should have been two commits, but it's already done...


egallesio:
Thanks @jpellegrini,

I have merged this one.
(Not a lot of time, in this period, I merge the easy ones only).

PR #298, opened 2021-11-27
SRFI 217: now as an R7RS library; also add SRFI 224, Integer Mappings

jpellegrini:
One more! :grin:


egallesio:
I have a question with this one. 
This module requires the file trie.stk.
The reference implementation for chicken and Chibi include the trie file, but your implementation require it. That means that this file is dynamically loaded (instead of in-lined in the compiled version). 
I propose to migrate the trie.stk in directory `lib/srfi/include/trie.stk` and include it in `srfi-214.stk`.
The advantage is that `srfi-214` will load only one file (and we don't need to install `trie.ostk`). On counterpart, we cannot load `trie.stk` independently of `srfi-214`, but since it is not packaged as a module, it is probably better not to have it directly usable in this state). 

I can do that, if you think it is OK.

jpellegrini:
Hm, wait a little bit...
I think there are other SRFIs that use tries (at least one). We should probably have something like one single trie implementation (also one single AVL or red-black tree implementation) and adapt SRFIs to use that. Do you agree? Otherwise there will be duplicated code...

egallesio:
Yes you are correct we must avoid code duplication  (however, I don't see which other srfi uses trie)
Or perhaps, i have not merged it yet :thinking:

jpellegrini:
Oh, I think I have not made the PR yet :)
But one of the latese SRFIs seems to also use a trie (224 I think)

jpellegrini:
I am working on a trie implementation that may be useful itself, so it could prehaps be called `(stklos trie)`. Or `(stklos patricia)` :)
Will make a PR when it's ready.

jpellegrini:
@egallesio :  I'm still working on a trie implementation, but I'm not sure about one thing.
Would it be useful to have STklos-specific libraries in `lib/stklos`, with a `Makefile` similar to that one in `lib/srfi`? Then we could have some STklos-specific libraries, and I'd put `(stklos trie)` there.
But on the other hand, there already is the extensions directory. But the trie wouldn't be an extension... I't s more fundamental, and needed by two SRFIs.
Other things that could be in `lib/stklos` are
* `(stklos avl)`, or `(stklos bsearch-tree)` (at least one SRFI depends on it)
* `(stklos heap)`
* `(stklos doulbly-linked-list)`
* `(stklos b-tree)`
* maybe other (non data structures) libraries also?

jpellegrini:
Actually, what I have worked on is an implementation of an trie with fixnum keys, with or without associated values (this covers both SRFIs that I mentioned) -- but not a trie for general string dictionaries and such, because the implementation would be too different.

So: I have a partial implementation in C that will create two new data types, `<fxmap>` and `<fxset>`, and they share a lot of code (so it wouldn't make sense to split the C code in two files, there would be a lot of duplication).

jpellegrini:
An update: I've been working on a single mixed C/Scheme implementation of trie for these two SRFIs (217 and 224), and it's becoming interesting. But it'll take a little while more.

jpellegrini:
Hi @egallesio !
I'm updating the pull request, but I think the code can be cleaned up a bit. But it already passes all tests for both SRFIs, 217 and 224! :smiley:

jpellegrini:
One issue only (but not really serious):
The delete procedure is faster than the referene implementation, but could be faster.
What I did was to write delete on top of filter, but then for each deleted element, a Scheme call to "=" is performed on each element of the trie.  Maybe this can be enhanced in the future.

egallesio:
Great. I'll try to integrate this PR asap. 
As you said, `trie` could be in the `lib/stklos` directory. I have already one new file in it (a simple text preprocessor to embed Scheme code in texts). I will try to place `trie` in `lib/stklos` when I'll merge your PR. 

FYI, I use this new preprocessor for the new documentation, which is no more in Skribe but in `adoc`. 
I have made a new backend to help me in the translation process and I have nearly 2/3 of the documentation which is converted. It is a quite boring task, but I hope to finish it for next week.  The advantage is that we will be able to regenerate the documentation at each build. Stay tuned.

jpellegrini:
Ok @egallesio !
I have pushed a new commit with some fixes to arguments checking.

jpellegrini:
Hi @egallesio !
I see yo uhave incorporated this! :)
One thing: it seems that there is some small problem when building the documentation:

```
make distclean && autoreconf && ./configure CFLAGS="-g -Wall -O2" && make -j10
...
Extracting documentation from itrie.stk
make[3]: Leaving directory '/home/jeronimo/pkg/scheme/STklos-github/lib/stklos'
make[2]: Leaving directory '/home/jeronimo/pkg/scheme/STklos-github/lib'
cat ../src/DOCDB ../lib/DOCDB > DOCDB
for d in refman; do (cd $d; make all); done
make[2]: Entering directory '/home/jeronimo/pkg/scheme/STklos-github/doc/refman'
make[2]: *** No rule to make target 'libs.adoc', needed by 'tmp-html.txt'.  Stop.
make[2]: Leaving directory '/home/jeronimo/pkg/scheme/STklos-github/doc/refman'
make[1]: *** [Makefile:688: all] Error 2
make[1]: Leaving directory '/home/jeronimo/pkg/scheme/STklos-github/doc'
make: *** [Makefile:428: all-recursive] Error 1
```

egallesio:
Hi @jpellegrini,

I have merged (manually) this PR. The `itrie` is now in `(stklos itrie)`, as we discussed before. It can be imported just with `(import (stklos itrie))`. BTW, there are no more `require` in the code. 
The documentation has been integrated in a new appendix about libraries (file is `doc/refman/libs.adoc`). 

Two lines of introductions about `itrie` would be cool and the order of the functions is alphabetic, which is probably not the better choice. Feel free to change things as you like.

jpellegrini:
And you have just commited the fix! :)

PR #300, opened 2021-11-28
SRFI 196: now as R7RS library

jpellegrini:
Hi @egallesio !
I'm sending individual PRs for each migrated SRFI. If you prefer a single PR with several commits, tell me.


egallesio:
That's fine as is. It is easier to review and merge. 
Thanks.

PR #303, opened 2021-11-30
Better error messages for SRFI-215

jpellegrini:
Make the error messages more like the STklos style, and add two tests for errors.

Also test for integer before testing for exact, otherwise in case a bad number is given, the `exact?` procedure will complain that `x is not a number`, and the user won't see that the error was triggered in SRFI-215.

### DIFF
--- a/lib/srfi/215.stk
+++ b/lib/srfi/215.stk
@@ -76,7 +76,7 @@
 (define (dequeue! q)
   (let ((first-pair (car q)))
     (if (null? first-pair)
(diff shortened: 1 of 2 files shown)


egallesio:
Didn't see that it was so simple. I should have merged it a looong time ago. Sorry @jpellegrini ?

jpellegrini:
> Didn't see that it was so simple. I should have merged it a looong time ago. Sorry @jpellegrini ?

No problem @egallesio ! I'm happy it's merged! :)

PR #304, opened 2021-11-30
Document MOP

jpellegrini:
Hi @egallesio !
Some more documentation :)
STklos implements a great object system with a MOP, and I'll try to document it little by little.
Some questions:

* What does `slot-init-function` do? It seems to always return the init form, independent of what argument I pass to it
* Shouldn't `slot-init-function` be called `class-slot-init-function`, since it also takes a class as argument?
* The names `class-direct-supers` `class-direct-subclasses` seem a bit inconsistent.  Shouldn't those be either both abbreviated (`class-direct-supers` `class-direct-subs`) or both extended/complete (`class-direct-superclasses` `class-direct-subclasses`)?

### DIFF
--- a/lib/object.stk
+++ b/lib/object.stk
@@ -158,9 +158,87 @@
 ;;;
 ;;; Classes
 ;;;
+#|
+<doc EXT class-name
+ * (class-name c)
+ *
+ * Returns the name of the class |c|. Note that |c| must be a class
+ * (not an object).
+ *
+ * @lisp
+ * (define-class <A> () ())
+ * (define X <A>)
+ * (class-name X) => <A>
+ * @end lisp
+ *
+ * The result is a symbol, not a string.
+doc>
+|#
 (define (class-name C)
   (if (class? C) (slot-ref C 'name) (%error-bad-class 'class-name C)))
 
+#|
+<doc EXT class-direct-supers class-direct-subclasses class-precedence-list class-slots class-direct-slots
+ * (class-direct-supers c)
+ * (class-direct-subclasses c)
+ * (class-precedence-list c)
+ * (class-slots c)
+ * (class-direct-slots c)
+ *
+ * |Class-direct-supers| and |class-direct-subclasses| return a list with
+ * all direct superclasses or subclasses of |c|, which must be a class
+ * (not object, string or symbol).
+ *
+ * |class-precedence-list| returns the precedence list for class |c|. If
+ * |c| is a superclass of several classes, this list shows what precedence
+ * each superlclass has when resolving names of attributes and methods in
+ * |c|.
+ *
+ * @lisp
+ * (define-class <A> () (slot-one slot-two))
+ * (define-class <B> () ())
+ * (define-class <C> (<A> <B>) (slot-one))
+ * (define-class <D> (<A>) ())
+ *
+ * (class-direct-supers <C>)
+ *      => (#[<class> <A> 7fbc628e6ba0] #[<class> <B> 7fbc626edb10])
+ * (class-direct-supers <D>)
+ *      => (#[<class> <A> 7fbc628e6ba0])
+ * (class-direct-subclasses <A>)
+ *      =>(#[<class> <D> 7fbc6270f2a0] #[<class> <C> 7fbc62704630])
+ * (class-precedence-list <C>)
+ *      =>(#[<class> <C> 7fbc62704630]
+ *         #[<class> <A> 7fbc628e6ba0]
+ *         #[<class> <B> 7fbc626edb10]
+ *         #[<class> <object> 7fbc62aacf30]
+ *         #[<class> <top> 7fbc62aacf60])
+ * @end lisp
+ *
+ * |Slot-one| in class |<C>| refers to the slot defined in |<C>| (because
+ * it is present in both |<A>| and |<C>|, but |<C>| has precedence over
+ * |<A>|). |Slot-two| in |<C>| can only be the one defined in |<A>|.
+ *
+ * The items retruned in the list are not class names; they are the
+ * classes themselves. Use |class-name| to obtain their names.
+ *
+ * |Class-slots| returns a list with the names of the slots in class |c|,
+ * including those inherited from other classes.
+ *
+ * @lisp
+ * (class-slots <A>) => (slot-one slot-two)
+ * (class-slots <C>) => (slot-one slot-two slot-one)
+ * @end lisp
+ *
+ * The names of slots in the list are symbols.
+ *
+ * |Class-direct-slots| returns a list with the names of the slots that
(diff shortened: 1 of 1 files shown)


jpellegrini:
Also, in `(allocate-instance c init-args)`, what is `init-args` used for?

jpellegrini:
Also, `STKLOS-OBJECT` seems to export symbols that I wouldn't expect: `compute-get-n-set`, for example. Is that intentional?

jpellegrini:
One more question...

```
(define-module STKLOS-OBJECT
  (import SCHEME)
  (export       ...)
) ; define-module ends here

(select-module STKLOS-OBJECT)
```

This is different from the style used in other parts of STklos?
Do we change it to have everything inside the `define-module`, making the `select-module` unnecessary?

jpellegrini:
@egallesio I suppose these should also go into the manual in `doc/skb`, ir does just documenting the functions as I did already include the documentation there?

jpellegrini:
@egallesio -- do you think there's anything missing in the `define-method` documentation?
It's hard to explain how it works, without explaining all of CLOS... :grin:

jpellegrini:
I've made also a commit that includes the documentation in the reference manual. I'm not sure if it's OK, maybe you'd like to change the formatting, or the order in which the procedures/syntaxes show up.

jpellegrini:
I've squashed the commits into one, I hope it is clearer...

egallesio:
Hi @jpellegrini,

I don't understand why it takes me so much time to merge this PR. Finally, here it is. 

Responses to questions in the text that follows. 

* What does `slot-init-function` do? It seems to always return the init form, independent of what argument I pass to it

In effect, for instance slots, it returns the init form, but for a class slot (a slot which is shared by all instances of the class), it ensures to initialize this slot only when the first instance of the class is instantiated. Thus, the
initial form is evaluated only one time for class slots).

```scheme
(define-class <A> ()
  ((foo :init-form (random-integer 100))   ;; Here, :allocation is :instance
   (bar :allocation :class    :init-form (random-integer 100))))
stklos> (repeat 5 (describe (make <A>)))
#[<A> 7efe0e0e37e0] is an instance of class <A>.
Slots are:
     bar = 33
     foo = 28
#[<A> 7efe0dfec4b0] is an instance of class <A>.
Slots are:
     bar = 33
     foo = 58
#[<A> 7efe0dfec240] is an instance of class <A>.
Slots are:
     bar = 33
     foo = 31
#[<A> 7efe0dff4fc0] is an instance of class <A>.
Slots are:
     bar = 33
     foo = 44
#[<A> 7efe0dff4d50] is an instance of class <A>.
Slots are:
     bar = 33
     foo = 18
stklos>
```

Here, we see that `foo` is always a different value, whereas the `bar` is always the same value 33.
* Shouldn't `slot-init-function` be called `class-slot-init-function`, since it also takes a class as argument?

IMHO, this would be ambiguous since it returns the function for initializing class slot and instance slots (`class-slot-init-function` would seem to be for class slots only).

* The names `class-direct-supers` `class-direct-subclasses` seem a bit inconsistent.  Shouldn't those be either both abbreviated (`class-direct-supers` `class-direct-subs`) or both extended/complete (`class-direct-superclasses` `class-direct-subclasses`)?

You are right. I have changed that.

* Also, in `(allocate-instance c init-args)`, what is `init-args` used for?

It is used to have the parameters used by the user. Suppose that you have a class `<random-int>` and a class `<random-real>` which maps some classes in a C lib, and that there is no general class for random numbers in this lib.
You can then create a class `<random>` which will either instantiate the later or the former depending of the `:type`  parameter given at object creation: `(make <random> :type 'real)`

I remember to have use used this for graphical gauges in GTK2 which exists as vertical or horizontal, and I wanted to have a class <gauge> with a parameter `:orientation` used to instantiate the good gauge type (the user didn't need to know that the underlying lib distinguish the two type of widgets.

* Also, `STKLOS-OBJECT` seems to export symbols that I wouldn't expect: `compute-get-n-set`, for example. Is that intentional?

Yes, this is probably the most important method to customize your classes. This is this function that is generally used when you make your meta-class. One good example is in `/extensions/gtklos/lib/stklos/classes/meta.stk`.
Once this `compute-get-n-set` method is defined, you have two new `:allocation` schemes defined: `gtk-property` and `gtk-property-enum`. Slots with one of this allocation scheme are in fact virtual and their getters and setters are built thanks to the `<gtk-virtual-slots-metaclass>`.

Another example is when you want to map an object over a SQL line in a table. You can imagine, that you have only one real slot in your object which is a primary key. Other slots correspond to row of your table. Your `compute-get-n-set` function has just to return functions which access a given cell in your DB by issuing the right SQL orders.

This is also useful for building synchronized objects. You can have something like

```scheme
(define-class <synchronized-object> (<class>)
  ; create a mutex shared by all <synchronized-object>
  ((mtx :allocation :class :init-form (make-mutex))))

(define-method compute-get-n-set ((class <synchronized-object>) slot)
  (if (eq? (slot-definition-allocation slot) :instance)
      (let ((index (slot-ref class 'nfields)))
        (slot-set! class 'nfields (+ index 1)) ; ;; Allocate this new slot
        (list
         (lambda (o)
           (with-mutex (slot-ref (class-of o) 'mtx)
                       (lambda()
                         (eprintf "Reading slot ~s of ~s\n"
                                  (slot-definition-name slot) o)
                         (%fast-slot-ref o index))))
         (lambda (o v)
           (with-mutex (slot-ref (class-of o) 'mtx)
                       (lambda()
                         (eprintf "Writing slot ~s of ~s\n"
                                  (slot-definition-name slot) o)
                         (%fast-slot-set! o index v))))))
      (next-method)))

```

To build a class which use a mutex when you read/write an instance slot, you have just to use the `<synchronized-object>` metaclass.

For instance:

```scheme
(define-class <C> ()
  ((a :init-form 1)
   (b :init-form 2))
  :metaclass <synchronized-object>)

(define x (make <C>))

stklos> (slot-ref x 'a)
Reading slot a of #[<C> 7f6ba0584f00]
1
stklos> (slot-set! x 'a 1000)
Writing slot a of #[<C> 7f6ba0584f00]
stklos> (slot-ref x 'a)
Reading slot a of #[<C> 7f6ba0584f00]
1000
```



---
One more question...

```scheme
(define-module STKLOS-OBJECT
  (import SCHEME)
  (export       ...)
) ; define-module ends here

(select-module STKLOS-OBJECT)
```

This is different from the style used in other parts of STklos?
Do we change it to have everything inside the `define-module`, making the `select-module` unnecessary?


**Done.**

jpellegrini:
Hi @egallesio !

Great news! I'll work on documenting more of the MOP later, based on what you have explained here. But it will take a while.

PR #307, opened 2021-12-01
Small enhancements to documentation

jpellegrini:


jpellegrini:
Hi @egallesio !
I put back the `@l` tags I had removed, and also included several `R57RS` and `R7RS` tags.

egallesio:
Hi @jpellegrini, 

I have applied (not always exactly as in the PR), all your commits (except one) in the new documenation.
The commit I have not applied is  the  3db56b7, since the semantic of the tags I had is different that your interpretation : 
 - R5RS means that the symbol is defined in R5RS (and probably in R7RS)
 - R7RS means that the symbol is something new that R7RS defines
 - R5R7RS means that the symbols existed in R5RS but has been changed in R7RS. For instance, the assoc   function which has been in R7RS with a supplemental parameter. All the R57RS tagged symbols should have a remark indicating the new behavior.   
 
BTW, the R5RS legacy is probably too present in the current documentation, now that STklos is mostly R7RS. It should be changed, and  R5RS should not be seen anymore as the reference, but R7RS.

jpellegrini:
Ah, I didn't know that this is what the tags meant. Sorry!

Ok. Anyway, you said you'd change the documentation to adoc, right?  Does that mean we should wait to do more documentation work?

egallesio:
> Ah, I didn't know that this is what the tags meant. Sorry!

No problem. After all, it is explained nowhere....

> Ok. Anyway, you said you'd change the documentation to adoc, right? Does that mean we should wait to do more documentation work?

You will not have to wait too long. I should commit today or tomorrow morning :crossed_fingers:

jpellegrini:
Hi @egallesio -- do we close this one?

egallesio:
Yep. Done.

PR #308, opened 2021-12-01
Some fixnum optimizations

jpellegrini:
Hi @egallesio !
I'm not sure about this one. It would maybe make fixnum operations a bit faster, but also makes the code a little less readable.
Also, the optimizations in `fxodd?` and `fxeven?` look a bit like something that "a smart compiler would figure out anyway"...
But I didn't check the generated assembly for them. And I didn't run any benchmarks this time either... 

Commit message:

Try to avoid tagging/untagging in some places.

A new macro, `UNTAG`, was created that just clears the two first bits of a word.

### DIFF
--- a/src/fixnum.c
+++ b/src/fixnum.c
@@ -690,13 +690,17 @@ doc>
 DEFINE_PRIMITIVE("fxbit-count", fxbit_count, subr1, (SCM o))
 /* TODO: Somewhat efficient, but there is a better way, described
    in "Hacker's Delight", but it needs to be adapted to the
-   variable length datum. */
+   variable length datum.
+   Other possibility: do as in
+   http://graphics.stanford.edu/~seander/bithacks.html
+*/
 {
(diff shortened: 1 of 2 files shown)


jpellegrini:
I think it's possible to make all of `fxbit-count` (SRFI-143), `bit-count` (SRFI-151) and `bit-count`/`logcount` (SRFI-60) better. I'll take a look into that later.

jpellegrini:
And the current implementation in SRFI-143 seems wrong for negative numbers.

jpellegrini:
Ok, it was simpler than I thought.
The implementation of `fxbit-count` was wrong, but it's fixed and optimzied in this version.

Still not sure about the readability after this patch is applied...

egallesio:
Hi @jpellegrini,

Applied this PR. As you said, code is a bit less clear, but your comments really help. 
Thanks for this PR.

PR #317, opened 2021-12-03
Add SRFI-229: Tagged Procedures

jpellegrini:
Hi @egallesio !

STklos already has a property list field for procedures, so it was actually easy to implement an efficient version of this SRFI!
A very small file, Scheme-only. :)


egallesio:
Merging it. Thanks, @jpellegrini.

PR #320, opened 2021-12-03
Add SRFI-95: Sorting and Merging

jpellegrini:
It's basically a wrapper for SRFI-132.

But STklos tries to load the same module twice, in tests.


jpellegrini:
(See issue #321 )

egallesio:
See my comment in Issue #321

jpellegrini:
Ok @egallesio I got SRFI-95 working, but for some reason, after I added it SRFI-132 fails all tests (???)

jpellegrini:
I think I found what is wrong:

If I `require-feature` on SEFI 132, it works:
```
stklos> (require-feature "srfi-132")
stklos> list-sorted?
#[primitive list-sorted?]
```

But if I do 95 before 132, it doesn't:

```
stklos> (require-feature "srfi-95")
stklos> list-sorted?
**** Error:
%execute: variable `list-sorted?' unbound
stklos> (require-feature "srfi-132")
stklos> list-sorted?                                         ;; <== should have worked this time!
**** Error:
%execute: variable `list-sorted?' unbound
```

And this is what the tests do. I think the first time SRFI 95 is loaded, it imports SRFI 132, and the feature is registered (shouldn't?), but the symbols from SRFI 132 are *not* exported. So when later the tests `(require-feature "srfi-132)"`, nothing is done, since the feature is already registered...

Or did I get this wrong?

jpellegrini:
> And this is what the tests do. I think the first time SRFI 95 is loaded, it imports SRFI 132, and the feature is registered (shouldn't?), but the symbols from SRFI 132 are _not_ exported. So when later the tests `(require-feature "srfi-132)"`, nothing is done, since the feature is already registered...

And in `require-feature` (in `srfi-0.stk`), ti is said in the comments that "_.so files cannot be loaded more than once._" -- so maybe this is what happened. SRFI-132 is loaded once, but not twice (because loading the .so file twice can't be done), and in the first time, its symbols are inported into module `srfi/95` only.

Not that this SRFI is of much importance (SRFI-132 is much better fpr sorting/merging), but the problem of `require-feature` that it brings may be... Do you agree?

egallesio:
I have some problems with this PR. I'll try to be clear.
1. if I merge it as is, I have 260 errors when testing SRFI-132
2. a quick fix consists of placing `(import (srfi 132))` at the beginning of the test file => no error, all the tests pass.
3. Since my goal is to avoid the use of the `require-feature` and the hack which transforms some "srfi-n" in "srfi/n", I  plan to import libraries instead of requiring file. The code is partially commented since several weeks in `test/test-srfi.stk` and now that all SRFIs are in the `srfi` directory, this should work.... Except that I have 2 errors with SRFI-144 where there are 2 **FIXME** in the code (test 118 and 120).  I have tried to figure why, but I have bot found yet the reason... If you have an idea, it would be great.

Here is a temporary version of `test/test.srfi.stk`, if you have some time to look at this issue. 
```scheme
(require "srfi-48")   ;; for ~F format


(test-section "SRFIs")

;;;
;;; Run all the tests in "./srfi" directory
;;
(load "../lib/srfis.stk")

(define-library (srfi 15))
(define-library (srfi 62))
(define-library (srfi 87))
(define-library (srfi 143))
(define-library (srfi 169))
(define-library (srfi 176))
(define-library (srfi 193))
(define-library (srfi 219))


(define trace-srfi
  (let ((count 0))
    (lambda (nb)
      (if nb
          ;; print the SRFI number
          (begin
            (when (zero? (modulo count 12))
              (eprintf "\n  ")
              (set! count 0))
            (eprintf "~5f" nb)
            (set! count (+ count 1)))
          ;; print spaces before failed/passed status
          (eprintf "~a   " (make-string (- 60  (* count 5)) #\space)))
      (flush-output-port *err*))))

(dotimes (nb (+ (last-implemented-srfi) 1))
  (let ((desc (find-srfi nb)))
    (when desc                            ;; We have an impl. for this SRFI
      (let ((test-file (format "./srfis/~a.stk" nb)))
        (when (file-exists? test-file)    ;; and we have a test for it
          ;; Add a subsection title (and a trace)
          (test-subsection (format "SRFI ~S - ~S ..." nb (%srfi-title desc)))
          (trace-srfi nb)
          ;; Require the SRFI and run the tests
          ;;  FIXME: Should be:
          (eval `(import (srfi ,nb)) (current-module))
          ;;(require-feature nb)
          (load test-file))))))

(trace-srfi #f)

(test-section-end)
```

egallesio:
Oh I think that I have something. When running, the test of SRFI-144, `expt` is no more the STklos primitive but a closure. This explains why testing only 144 was OK.

jpellegrini:
> if I merge it as is, I have 260 errors when testing SRFI-132

Yes - I think it's because of the problem I described before... No?

egallesio:
>  I think it's because of the problem I described before... No?

Yep and adding `(import (srfi 132))` at the top of `tests/srfi/132.stk` solves the problem. However, I really would like to avoid this kludge and imports all SRFIs which have a test file. In this case, SRFI-144 doesn't work anymore :disappointed:

jpellegrini:
Is it worth taking  look at how the .so files are loaded (the problem described initially), so this would just automatically work?
If not, well, this is not that important a SRFI, I just thought it would be easy to implement on top of the newer sorting SRFI...

egallesio:
> Is it worth taking look at how the .so files are loaded (the problem described initially), so this would just automatically work?

The problem is that multiple initialization of shared libraries could lead to problems and some system refuse to "mmap-ping" a library which has already been loaded.

Anyway, this is merged now, and I don't use anymore `require-feature`, but `(import (srfi n))`.

PR #323, opened 2021-12-04
SRFI 118: now an R7RS library

jpellegrini:
Just take the names from the STkos module and export them. I guess I didn't miss anything...


egallesio:
It seems perfect. Thanks

PR #327, opened 2021-12-04
SRFI 130: R7RS library

jpellegrini:
@egallesio , when I implemented this SRFI I forced it to override STklos' procedures (see the end of 130.stk).
With the R7RS library system, this seems to not make sense anymore, so I commented that out.
If you agree, I can update this PR removing the commented code.


egallesio:
It's already done. Thanks.

PR #332, opened 2021-12-05
SRFI 88 -> R7RS

jpellegrini:
Hi @egallesio !
So, although keywords are native, I see that the SRFI defines three names... So I added the library wrapper, and also changed `srfis.stk` to list `srfi-88.stk` as a required file.

By the way, sorry for flooding you with so many pull requests. I had some unusual days with more free time :)


egallesio:
> By the way, sorry for flooding you with so many pull requests. I had some unusual days with more free time :)

No problem at all. I do the integration ta my own (slow) pace. That's impressing the amount of work you have done. Thanks again, @jpellegrini for all of this.

PR #333, opened 2021-12-05
SRFI 195 -> R7RS

jpellegrini:


egallesio:
I have made changes to this SRFI (and analogous "relay" SRFIs)

`(import STklos)` is replaced by `(import SCHEME)` since module SCHEME cannot be modified (not really looked yet at your code for that).

The `srfis.stk` must not cite "srfi-195" to be loaded since this is not always necessary. 
If you say `(require-feature 95)` (or cond-expand), the file doesn't need to be loaded since the SRFI is embedded in STklos.

However, `(import (srfi 95))` which can be used in a more restricted environment that STklos, will load the file `lib/srfi/95`

PR #339, opened 2021-12-07
Add library-name

jpellegrini:
Module-name will return the name of a module.
If the module is also an R7RS library, we could also ask its library name, which is different fro mits module name!

```
(define-library (example cool-library))
(library-name (find-module 'example/cool-library)) => (example cool-library)
(module-name  (find-module 'example/cool-library)) => example/cool-library

(define-module example/a-module)
(library-name (find-module 'example/a-module))     => error

(library-name quotient)                            => error
```

The error message tries to be informative: if the object is a module, say "the module is not a library"; if it's not even a module, then "bad library/module"...

Is this the best way to do this?

### DIFF
--- a/lib/module.stk
+++ b/lib/module.stk
@@ -534,3 +534,35 @@ doc>
 |#
 (define-macro (in-module mod symb . default)
   `(apply symbol-value* ',symb (find-module ',mod) ',default))
+
+;;=============================================================================
+;;
+;;                              R7RS LIBRARIES
+;;
+;;=============================================================================
+
+#|
+<doc EXT library-name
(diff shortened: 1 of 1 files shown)


jpellegrini:
This would make it possible to simplify the code in PR #337

jpellegrini:
Oops. Is it in the wrong file? Now I see there is `library.stk`...

egallesio:
Thanks @jpellegrini, 
Just moved the fucntion from `module.stk` to `library.stk`

egallesio:
> This would make it possible to simplify the code in PR #337

Done.

Issue #335, opened 2021-12-07
Shouldn't the SCHEME module be immutable?


jpellegrini:
Hi @egallesio !
As per the issue title... Since, as in `boot.stk`, 
```scheme
;;; The SCHEME module is a copy of the STklos module before the user
;;; may have the possibility to modify standard bindings.
;;; If a module needs to be sure to have the original bindings, it
;;; can import this module which will be visible before the STklos
;;; module. Note that this is only partially true since bindings
;;; in SCHEME module are mutable
```
then shouldn't the `SCHEME` module be immutable? (<s>I actually didn't find the definition of modules in the code...</s> Ok, I found - it's in `env.c`)

I'm asking because I would like to (in the future) count on being able to get the original bindings of the standard Scheme symbols.

egallesio:
It is on my TODO list....

jpellegrini:
Can I help?
Is if the only module that should be immutable, or should it be a property of modules? The second sounds better.
What about this?

* Include a mutable flag in all modules
* When a module is created, it is mutable by default, except for the `SCHEME` module
* Add a `module-mutable?` predicate
* Later we may decide if other modules should be mutable, or if it should be possible to lock or unlock a module.

This sounds easy, I could make a PR in the next days. Is that OK?

jpellegrini:
Hi.
I think I got it working... :)
See PR #340 .

jpellegrini:
Hi @egallesio !
I think this one is fixed, no?

egallesio:
Hello  @jpellegrini 

Happy new year.  I was  off the net for the last couple of wee, but I'm back.  Sorry for that.

About this issue, you are right: I close it.

jpellegrini:
> Happy new year. I was off the net for the last couple of wee, but I'm back. Sorry for that.

There's nothing to be sorry about! :)
Happy New Year!

PR #340, opened 2021-12-08
Add non-mutable bindings in modules

jpellegrini:
Hi @egallesio !

About issue #335 ... I think this PR implements it.

I tried to not put more code in the VM, except for when the error is signaled (the module name is printed), so STklos won't get a slower `set!`. But `define` would need to check if the modules allows new bindings (should be very fast anyway).

The only part which seems a bit inefficient is that when we call `define`, it retrieves the old binding, if it exists, then makes it mutable again (if it was not), and that is a bit expensive. But I suppose people won't call global `define`s inside loops. :)

The commit message:

Adds the following primitives:

* `module-lock-define!`     disallows define in a module
* `module-define-allowed?`  checks if define is allowed

* `module-lock-binding!`    makes one binding non-mutable
* `module-lock-bindings!`  makes all module bindings non-mutable
* `binding-mutable?`        checks if a binding is mutable

The `SCHEME` module does not allow new bindings do be `DEFINE`d, and does not allow its bindings to be `SET!`, so it now can be safely used as a reference for the original Scheme bindings.

Update:
* Added a `define-constant` macro (it's trivial now)
* Added the module mutability info to `describe`

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -1643,6 +1643,24 @@ doc>
         value)))
 
 
+#|
+<doc EXT define-constant
+ * (define-constant name expr)
+ *
+ * Evaluates the expression |expr|, then creates a new, non-mutable
+ * binding for |name| with the resulting value.
+ *
+ * (define-constant a 'hello)
+ * a                 => 'hello
+ * (set! a 'goodbye) => error
+ * (define q 2)      => ok (it's a new binding)
+doc>
+|#
+(define-macro (define-constant name expr)
+  `(begin (define ,name ,expr)
+          (module-lock-binding! ',name)))
+
+
 ;;;; ======================================================================
 ;;;;
 ;;;; SRFIs support
--- a/lib/boot.stk
+++ b/lib/boot.stk
@@ -199,13 +199,15 @@
 ;;; may have the possibility to modify standard bindings.
 ;;; If a module needs to be sure to have the original bindings, it
 ;;; can import this module which will be visible before the STklos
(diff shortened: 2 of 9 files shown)


jpellegrini:
* I'm not sure about the procedure names. Those were the best I could think of, but they still sound a bit strange. Tell me if you think of better names!

jpellegrini:
I forgot: what about a description of the SCHEME module in the manual? In Chapter 3, "Program Structure"?

jpellegrini:
Oops, there was a thinko in `env.c`, I corrected and force-pushed

jpellegrini:
Hm, there is one problem: `define-constant` works for globals, but I didn't think of the situation when one would (wrongly?) use it locally.

jpellegrini:
Yes, this needs a tiny bit more work:

```
stklos> (let ((b 2)) (define-constant c 4) (print c))
Received a SIGSGV signal.
```

jpellegrini:
I have pushed another commit, and now there are no crashes anymore.
But `define-constant` only deals with globals. To make it work on local variables, probably we'd need to add a special form to the compiler.

Do I change its name to `module-define-constant`?

egallesio:
Hi @jpellegrini, 

I have changed the way I have implemented things (more on that later).

jpellegrini:
> I have changed the way I have implemented things (more on that later).

Ok @egallesio , it's great that it's implemented! :)

egallesio:
I have added also `define-constant` on functions. You can do:

```scheme 
(define-constant (foo n) 
   (* n n))
```

PR #343, opened 2021-12-09
SRFI 70 -> R7RS library

jpellegrini:
Ok - reexported all symbols mentioned in the SRFI, even those which were already available in R5RS / R7RS, so if the user changes them with `(define quotient *)`, for example, loading the SRFI will bring back the original one. This seems correct, since the intentin of the SRFI is to impose a certain semantics to the numeric fnuctions.


egallesio:
I think it's OK @jpellegrini.
Thanks.

PR #349, opened 2021-12-17
Differentiate "not a number" and "not a real number"

jpellegrini:
The error message when we tried to apply abs, floor,
positive? and similar procedures on a complex number was
"bad number", the same message as if it were a string,
symbol or anything not numerical. We can be more precise
in the error message, without performance penalty and
simplifying a tiny bit the file `number.c`.

This is what weget after this patch is applied:

```
stklos> (abs -1+2i)
**** Error:
abs: `-1+2i' is not a real number
stklos> (abs 'a)
**** Error:
abs: `a' is a bad number
stklos> (positive? -1+2i)
**** Error:
positive?: `-1+2i' is not a real number
stklos> (positive? "x")
**** Error:
positive?: `"x"' is a bad number
stklos> (floor "x")
**** Error:
floor: `"x"' is a bad number
stklos> (floor 2-1i)
**** Error:
floor: `2-1i' is not a real number
```

And it passes all tests :)

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -134,7 +134,7 @@ static void error_bad_number(SCM n)
   STk_error("~S is a bad number", n);
 }
 
-static void error_bad_number_in_comp(SCM n)
+static void error_not_a_real_number(SCM n)
 {
   if (COMPLEXP(n))
     STk_error("~S is not a real number", n);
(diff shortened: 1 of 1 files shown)


egallesio:
This is perfect. Thanks, @jpellegrini.

PR #352, opened 2021-12-24
Fix boolean operator precedence warning

jpellegrini:
GCC complained that 'not' was only being applied to the
left-hand side of an expression... Indeed, it was!
This commit fixes it.

### DIFF
--- a/src/env.c
+++ b/src/env.c
@@ -503,7 +503,7 @@ DEFINE_PRIMITIVE("symbol-mutable?", symbol_mutablep, subr12, (SCM symb, SCM modu
   tmp = STk_hash_get_variable(&MODULE_HASH_TABLE(module), symb);
   if (!tmp)
     STk_error("symbol ~S is not bound in ~a", symb, module);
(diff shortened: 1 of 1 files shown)


egallesio:
GCC is a wise  compiler :smile: 

BTW, which options do you use to see this warning. It does appear with `-Wall -Wextra`?

jpellegrini:
> BTW, which options do you use to see this warning. It does appear with `-Wall -Wextra`?

Hm, `-Wall` should be enough. This is `err.c`:

```c
#include <stdio.h>

int main() {
    int a;
    int b = 0x3;
    scanf("%d",&a);
    printf("b: %x", !a & b);
}
```

Then,

```
$ gcc -Wall err.c
err.c: In function ‘main’:
err.c:6:21: warning: suggest parentheses around operand of ‘!’ or change ‘&’ to ‘&&’ or ‘!’ to ‘~’ [-Wparentheses]
    6 |     printf("b: %x", !a & b);
      |                     ^~
```

Works with GCC 8 and GCC 11 (and presumably with 9 and 10 also).

PR #356, opened 2022-01-24
Explain small ints and fixnums

jpellegrini:


jpellegrini:
By the way, you said you were switching to adoc. Should I also change `vm.md` and `hacking.md` to adoc?

egallesio:
I don't know. In fact, I use asciidoctor (which extends asciidoc). More precisely, I use a preprocessor, which generate some asciidoctor (we can embed scheme expressions in the document). 

It is interesting for multi-chapter documents with intra references, and I like the HTML and PDF rendering. However, one big interest of Markdown is that it is easier to read and the rendering in Github/Gilab is excellent (need to test, how it is with adoc files).  

In any case, we will  be able to convert these document in asciidoc wusing   `pandoc` if needed.

egallesio:
Merged! Thanks @jpellegrini

egallesio:
Hi @jpellegrini,

Finally, I have converted hacking.md and vm.md in asciidoc and moved them in the `doc` directory. I have also set you as the author of the document, since it was not the case in the original document. Tell me if something is not correct for you.

jpellegrini:
Hi @egallesio !

> Tell me if something is not correct for you.

I actually think the documentation looks much nicer now! And it's now generated during the build. In my opinion, this is a very large step ahead!

egallesio:
Hi @jpellegrini ,
Glad to see that you like it.

PR #357, opened 2022-01-25
Use fixed-size integers for opcodes

jpellegrini:
Not for everything in STklos -- using `long` is good, since
it makes STklos use the ideal integer size for integer arithmetic
depending on the platform. But the VM will now always use exactly
two bytes for opcodes, so we can be sure that the bytecode will
always be portable.

```
short          -> int16_t
unsigned short -> uint16_t
```

One thing: when compiling SRFI 25, gcc complains that we're converting `int` into `int16_t`, producing overflows:

```
25-incl.c:104:1: warning: overflow in conversion from ‘int’ to ‘short int’ changes value from ‘65535’ to ‘-1’ [-Woverflow]
  104 | 0xffff,  0x25,  0x75, 0x100,  0x25,  0x65,  0x56,  0x3f,   0x1,  0x6e,
      | ^~~~~~
25-incl.c:109:15: warning: overflow in conversion from ‘int’ to ‘short int’ changes value from ‘65532’ to ‘-4’ [-Woverflow]
  109 |  0x23,  0xbd, 0xfffc,  0x25,  0x25,  0x65,  0x56,  0x1b,   0x1,  0x1d,
      |               ^~~~~~
```

I don't see why it does this since it's the initialization of an array - it should interpret the elements as having the same type of  the base type (use `-Wall --pedantic -fanalyzer` to reproduce). Maybe it was expecting `-0x1` instead of `0xffff` and so on. Anyway, it's bytecode, and that doesn't seem to insert any bugs.

Tests pass, and I've also tested this:

```
stklos> 32767
32767
stklos> (- 32767 1)
32766
stklos> (+ 32767 1)
32768
stklos> (- 32768 1)
32767
```

### DIFF
--- a/src/stklos.h
+++ b/src/stklos.h
@@ -46,7 +46,7 @@ extern "C"
 #include <setjmp.h>
 #include <memory.h>
 #include <locale.h>
-
+#include <stdint.h>
 #ifndef THEADS_NONE
 #  include <pthread.h>
 #  define GC_THREADS 1
@@ -218,7 +218,7 @@ typedef struct {
    * on the compiler. If you change this definition, change DEFINE_PRIMITIVE
    * accordingly
    */
-  short type, cell_info;
+  int16_t type, cell_info;
 } stk_header;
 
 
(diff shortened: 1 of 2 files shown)


jpellegrini:
Now that I think of it, perhaps it's better to change both `int16_t` and `STk_instr` to one single macro, maybe only `STk_instr`? (Or `STk_opcode`?)

PR #362, opened 2022-02-10
Fix fxbit-count

jpellegrini:
It was returning the wrong reult for negative numbers.
The SRFI says that when the argument is negative, it should
return the number of zeros in its representation.

Also changed to a more efficient algorithm with a lookup
table.

(Later we may make itrie use this function instead of a copy of it.)

### DIFF
--- a/src/fixnum.c
+++ b/src/fixnum.c
@@ -653,6 +653,23 @@ DEFINE_PRIMITIVE("fxcopy-bit", fxcopy_bit, subr3, (SCM o1, SCM o2, SCM o3))
   }
 }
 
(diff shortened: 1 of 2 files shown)


egallesio:
Done by hand (due to conflicts, with the merge of previous PR #308).
I hope that everything is OK (tests are passing). 
The table idea is clever. I have made the table local (static of course)  to the function `fxbit-count`.

egallesio:
> (Later we may make itrie use this function instead of a copy of it.)

:+1: I let you do a PR for that? (I don't know if you prefer to export the static `bit_count` or if you want to use the primitive which takes a  Scheme number).

PR #363, opened 2022-03-09
Don't duplicate lookup table

jpellegrini:
The bit-counting code that uses a lookup table was duplicated
(in `fixnum.c` and `itrie.c`). This commit makes `itrie` use the
version in `finum.c`

- Changed `bit_count` in `fixnum.c` from `static` to `extern`
- Made `itrie.c` use it

@egallesio is this ok?

I think it would not be a good idea to make itrie use the Scheme primitive, since all itrie internal functions are optimized to use native long integers, and using the Scheme primitive would require boxing-then-unboxing.

### DIFF
--- a/lib/stklos/itrie.c
+++ b/lib/stklos/itrie.c
@@ -496,43 +496,15 @@ trie_bitmap_delete_max(long b) {
     return b & (~ trie_highestbit(b, trie_lowestbit (b)));
 }
 
-/* A lookup table with the bit count for every possible byte.  Will be
-   used by trie_count_aux and trie_bit_count, since the number of
-   elements in a leaf is the number of bits set in the bitmap. */
-static unsigned int bc[256] =
-{ 0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4, 1, 2, 2, 3,
-  2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5, 1, 2, 2, 3, 2, 3, 3, 4,
-  2, 3, 3, 4, 3, 4, 4, 5, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5,
-  4, 5, 5, 6, 1, 2, 2, 3, 2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5,
-  2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 2, 3, 3, 4,
-  3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 3, 4, 4, 5, 4, 5, 5, 6,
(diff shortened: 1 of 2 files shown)


jpellegrini:
Just a minor comment: it's not cryptographically secure. Here:
```c
  while(n) {
    c += bc[n & mask];
    n = n >> 8;
  }
```
the loop may end after 1, 2, 3 or 4 rounds, so this should not be used in cryptographic code, since it would likely allow for side-channel attacks. Easily fixable, but then it would be perhaps a tiny bit slower.

egallesio:
Hi @jpellegrini,

Thanks for this PR.

I have made some small changes to your code:
-  prefixed the name of the function `bit_count` with a `STk_` (all the public functions start with this prefix)
- replaced the call to the inline function in `itrie.c` by `STk_bit_count` (this should be equivalent, but I'm always suspicious on the fact that the inlining will be effective or not since compilers sometimes refuse to do it).
- changed the table bc to a table of chars to occupy less space

egallesio:
Hi @jpellegrini, 

I have a warning with gcc 12 that I was not able to eliminate easily with a gcc pragma. It is in `itrie.c`

```
itrie.c: In function ‘trie_make_leaf’:
itrie.c:229:65: warning: array subscript ‘struct trie_branch_obj[0]’ is partly outside array bounds of ‘void[24]’ [-Warray-bounds]
  229 | efine TRIE_PREFIX(p)       (((struct trie_branch_obj *)   (p))->prefix)
      |                                                               ^~

itrie.c:732:9: note: in expansion of macro ‘TRIE_PREFIX’
  732 |         TRIE_PREFIX(trie)  = prefix;
      |         ^~~~~~~~~~~
In file included from itrie.c:29:
../../src/stklos.h:239:16: note: object of size 24 allocated by ‘GC_malloc’
  239 |         _var = (SCM) STk_must_malloc(_len);             \
      |                ^
itrie.c:724:9: note: in expansion of macro ‘NEWCELL_WITH_LEN’
  724 |         NEWCELL_WITH_LEN(trie, iset, sizeof(struct trie_leaf_iset_obj));
      |         ^~~~~~~~~~~~~~~~
../../src/stklos.h:239:16: note: object of size 24 allocated by ‘GC_malloc’
  239 |         _var = (SCM) STk_must_malloc(_len);             \
      |                ^
itrie.c:721:9: note: in expansion of macro ‘NEWCELL_WITH_LEN’
  721 |         NEWCELL_WITH_LEN(trie, fxmap, sizeof(struct trie_leaf_fxmap_obj));
      |         ^~~~~~~~~~~~~~~~
```

If you have an idea....

egallesio:
It worked with previous version of gcc (and it still works with clang).

jpellegrini:
> I have a warning with gcc 12 that I was not able to eliminate easily with a gcc pragma. It is in `itrie.c`

Oops. I'll take a look.

jpellegrini:
PR #391

PR #364, opened 2022-03-30
Use $(STKLOS_BINARY) instead of hardcoded stklos relative path

jpellegrini:
Hi @egallesio !

I just realized there was a problem with my ports to wireless routers -- I thought they could be built without having STklos installed, but yesterday I tried compiling on a container, and it didn't work.

There was one place where `$(STKLOS_BINARY)` wasn't used, and compilation failed there.
This PR fixes it. Do you think it's OK?


jpellegrini:
There are some other changes in the PR, because I ran `autoreconf`. I hope this is not a problem (they seemed fine to me).

egallesio:
Hi @jpellegrini, 

Thanks for this PR.

> There are some other changes in the PR, because I ran autoreconf. I hope this is not a problem (they seemed fine to me).

It's not a problem, of course. 
BTW, I have done the PR by hand and have added the initialization of `STKLOS_BINARY` in `Makefile.am`.

If it is OK for you, we can close this PR.

jpellegrini:
> If it is OK for you, we can close this PR.

Yes, sure! I have tested your commit and it compiles without problem!

PR #365, opened 2022-04-29
Optimize vector sorting with fx<, fx>

jpellegrini:
This commit creates fast paths for vector sorting when
the comparison predicate is either `fx<` or `fx>`.

In order to check if the predicate *really* is one of those,
we check if it is `eq` to the binding of the symbols `fx<` and
`fx>` in the immutable `SCHEME` module

Remark 1: Fast paths for `fx<=` and `fx>=` would be possible,
but I'm not sure many people would use those, and they'd
clutter the code.

Remark 2: Fast paths for `fl<` and `fl>` would certainly be
possible, but those are not in the immutable `SCHEME` module
(it's a SRFI), so things become more complicated.

The following is not really a benchmark, but hints at how large the speedup is.

```
stklos> 
(import (srfi 27))
(import (srfi 132))

(define v (make-vector 5000000))

(dotimes (i 5000000)
  (vector-set! v i (- 20000 (random-integer 50000))))

(define w (make-vector 5000000))

(dotimes (i 5000000)
  (vector-set! w i (- 20000 (random-integer 50000))))
;; v
;; w
stklos> (time (vector-stable-sort! fx< v))
Elapsed time: 900.284000000003 ms
stklos> (dotimes (i 5000000)
  (vector-set! v i (- 20000 (random-integer 50000))))
stklos> (time (vector-stable-sort! fx< v))
Elapsed time: 901.396999999997 ms
stklos> (time (vector-stable-sort! < w))
Elapsed time: 6557.967 ms
stklos> (dotimes (i 5000000)
  (vector-set! w i (- 20000 (random-integer 50000))))
stklos> (time (vector-stable-sort! < w))
Elapsed time: 6506.484 ms
```

### DIFF
--- a/lib/srfi/132.c
+++ b/lib/srfi/132.c
@@ -27,6 +27,22 @@
 #include <stklos.h>
 #include "132-incl.c"
 
+
+/* We have fast paths for sorting vectors when the predicate is either
+   fx< or fx>.
+   In order to check if the predicate *really* is one of those, we check
+   if it is EQ to STk_fxlt and STk_fxgt.
+   
+   Remark 1: Fast paths for fx<= and fx>= would be possible, but I'm not
+   sure many people would use those, and they'd clutter the code.
+   
+   Remark 2: Fast paths for fl< and fl> would certainly be possible,
+   but those are not loaded by default (it's a SRFI), so things become
+   more complicated.  */
+
+EXTERN_PRIMITIVE("fx<?", fxlt, vsubr, (int argc, SCM *argv));
+EXTERN_PRIMITIVE("fx>?", fxgt, vsubr, (int argc, SCM *argv));
+
 static void error_bad_list(SCM x)
 {
   STk_error("bad list ~W", x);
@@ -276,12 +292,22 @@ long gallop(SCM pred, SCM to,
          long saved_starta = starta;
          long tmp = starta;
          long i = 2;
+	 
+         int go;
          while (starta == tmp && starta < enda) {
              tmp = starta + i - 1;
-             if (tmp < enda &&
-                 STk_C_apply(pred,2,
-                             VECTOR_DATA(a)[tmp],
-                             VECTOR_DATA(b)[startb]) == STk_true) {
+
+             /* Fast path for fx<, fx> */
+             if (pred == STk_fxlt) /* fx< */
+                 go = INT_VAL(VECTOR_DATA(a)[tmp]) < INT_VAL(VECTOR_DATA(b)[startb]);
+             else if (pred == STk_fxgt) /* fx> */
+                 go = INT_VAL(VECTOR_DATA(a)[tmp]) > INT_VAL(VECTOR_DATA(b)[startb]);
+             else
+                 go = STk_C_apply(pred,2,
+                                  VECTOR_DATA(a)[tmp],
+                                  VECTOR_DATA(b)[startb]) == STk_true;
+             
+             if (tmp < enda && go) {
                  starta = tmp;
                  i = i*2;
              }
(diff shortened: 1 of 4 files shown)


jpellegrini:
Oh - this also makes `Scheme_module` visible to all of STklos (it was `static` before).

jpellegrini:
@egallesio -- this is a nice speedup, but I also thought it would perhaps be interesting, now that we have the immutable `SCHEME` module, trying this also:
- implement counting sort, for both `<`, `>` and `fx<`, `fx>`, in a similar way
- check at which vector size it becomes better than timsort, and use sounting sort when the predicate is one of those, *and* the vector is large enough.

What do you think?

jpellegrini:
Probably `Scheme_module` should be renamed `STk_Scheme_module`? I can do that if you think so.

egallesio:
Hi @jpellegrini,

I think you don't need to use the SCHEME module since you can access the function directly in C. For instance, `fx<` is implemented by the C function `STk_fxlt`. Furthermore, it permits to avoid the use of `find-module` and `symbol-value` and we don't need to rely on the presence of the SCHEME module.

jpellegrini:
> I think you don't need to use the SCHEME

Ooooh, you're right! That was silly of me, the object already is a tagged procedure. Not even the extra argument is necessary. I'll change that.

jpellegrini:
Ok @egallesio -- I changed it so we don't need the SCHEME module. :)

jpellegrini:
Wait, it's not yet working properly. I'll have to review this, it passes the tests, but crashes on some specific inputs on the command line.

jpellegrini:
Hm, this doesn't seem to work:

```c
    if (STk_eq(less,STk_fxlt)==STk_true)
	fprintf(stderr,"fx<?\n");
    else
	fprintf(stderr,"NO fx<?\n");
```

It never seem to recognize `STk_fxlt`.

When `less` is `fx<`, this code:
```c
    fprintf(stderr,"less = %p %ul, STk_fxlt = %p %ul\n",
	    less,
	    (unsigned long) less,
	    STk_fxlt,
	    (unsigned long) STk_fxlt);
```

prints this:
```
less = 0x557dd328d140 3542667584l, STk_fxlt = 0x557dd31fd480 3542078592l
```

jpellegrini:
Ok. Perhaps we really need the SCHEME module.

```c
    SCM fxl   = STk_symbol_value(STk_intern("fx<"), Scheme_module, NULL);

    fprintf(stderr,"STk_fxl = %p %lu\n",
	    (void*) fxl,
	    (unsigned long) fxl);

    fprintf(stderr,"less = %p %lu, STk_fxlt = %p %lu\n",
	    less,
	    (unsigned long) less,
	    STk_fxlt,
	    (unsigned long) STk_fxlt);
```
The result is:
```
STk_fxl = 0x55f3f821e140 94506328383808
less = 0x55f3f821e140 94506328383808, STk_fxlt = 0x55f3f818e480 94506327794816
```

Maybe it's OK to access the binding of `STk_fxlt` from the SCHEME module when SRFI-132 is initialized, and store it in a global variable, visible only in the SRFI? That would get everything working, I guess...

egallesio:
Since you have two C pointers, you cannot do  `if (less == STk_fxlt)` ?

jpellegrini:
> Since you have two C pointers, you cannot do `if (less == STk_fxlt)` ?

I tried, but when I pass `fx<` to the procedure, the pointers are different, for some reason:

```
    less = 0x55f3f821e140
STk_fxlt = 0x55f3f818e480
```

I don't know why this happens. It should be the same pointer, shouldn't it?

egallesio:
> I don't know why this happens. It should be the same pointer, shouldn't it?

Yep! That is really weird. I'll look at it when merging the PR.

egallesio:
Hi@jpellegrini,

I think that this PR is OK (I have the same ratio between fast path and normal path that you had in the first post).
If it is OK for you, we can close.

jpellegrini:
> If it is OK for you, we can close.

Yes, I'm OK with it the way it is.

Maybe later it would be nice to also optimize:

* `fl<`, `fl>` when SRFI-144 is loaded
* `<` and `>` when the vector is a suitable `uvector`

egallesio:
I permit myself to close this PR.
Thanks for testing.

egallesio:
Hi @jpellegrini,

II had to change the call of `memcpy` in the  `gallop` function, since the tests didn't run on Alpine Linux (used when building the Docker image). With `memcpy`, the program crashed on an invalid instruction (probably a `SIGSEGV`, but I don't know how `musl` reports errors). 

I have not changed the other calls to `memcpy` in the rest of the file. I suppose that changing all the `memcpy` should not lower too much the performances.

jpellegrini:
Hi @egallesio !
No problem, I think the difference between `memmove` and `memcpy` is not big.

PR #370, opened 2022-05-06
Add SRFI 116 -- Immutable List Library

jpellegrini:
Hi @egallesio ! 

Here's SRFI 116 -- one more from R7RS large. The only ones missing for STklos to have full red edition implemented are this one, and also 101 (another list library, I have it partially implemented); 125, intermediate hash tables; and 124, ephemerons (this last one is tricky, it requires extra non-libgc memory management code).
But it's close. :)

A fundamental part of the SRFI (and some other procedures) is written in C, and the rest is in Scheme.
A large part of the reference implementation as used, with some ad-hoc procedures for STklos.

Some remarks:

* STklos has immutable list literals, so some names are bound to
  already existent primitives (for example, `ifor-each` is bound to
  `for-each`).
* The `idelete-duplicates` procedure works in expected linear time
  (the reference implementation has a simpler procedure that runs
  in quadratic time).
* The `list-lock+!` procedure is used to change a list, in-place, so
  it becomes immutable.
* I have tried to include documentation for all (or almost all) of the API.
* Passes all tests from the reference implementation!


jpellegrini:
@egallesio -- in `srfis.stk` I have used the name `immutable-lists` for this SRFI. But R7RS Large specifies names for the SRFIs, and this one would be `(scheme ilist)`. Do you think we should already adapt the names in `srfis.stk` to reflect the R7RS names? Or should we later create aliases, or some way for the users to access R7RS Large libraries as `(scheme ...)`, with their names?

https://github.com/johnwcowan/r7rs-work/blob/master/RedEdition.md

egallesio:
@jpellegrini , I have problems with this one. `itake` is exported in the scheme file, but its code is commented. Don't  really know what to do with it.

jpellegrini:
> @jpellegrini , I have problems with this one. `itake` is exported in the scheme file, but its code is commented. Don't really know what to do with it.

I'll take a look.  It's weird, because it passed the tests... And there is a test for `itake`.

jpellegrini:
Oh, I'm sorry! It was a mixup here... The 116.stk file was not the final version... This one, which I pushed now, is.

PR #372, opened 2022-05-11
Fix typos & spelling; fix multiline macros

zdimension:
Multiline macros enclosed in braces are a footguns if used in single-line ifs, as in
```c
#define MACRO { code; }

if (x)
    MACRO;
else // breaks since the ; above counts as an additional statement
```
Replacing that with a `do { ... } while(0)` prevents this.


egallesio:
Hi Tom, 

Thanks a lot for your help with this PR. I have merged it nearly unchanged (just place the star against the variable instead of the type, to be coherent  with the rest of the code (and because, I think it's  the right place :wink: )) .

Thanks again

PR #377, opened 2022-05-15
Add SRFI 152: String Library (reduced)

jpellegrini:
Hi @egallesio !
One more SRFI :smiley: 

* The implementation of SRFI 130 had to be adjusted.
  SRFI 130 impoted and redefined some identifiers from SRFI
  13, and that caused SRFI 152 to fail to re-redefine them
  (can't tell why, but it's not a good idea to redefine an
  imported symbol anyway).

* string comparison functions are redefined so they will accept
  zero arguments: (string=?) is legal and retruns #t


egallesio:
Thanks @jpellegrini  for this PR. 

Merged.

PR #378, opened 2022-05-16
Add bash tab completion to STklos

jpellegrini:
Hi @egallesio !

The file `etc/completion/stklos` is a bash script that registers a completion function for the `stklos` command in Bash. 

Source this file to get tab completion for the STklos command in Bash.  For a permanent effect, either source this from your
`.bash_profile` or put it with the other bash completion files in your system -- for example, in Debian and Debian-based systems, this will be `/usr/share/bash-completion/completions/`

### DIFF
--- a/etc/completion/stklos
+++ b/etc/completion/stklos
@@ -31,12 +31,27 @@
 
 #
 # Source this file to get tab completion for the STklos command in
-# Bash.  For a permanent effect, either source this from your
-# .bash_profile or put it with the other bash completion files
-# in your system -- for example, in Debian and Debian-based
-# systems, this will be
+# Bash.
 #
-# /usr/share/bash-completion/completions/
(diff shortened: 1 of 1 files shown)


jpellegrini:
I'll do the same for `stklos-compile` later.

jpellegrini:
Oh, I just learned about `apropos` (in `bonus.stk`). I could have used it in this patch...

egallesio:
Thanks, @jpellegrini for this contribution. This is really useful.


> Do you think it would be useful to have completion also for all the executables?
> 
>     * stklos-config
> 
>     * stklos-genlex
> 
>     * stklos-pkg
> 
>     * stklos-pp
> 
>     * stklos-script

No. I think that `stklos-compile` and `stklos` are the only two which are really important. 

BTW, I have also added completion for `zsh`.

PR #382, opened 2022-05-22
Add SRFI 19

jpellegrini:
Hi @egallesio .

This PR is based on the SRFI-19 implementation you had started:
* adds nanoseconds to the `date` structure
* fixes some small problems in SRFI 19
* uses the new module system
* uses an up-to-date leap second table

But some of the functions do not work properly, because `make-date` uses Unix time, which does not take leap seconds into account.

For example, Guile doees it correctly:
```
(make-date 0 60 59 23 31 12 1998 0)
$3 = #<date nanosecond: 0 second: 60 minute: 59 hour: 23 day: 31 month: 12 year: 1998 zone-offset: 0>
```
But with `make-date` in STklos, we  have
```
stklos> (describe (make-date 0 60 59 23 31 12 1998 0))
#[date 1999-1-1 0:0:0] is an instance of the structure type %date.
Slots are:
     nanosecond = 0
     second = 0
     minute = 0
     hour = 0
     day = 1
     month = 1
     year = 1999
     week-day = 5
     year-day = 1
     dst = 1
     tz = -10800
```

The function `%seconds->date` calls `localtime`, which ignores leap seconds and never fills in the seconds part of the structure with "60".
It is possible to
* have the C function actually also have a leap table (cumbersome)
* implement a different one in SRFI-19, and have `make-date` in C behave differently from the one in SRFI-19 (confusing)
* remove the date structure from `system.c` entirely, and keep it in SRFI-19. It would not be a totally breaking change, it's moving code from the core to a library, but still feels strange

What else could we try?


jpellegrini:
>     * remove the date structure from `system.c` entirely, and keep it in SRFI-19. It would not be a totally breaking change, it's moving code from the core to a library, but still feels strange
 
I meant only the date functions, not the basic time functions. Basically we'd move `date.stk` and part of `system.c` to `srfi-19.stk`

jpellegrini:
I'll try to keep working on this then. 

From `system.c`:
- Keep `%seconds->date` in C, since it is useful to calculate week days and the local time zone offset
- Move `date->seconds` to SRFI-19 -- also adding leap seconds when appropriate
- `date` just calls `ctime`. Not sure. for now, I'd keep it in `date.stk`
- Keep `%seconds-.string` in `system.c`


From `date.stk`:
- `seconds->date` becomes `time-tai->date`, `time-utc->date` in SRFI-19
- Move `make-date` to SRFI-19 -- this one will add leap seconds when appropriate
- Move `current-date` to SRFI-19 -- also adding leap seconds when appropriate
- The same for the date structure and accessors
- Not sure about `seconds->list` `seconds->string`. I would keep them where they are now.
- `date->string` goes into SRFI-19.
(and rename it to `time.stk`?)

Is that ok?

egallesio:
Hi @jpellegrini,

I didn't have time yet to see the problem you have. The only thing which is important to me, is to be able to have simple manipulations on date/time without importing SRFI-19. I really don't like this SRFI, which is the only one for which we have still messages in the mailing list, whereas it is more than twenty year old (plus  the need to maintain a table of leap seconds). Ideally, for me, this problem should be taken into account by the OS. 
hope you didn't find me too crude...

jpellegrini:
Hi @egallesio !

I have changed the PR to not change much of STkos, and keep the complex stuff in `srfi/19.stk`. The changes that I've madein `system.c` and `date.stk` were necessary in order to accomodate a new date slot (nanosecond) and a difference in the parameter order for `make-date` and `make-time` (otherwise, they'd be reversed when the SRFI is loaded, which could be confusing).

So... It was necessary to change `boot.c`, `boot.img` and a few other files because I renamed `current-seconds` to  `%current-seconds`. The commit message has a lot more details.

Tell me if this is OK.

jpellegrini:
I'm not really a fan of this SRFI, it's kind of confusing and somewhat overengineered I think... But well, it's there, and lots of implementations have it, so maybe for compatibility this would be nice (also, SRFI 120, Timer APIs, has a reference implementation that uses time durations from SRFI-19 -- although we could implement SRFI 120 using some other object for time durations, I suppose it would be better to use the already standardized objects from SRFI 19...)

jpellegrini:
There a problem running the tests (worked before, now wont'...), I'm checking.

jpellegrini:
Ok fixed.
Sorry... In the last minute I added argument type check to `make-time`, and didn't realize that I, myself, had used bogus arguments for temporary time objects in the code... :grinning:

egallesio:
Hi @jpellegrini,

I cannot compile this PR. It complains that the function `current-time` doesn't exist. Obviously, there is a problem with the renaming of this function, but I don't understand where the problem is. BTW, is it really necessary  to have TAI time available without loading SRFI-19?

jpellegrini:
I think `current-time` is being referenced in `boot.img`, but not anywhere else. I'm fixing it.

jpellegrini:
Ok, first problem solved: I hadn't committed the boot files. Sorry, I promise I'll do this next time.

I'll take a look at moving `time-tai` outisde the core.

jpellegrini:
Ah - also, without TAI and the conversion table, we can represent current time just fine, but dates in the past may need the calculation of leap seconds, and some dates (those with "second = 60") wouldn't be represented at all, since Unix time doesn't account for them, and we'd need the table to actually calculate them properly (and then we'd have to keep the table in the core, and the SRFI woudl use it). So,

1. Do we remove TAI and stay liek most Unix programs (ignoring leap seconds)?
2. Or do we include just the table, and produce UTC with forced leap seconds? (some dates with sec=60)
3. Or - do we keep TAI time in the core, which is a really small addition after including (2)?

I can fix it in any of the three ways :)

(The UTC/TAI confusion is not good. There is a rant by Dan Bernstein about Unix time about it... https://cr.yp.to/proto/utctai.html)

jpellegrini:
I'm not sure why, but fixing the boot file didn't work, so I didn't fix the conflicts on this one. I'll try again later.

egallesio:
> I'm not sure why, but fixing the boot file didn't work, so I didn't fix the conflicts on this one. I'll try again later.

Forget this one. I have already taken your PR and adapted it several weeks ago (I had a version quite different from yours to minimize some dependencies). I'm just seeing that I forgot to merge it in master (and now , I wall have the boot problems :angry: ). I will do it soon.

egallesio:
Hi @jpellegrini,

I have finally merged the SRFI-19 :champagne: Sorry for the long delay. 
Thanks again for your help.

jpellegrini:
This is great! :) I'm happy we have this SRFI now.
At this moment, the lowest-indexed SRFI that we don't have is SRFI 42. I'll take a look at it later (after 146)

PR #383, opened 2022-05-22
Add immutability to hash tables, and adjust wording in other places ("locked" -> "immutable")

jpellegrini:
Hi @egallesio !
This may be useful when implementing SRFI 125, which deals with immutable hash tables.

`(hash-table-immutable! h)` makes the hash table h immutable.
No entries can be added, removed or changed after that.

`(hash-table-mutable? h)` determines if a hash table is mutable.

`hash-table-copy` will create a mutable hash table.

`describe` also talks a bit more about hash tables now. :)

I have used 4 ("100") as `HASH_CONST`, although I suspect it would be a good idea to have a single "immutable" bit for all objects later.

### DIFF
--- a/doc/refman/srfi.adoc
+++ b/doc/refman/srfi.adoc
@@ -334,7 +334,7 @@ NOTE: To obtain the symbol `:` in {{stklos}}, you must use `|:|`.
 {{insertdoc 'not-ipair?}}
 {{insertdoc 'null-ilist?}}
 {{insertdoc 'ilist=}}
-{{insertdoc 'list-lock!}}
+{{insertdoc 'list-immutable!}}
 
 {{insertdoc 'itenth}}
 {{insertdoc 'icar+icdr}}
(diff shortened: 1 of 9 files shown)


jpellegrini:
We already used `-lock!` for modules, and here I used `-immutable!`. Which one is better?

lassik:
Mutable and immutable is the usual terminology for this. "Lock" normally refers to concurrency (threads, mutexes, etc.)

jpellegrini:
> Mutable and immutable is the usual terminology for this. "Lock" normally refers to concurrency (threads, mutexes, etc.)

Yes... For some silly reason I have written `lock!` when writing some module-related code. maybe it'd be better to change all procedures that make things immutable to `-immutable!`, as in this PR? (@egallesio ?)

jpellegrini:
> maybe it'd be better to change all procedures that make things immutable to `-immutable`

Before the next release, so we'd not have the `-lock!`-named procedures on a released version?

egallesio:
Thanks for this PR. Merged!

PR #384, opened 2022-05-26
Fixes two issues with SRFI 133

jpellegrini:
Hi @egallesio !
While working on some other SRFIs, I've found two problems with my implementation of SRFI 133. This PR fixes both.

# First issue:

When an exported symbol is already defined in STklos, we
actually define it inside the module, as in

```
(define vector-fill! (symbol-value 'vector-fill! scheme))
```

because otherwise SRFI-133 would require the SCHEME (or STklos)
module in order to work. Also, if someone changed the binding of
vector-fill!, for example, in module STklos, and SRFI-133 counts on
its original binding, then SRFI-133 will not behave as intended.

# Second issue:

R7RS specifies that, for vector-copy!,

> _"The order in which elements are copied is unspecified, except that if
>  the source and destination overlap, copying takes place as if the
>  source is first copied into a temporary vector and then into the
>  destination. This can be achieved without allocating storage by
>  making sure to copy in the correct direction in such circumstances.",_

and that `vector-reverse-copy!` is
>  _"Like `vector-copy!`, but the elements appear in `to` in reverse order."_

So that restriction is also true for `vector-reverse-copy!`.

However, I know of no way to do that without allocating extra space
for _REVERSE_ copying of overlapping parts of the same vector.

We allocate new space *of the size of the part being copied*, so if
the vector has a billion elements and only 10 are being copied, then we
allocate 10 cells.

We split this in two cases, overlapping and non-overlapping...

Also:

- two tests, for this situation, were added to test section for this
  SRFI;
- the Scheme wrapper was rejecting the case when the source and
  target overlaps, but this is implicitly allowed by the spec, so this
  check was removed.

### DIFF
--- a/lib/srfi/133.c
+++ b/lib/srfi/133.c
@@ -138,10 +138,49 @@ DEFINE_PRIMITIVE("%vector-reverse-copy!",srfi_133__nvector_reverse_copy,subr5,(S
     int csstart = INT_VAL(sstart);
     int csend   = INT_VAL(send) - 1;
 
-    while (csend >= csstart) {
-        VECTOR_DATA(target)[ctstart] = VECTOR_DATA(source)[csend];
-        ctstart++;
-        csend--;
+    /* R7RS specifies that, for vector-copy!,
+
+       "The order in which elements are copied is unspecified, except
+       that if the source and destination overlap, copying takes place
+       as if the source is first copied into a temporary vector and
+       then into the destination. This can be achieved without
+       allocating storage by making sure to copy in the correct
+       direction in such circumstances.",
+
+       and that vector-reverse-copy! is "Like vector-copy!, but the
+       elements appear in to in reverse order."
+
+       So that restriction is also true for vector-reverse-copy!.
+
+       However, I know of no way to do that without allocating extra
+       space for REVERSE copying of overlapping parts of the same
+       vector.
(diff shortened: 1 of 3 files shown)


jpellegrini:
> if someone changed the binding of `vector-fill!`, for example, in module STklos, and SRFI-133 counts on its original binding,

This is actually how I found out the problem... It does happen!

egallesio:
Hi @jpellegrini,

I'm sorry to be so slow to integrate all your contributions. I hope to have more free time soon.

For the first issue, I rewrote the code to avoid allocation (I hope this is correct). I write the element in the vector and invert the slice in place. 

For the second issue, I think that just `(import (only (SCHEME) vector-fill! vector-copy ...))` should work (or perhaps I have not understood the problem). Note: I have not modified the code in `133.stk` accordingly.

jpellegrini:
> I'm sorry to be so slow to integrate all your contributions. I hope to have more free time soon.

No problem! :grin: 
 
> For the second issue, I think that just `(import (only (SCHEME) vector-fill! vector-copy ...))` should work (or perhaps I have not understood the problem). Note: I have not modified the code in `133.stk` accordingly.

I'll check again if that works. If so, I'll make a new PR to make the adjustment!

PR #385, opened 2022-05-29
Optimize (void ...)

jpellegrini:
Hi @egallesio !
I think the `void` procedure could be inlined.  Is there some reason for not doing it?

This is what STklos does, currently:

```
stklos> (disassemble-expr '(void))

000:  PREPARE-CALL
001:  GREF-INVOKE          0 0
004:

stklos> (disassemble-expr '(void 1 2))

000:  PREPARE-CALL
001:  ONE-PUSH
002:  INT-PUSH             2
004:  GREF-INVOKE          0 2
007:
```

After this patch -- not only we have less instructions, but we also avoid a procedure call.

```
stklos> (disassemble-expr '(void))

000:  IM-VOID
001:

stklos> (disassemble-expr '(void 1 2))

000:  IM-ONE
001:  SMALL-INT            2
003:  IM-VOID
004:
```

Which is the same as if we had no procedure call, and just either
used `#void` or `(begin  [args]  (void))` directly:

```
stklos> (disassemble-expr '#void)

000:  IM-VOID
001:

stklos> (disassemble-expr '(begin 1 2 #void))

000:  IM-ONE
001:  SMALL-INT            2
003:  IM-VOID
004:
```

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -73,7 +73,8 @@
                          (string-set!    . ,string-set!)
                          (eq?            . ,eq?)
                          (eqv?           . ,eqv?)
-                         (equal?         . ,equal?)))
+                         (equal?         . ,equal?)
+                         (void           . ,void)))
 
 (define *inline-symbols* (map car *inline-table*))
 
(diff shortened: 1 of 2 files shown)


jpellegrini:
And it seems that the form `(void)` is not too unfrequent:

```
$ grep '(void)' lib/*.stk|grep -v 'lib/compiler'|wc -l
41
```

:grin:

jpellegrini:
For the motivation :)

```scheme
(time
(dotimes (i size)
  (vector-set! v i (void))
  (vector-set! w i (void))))
```
After running the above code 10x, the average time was around 151ms

```scheme
(time 
(dotimes (i size)
  (vector-set! v i #void)
  (vector-set! w i #void)))
```
After running the above code 10x, the average time was close to  109ms

egallesio:
Oops.

Thanks foo seeing this one. In previous versions `(void x y)` ignores its parameters and didn't calculate them. That means that  `(void (display "Hello"))` didn't print anything. I wanted that the arguments were evaluated,but  I have not seen tht my change was so inefficient. 
Hopefully, you were here :wink:

PR #387, opened 2022-05-29
Add setters for SRFI-25 arrays

jpellegrini:
A small thing, but I thought it would be nice to have...

### DIFF
--- a/lib/srfi/25.stk
+++ b/lib/srfi/25.stk
@@ -101,6 +101,9 @@
                        shape-for-each array?         shape?
                        share-array    array-size)
 
+;; And why not have a setter for arrays?
+(set! (setter array-ref) array-set!)
+
 ;; EXTRA: almost all procdures in arlib.scm, not part of the
 ;;        SRFI spec but included with the reference implementation.
 ;;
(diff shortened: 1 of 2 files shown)


jpellegrini:
@egallesio - I see there are other SRFIs where this could be done. I have made PRs for 25 and 214, but it would also be possible for

* 74 (Octet-Addressed Binary Blocks, or all the `blob-*-set!` procedures)
* 66 (Octet vectors, for `u8vector-set!`, but this is actually defined in SRFI 4...)
* 4 (Homogeneous numeric vector datatypes, for all the `*-vector-set!` procedures)

If it's OK, I'llmake PRs for these too (tell me)

egallesio:
> A small thing, but I thought it would be nice to have...

For sure :smiley:. Merging it..
Thanks


>  I see there are other SRFIs where this could be done. I have made PRs for 25 and 214, but it would also be possible for
> 
>     * 74 (Octet-Addressed Binary Blocks, or all the `blob-*-set!` procedures)
> 
>     * 66 (Octet vectors, for `u8vector-set!`, but this is actually defined in SRFI 4...)
> 
>     * 4 (Homogeneous numeric vector datatypes, for all the `*-vector-set!` procedures)
> 
> 
> If it's OK, I'llmake PRs for these too (tell me)

Good idea. It would be great.

jpellegrini:
> Good idea. It would be great.

Ok, will push.

PR #389, opened 2022-05-31
Add the R7RS "environment" procedure

jpellegrini:
This adds the R7RS environment procedure (an extended version, actually).

`(environment [name] lib1 lib2 ...)`
    
This procedure returns a specifier for the environment that
results by starting with an empty environment and then
importing each lib (list or symbol), considered as an import
set, into it. The bindings of the environment represented by
the specifier, as is the environment itself.
    
In STklos, it is extended so that:
- each |lib| argument can be a list (specifying an R7RS
  library) or a symbol (specifying a module).
- the return environment is an R7RS library (which can be
  passed to |eval|).
- if |name| is a string, it will be set as the new environment
  (library) name.

This PR also adds tests to it, and to the other procedures of section 6.12 or R7RS. I've added the tests in `test-r7rs.stk`. Not sure if they should go into `test-r7rs-libraries.stk`

### DIFF
--- a/lib/env.stk
+++ b/lib/env.stk
@@ -320,4 +320,105 @@ doc>
   (find-module 'R5RS))
 
 
+
+          
+#|
+<doc environment
+ * (environment [name] lib1 lib2 ...)
+ *
+ * This procedure returns a specifier for the environment that
+ * results by starting with an empty environment and then
+ * importing each lib (list or symbol), considered as an import
+ * set, into it. The bindings of the environment represented by
+ * the specifier, as is the environment itself.
+ *
+ * In STklos,
+ * - each |lib| argument can be a list (specifying an R7RS
+ *   library) or a symbol (specifying a module).
+ * - the return environment is an R7RS library (which can be
+ *   passed to |eval|).
+ * - if |name| is a string, it will be set as the new environment
+ *   (library) name.
+doc>
+|#
+(define (environment . args)
+
+  (define (check-module-names args)
+    (for-each (lambda (arg)
+                (unless (or (symbol? arg) (list? arg))
+                  (error "bad library or module name ~S" arg)))
+              args))
+
+  ;; Given a list of library or module names, return a list of
+  ;; modules.
+  (define (get-modules-from-names args)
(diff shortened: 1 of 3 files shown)


jpellegrini:
If you think it'd be better to have a STklos-specific `make-environment` with more features like accepting a name and a mutability argument, and just have `environment` as a wrapper, strictly abiding by R7RS (no names, results in immutable module) I can do that (it's a small procedure anyway, no big deal)

jpellegrini:
Hm, this one depended on the (now gone) `%make-library` primitive. I'll adapt it.

jpellegrini:
Ok... I pushed some changes.

Still not perfect, I guess, but works:

- uses eval twice (ok I guess)
- uses `(string->symbol (symbol->string name))` on a gensymed symbol
- leaves two extra symbols in the module, `%module-restore` and `%make-syntax` which, even when not exported, can be used, but then, all libraries have them.

jpellegrini:
Or aybe this should actually go into `lib/scheme/eval`? It's R7RS, and _not_ R5RS...

egallesio:
Hi @jpellegrini,

Thanks for this PR. However, I have implemented it differently (I have permitted to have module without name), and it allows having full import specs. That means that you can have:

```scheme
(let ((e (environment '(prefix (only (scheme base) car)
                                 foo-))))
     (module-symbols e))                            => (foo-car)
```

> Or maybe this should actually go into lib/scheme/eval? It's R7RS, and not R5RS...

I left the definition of `environment`, which is now a few lines long, in `env.stk`. It is true that it is R7RS, but all the R7RS functions are available without specific user action. For now the function is an autoload function as the other environment relative functions.

About your extension (a string to name the new environment), I have not implemented it. I don't know what should be the semantic of a call to `environment` with an already existing module or a previous call to `environment` with the same name. 
 
BTW, I stole your tests and added them to the dev tree. Thanks.

jpellegrini:
Hi @egallesio - I'm happy to know there's now an `environment` procedure! This will be useful later.

PR #391, opened 2022-06-06
Silence a GCC warning

jpellegrini:
Make the accessor macro work on the shorter variant of the struct, so GCC-12 will be happy.


egallesio:
Clever! 
Thanks, @jpellegrini .
Merging it.

PR #392, opened 2022-06-06
Check for NULL from malloc

jpellegrini:
`readline-complete.c` needs to use `malloc` instead of `libgc`, and I forgot to check for `NULL` when I wrote it.


egallesio:
Merged!

PR #395, opened 2022-06-19
Fix allocation of uvectors

jpellegrini:
A typo: the code was allocating room for `vector_obj`, and not `uvector_obj`. :)


egallesio:
Oops.  Thanks. 
Merged

PR #397, opened 2022-06-21
SRFI-25: use STk_procedurep instead of CLOSUREP

jpellegrini:
Otherwise, a primitive won't be allowed in share-array (even
though this is unlikely because of the requirements of share-array,
it could be done).


egallesio:
I have applied your patch and fix a type problem (`STk_procedurep` returns a Scheme boolean, not a C one).

There is another `CLOSUREP` later in the code, and I have first changed it to a call to `STk_procedurep` and seen that you test the arity just after with `CLOSURE_ARITY`. I don't know if it is useful to changed it (in this case, we have to use `STk_proc_arity` which accepts any kind of function). What do you think?

Issue #394, opened 2022-06-19
default hash does not work well on uvectors


jpellegrini:
Hi @egallesio !
I've been away for a while, got covid, but I'm recovering.

I was working on SRFI 160 and noticed this:
```scheme
(import (srfi 4))
(hash-table-hash #(1 1 1 1 1)) => 78881
(hash-table-hash #(1 1 1 1))   => 4640
(hash-table-hash #(1 1 1 2))   => 10099
```
But:
```scheme
(hash-table-hash #u64(10 20 30)) => 24
(hash-table-hash #u16(10 20 30)) => 24
(hash-table-hash #s16(10 20 30)) => 24
(hash-table-hash #s8(10 20 30))  => 24
(hash-table-hash #s8(10 20 30))  => 24
(hash-table-hash #c128(1-2i))    => 24
```
(The `c128` vector is part of SRFI-160, I'm half way through it)

I'll see if I can get the standard hash to work on uvectors, then I'll get back here.

Work on other SRFIs that I was doing depends on PR #384 (related to SRFI 133), PR #382 (SRFI 19 -- if the PR isn't ok tell me), issue #367 and the new macros, but no hurry, there are other things I'll be working on (like SRFI 160) :grin:

jpellegrini:
It works! :smiley: 

```scheme
(import (srfi 160))
(list (hash-table-hash #u64(10 20 30))
       (hash-table-hash #s64(10 20 30))
       (hash-table-hash #u16(10 20 30))
       (hash-table-hash #s16(10 20 30))
       (hash-table-hash #u8(10 20 30))
       (hash-table-hash #s8(10 20 30))
       (hash-table-hash #f32(10.0 20.0 30.0))
       (hash-table-hash #f64(10.0 20.0 30.0))
       (hash-table-hash #c64(10.0 20.0 30.0)) 
       (hash-table-hash #c128(10.0 20.0 30.0)))
       (hash-table-hash #c64(1-1.0i 0.0+2i 3.0-3.0i))
       (hash-table-hash #c128(1-1.0i 0.0+2i 3.0-3.0i))
=> (30135 34470 12659 16994 3985 8320 43016 38681 51690 47355 45032 40153)
```

PR #399, opened 2022-06-22
Add SRFI-160: Homogeneous numeric vector libraries

jpellegrini:
Hi @egallesio !

Here's R7RS Large's  SRFI-160,  Homogeneous numeric vector libraries

This one was quite an adventure...

* This SRFI has sublibraries, so for the first time we have autotools descending into a subdir of `lib/srfi`.
* Several sublibraries are almost identical, so we generate them from a template. Using `sed` (!) in the `Makefile`.
* A relatively small C core can perform a large amount of operations that are described in the SRFI.
* Exports in the convenience module `srfi/160` are done programatically, and not using `export`.

# Summary

Adds SRFI-160, which extends SRFI 4, bringing ideas from SRFIs 152 and 133.

The SRFI requires the creation of the libraries:

- `(srfi 160 base)`,  with basic procedures
- `(srfi 160 s8)`,
  `(srfi 160 u8)`,
  `(srfi 160 s16)`,
  ...,
  `(srfi 160 c128)`,
  with more specific procedures

The modules are named `srfi/160/base`, `srfi/160/s8`, `srfi/160/u8`,  and so on.

- A `srfi/160` convenience module is also provided.

Since the motivation for someone to use uniform vectors is likely to get shorter execution times, most of the procedures
are implemented in C in `160/base.c`.

The procedures in base.c are selectively exported. The C functions accept an integer as type tag, and hence work
for all uniform vectors.

# Changes in other parts of STklos

* `reader.c`: now accepts `#c64` and `#c128` syntax for uniform complex vectors

* `stklos.h`: `UVECT_C64` and `UVECT_C128` are defined

* `uvector.c`:
  - `vector_element_size` `uvector_set` `uvector_ref` `makeuvect` now handle uniform complex vectors. Complexes are stored with their real and imaginary parts together
  - `UVECT_C64` and `UVECT_C128` are used;
  - vector_element_size uvector_set uvector_ref makeuvect are not static anymore, since they're used in base.c
  - Cmake_complex was copied from number.c
  - an `EXTERN_PRIMITIVE` for exact->inexact was added
  - now includes `<float.h>` in order to use `FLT_MAX`.

* `lib/srfi/Makefile.am`, `configure.ac` were adjusted so that the Makefile in `lib/srfi/` recursively goes into `lib/srfi/160`, *and* cleans it up when we call `Make clean`.

# SRFI installation

The SRFI defines several libraries (but not "`(srfi 160)`":

```
(srfi 160 base)
(srfi 160 s8)
(srfi 160 u8)
...
(srfi 160 c128)
```

So we need a directory structure like

```
srfi/160
srfi/160/base.c
srfi/160/base.stk
srfi/160/tagvector-template.stk
srfi/160/Makefile.am
```

That automake will later install to

```
$PREFIX/lib/base.so
$PREFIX/share/base.ostk
$PREFIX/share/s8.ostk
$PREFIX/share/u8.ostk
$PREFIX/share/s16.ostk
...
$PREFIX/share/c128.ostk
```

For this to work, I changed `lib/srfi/Makefile.am` a bit, and also added `lib/srfi/160/Makefile` to `AC_CONFIG_FILES` in `configure.ac`.

The files produced are installed in their proper places.

# Convenience  SRFI-160 module

We also add a convenience `srfi/160` module, so the user can "`(import (srfi 160))`" and have all the SRFI-160
libraries loaded.

Since "`srfi/160.stk`" was not being used by the rest of the implementation, this module was implemented there.

Technically, this is interesting:

The original `TAG.stk` files are generated from the `tagvector-template.stk` file.

```
tagvector-template.stk
  |
   \_ s8.stk
   |_ u8.stk
   ...
   |_ c128.stk
```

Each one of those exports a large quantity of symbols, so having them all in exports clauses in a single file would be somewhat clumsy.

We do this: `srfi/160` will

- import `srfi/160/base` and all `srfi/160/TAG` modules
- export the same symbols as `srfi/160/base`
- from the template export list in `tagvector-template.stk`, a new list of symbols is obtained, and we programatically go through it, adding it to the module's export list.

# Tests

The tests from the reference implementation were adapted for STklos. They need the `(srfi 160)` module, which is not part of the spec, but was added for convenience.

# PENDING ISSUES

* Without PR #398 the tests will trigger two errors. Nothing serious.
* ~We use the default hash function, which is not nie for uvectors (see issue #394)~ (fixed!)
* I'm not sure about the NaNs in uniform vectors. Should `uvector=` use `memcpy`? Different schemes will treat nan's differently: see the section "Is (/ 0.0 0.0) the same as +nan.0?" here: https://github.com/jpellegrini/surveys/blob/master/surveys/non-finite-numbers.md

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -96,6 +96,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-154: First-class dynamic extents
     - SRFI-156: Syntactic combiners for binary predicates
     - SRFI-158: Generators and Accumulators
+    - SRFI-160: Homogeneous numeric vector libraries
     - SRFI-161: Unifiable Boxes
     - SRFI-169: Underscores in numbers
     - SRFI-170: POSIX API
--- a/configure
+++ b/configure
@@ -7417,7 +7417,7 @@ esac
 
 
 
-ac_config_files="$ac_config_files Makefile src/Makefile src/extraconf.h doc/Makefile doc/refman/Makefile doc/vm/Makefile doc/hacking/Makefile lib/Makefile utils/Makefile utils/stklos-config utils/stklos-script examples/Makefile lib/Match.d/Makefile lib/SILex.d/Makefile lib/Lalr.d/Makefile lib/ScmPkg.d/Makefile lib/scheme/Makefile lib/srfi/Makefile lib/stklos/Makefile lib/streams/Makefile tests/Makefile doc/stklos.1 doc/stklos-config.1 doc/stklos-compile.1 doc/stklos-genlex.1 doc/stklos-pkg.1 doc/stklos-script.1 pkgman/Makefile extensions/Makefile extensions/gtklos/Makefile extensions/gtklos/lib/stklos/Makefile extensions/gtklos/lib/stklos/gtklos-config.h extensions/gtklos/lib/stklos/gtklos.stk extensions/gtklos/demos/Makefile extensions/curl/Makefile extensions/curl/lib/stklos/Makefile extensions/curl/demos/Makefile extensions/curl/doc/Makefile extensions/fuse/Makefile extensions/fuse/lib/stklos/Makefile extensions/fuse/demos/Makefile"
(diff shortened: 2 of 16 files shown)


jpellegrini:
(Next SRFI could be 178, bitvectors, which has an API similar (but not completely) to this one. But since representation of bitvectors is different, some parts of the code (using memcpy and friends) would not be directly translatable... But it wouldn't be hard either.

jpellegrini:
One more comment: I have measured the times for using the already existent `uvector_set` and `uvector_get` functions, and  it seems that directly using memv (wrapped in proper macros, of course) is 2x faster, so I did that.

jpellegrini:
Sorry, had to force-push.
There were a couple of mistakes in Makefiles, I fixed them.
Also optimized some code that I thought I had done already.

jpellegrini:
One thing: this will crash on some tests without the fix in #401

jpellegrini:
Now that #398 has been merged, this one runs smoothly and passes all tests! :)

jpellegrini:
Great @egallesio !
I'll be working on SRFI-146 later... But I think I have made too many PRs already -- I'll wait so you can work on them.

PR #400, opened 2022-06-22
Remove extraneous parenthesis

jpellegrini:
Commit 4f41e8b55e784c4f740c7db09b3d33677a970e58 changed this file, and included an extra closing parens, actually closing 
 the module definition earlier than what it was done before. This patch reverts that.


egallesio:
Already corrected this one, but obviously I messed up things when merging things. Thanks for your (always precious) help

PR #404, opened 2022-06-24
Add SRFI-125: Intermediate Hash Tables

jpellegrini:
Hi @egallesio !

Here's R7RS large hashtables... There are still some issues, and it actually needs immutable hashtables (PR #383 ).

- This implmentation directly uses STklos' hashtable API (and doesn't even load SRFI 69)

- In srfis.stk, the alias given is the one defined for it in R7RS/Large Tangerine Edition ("hash-table"). But this doesn't (yet) make it possible to do (import (scheme hash-table)). Maybe later.

Not perfect yet:

- ~SRFI 125 requires immutable hashtbles, which STklos doesn't yet have;~ we have those now :)

- Maybe ~`hash-table-clear!`~ and `hash-table-empty-copy` could be implemented as primitives? They do excessive work because there's no way to remove all data in a hash table. But would it be OK to implement them in core, or d owe create a C sublibrary just for them? UPDATE: I added a patch for `hash-table-clear!`. I'm not sure if having a primitive for `hash-table-empty-copy` is that interesting...

- ~The implementation has some deprecation warning code. Maybe it'd be better to remove it~ ok, already did...

- For some reason, when we define `make-hash-table` in the module, then STklos internal code starts using it and gets confused because the procedure signature changed. The solution I found was to rename all symbols inernally, prepending `s125:` to them, and then renaming on export -- but this is a workaround...

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -77,6 +77,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-116: Immutable List Library
     - SRFI-117: Queues based on lists
     - SRFI-118: Simple adjustable-size strings
+    - SRFI-125: Intermediate hash tables
     - SRFI-127: Lazy Sequences
     - SRFI-128: Comparators (reduced)
     - SRFI-129: Titlecase procedures
--- a/lib/srfi/125.stk
+++ b/lib/srfi/125.stk
@@ -1,0 +1,759 @@
+;;;;
+;;;; 125.stk            -- Implementation of SRFI-125
+;;;;
+;;;; Copyright © 2022 Jeronimo Pellegrini - <j_p@aleph0.info>
+;;;;
+;;;;
+;;;; This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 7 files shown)


jpellegrini:
There are some procedures missing in this one.

jpellegrini:
@egallesio I have made adjustments, and with the new immutable hash tables, this passes all tests.
I have also removed the debugging statements.

jpellegrini:
> Anyway, do you know a reliable source for the already accepted libraries for R7RS large. For now, I use the associated category in the SRFI site, but don't know if there is a “official” list somewhere.

Hi @egallesio .
I have been using this:

https://github.com/johnwcowan/r7rs-work/blob/master/R7RSHomePage.md

Red edition: https://github.com/johnwcowan/r7rs-work/blob/master/R7RSHomePage.md
Tangering edition: https://github.com/johnwcowan/r7rs-work/blob/master/TangerineEdition.md

egallesio:
> I  have been using this:

Thanks a lot @jpellegrini.

Issue #386, opened 2022-05-29
Autoloading of SRFI 27 works, but makes it slower (?)


jpellegrini:
I really don't uderstand this one... Look:

In this first example, everything is fine, because I explicitly loaded SRFI 27:

```
(import (srfi 27))       ;; or (require "srfi-27"), doesn't make a difference for this test
(define size 10000)
(define v (make-vector size))
(repeat 10
        (time
         (dotimes (i size)
           (vector-set! v i (random-real)))))

Elapsed time: 16.462 ms
Elapsed time: 26.465 ms
Elapsed time: 28.466 ms
Elapsed time: 17.06 ms
Elapsed time: 22.249 ms
Elapsed time: 16.217 ms
Elapsed time: 28.235 ms
Elapsed time: 16.714 ms
Elapsed time: 22.987 ms
Elapsed time: 16.395 ms
```

Now, if I restart STklos and try again without `(import (srfi 27))`, then:

```
(define size 10000)
(define v (make-vector size))
(repeat 10
        (time
         (dotimes (i size)
           (vector-set! v i (random-real)))))

Elapsed time: 2179.292 ms
Elapsed time: 9195.538 ms
Elapsed time: 20910.795 ms
Elapsed time: 33241.144 ms
Elapsed time: 48143.346 ms
Elapsed time: 71543.498 ms
Elapsed time: 101263.815 ms
```

So, the first run was already slower, and each one got slower than the previous...

I have grepped for `random-real` in the sources, and it seems to only be defined in SRFI 27.
Is this related to the autoload code?

I have tried the same with another autoloaded module, and it worked fine:
```
(require "env")
(define e #f)

(repeat 10
        (time
         (repeat 100000
                 (set! e (scheme-report-environment 5)))))
Elapsed time: 48.066 ms
Elapsed time: 51.876 ms
Elapsed time: 55.204 ms
Elapsed time: 54.275 ms
Elapsed time: 52.5389999999999 ms
Elapsed time: 44.433 ms
Elapsed time: 48.787 ms
Elapsed time: 52.029 ms
Elapsed time: 49.721 ms
Elapsed time: 50.228 ms
```

Without the `require`:
```
(define e #f)

(repeat 10
        (time
         (repeat 100000
                 (set! e (scheme-report-environment 5)))))
Elapsed time: 60.793 ms
Elapsed time: 45.49 ms
Elapsed time: 55.856 ms
Elapsed time: 52.1 ms
Elapsed time: 50.919 ms
[...]

PR #405, opened 2022-06-25
autoload: delete wrapper after autoloading (this patch should not work, but... it does!?)

jpellegrini:
@egallesio - this is weird.
This patch is not suppsoed to work,  but it works!

Autoload was keeping a wrapper over the procedures to be loaded, and this was slowing down the execution in some cases. This changes the autoload macro to do remember if the file was really loaded (if the binding changed) and issue a `(set! name binding)` *outside* the definition of the wrapper:

```scheme
(macro-expand '(autoload "srfi/27"      random-integer random-real))

(begin
  (begin
    (define %%random-integer-found3 #f)
    (define random-integer
      (lambda G1 (let ((G2 random-integer))
              (require "srfi/27")
              (if (eq? G2 random-integer)
                  (error 'autoload "~S has not been defined in ~S" 'random-integer "srfi/27")
                  (begin (set! %%random-integer-found3 #t)
                         (apply random-integer G1))))))
    (when %%random-integer-found3
      (define random-integer %%random-integer-found3)))
  
  (begin
    (define %%random-real-found4 #f)
    (define random-real
      (lambda G1 (let ((G2 random-real))
              (require "srfi/27")
              (if (eq? G2 random-real)
                  (error 'autoload "~S has not been defined in ~S"
                         'random-real "srfi/27")
                  (begin (set! %%random-real-found4 #t)
                         (apply random-real G1))))))
    (when %%random-real-found4
      (define random-real %%random-real-found4))))
```

However, since the `(when %%random-integer-found3 ...)` is ***outside*** the `define`, it should ***not*** be triggered when the wrapper is called.  So it should not work at all!

But it does!

```
;;;; Fresh start of STklos
stklos> (define rr random-real)
;; rr
stklos> (random-real)
0.786820954867802
stklos> (define rr2 random-real)
;; rr2
stklos> (eq? rr rr2)
#f
stklos> (eq? random-real (symbol-value 'random-real (find-module 'srfi/27)))
#t
```

Also, the slowness problem reported in #386 doesn't happen anymore:

```scheme
;;;; Fresh start of STklos
stklos> (define size 10000)
(define v (make-vector size))
(repeat 10
        (time
         (dotimes (i size)
           (vector-set! v i (random-real)))))

;; size
;; v
Elapsed time: 86.519 ms
Elapsed time: 86.45 ms
Elapsed time: 79.391 ms
Elapsed time: 100.882 ms
Elapsed time: 95.914 ms
Elapsed time: 86.347 ms
Elapsed time: 79.581 ms
Elapsed time: 97.2769999999999 ms
Elapsed time: 85.1940000000001 ms
Elapsed time: 103.131 ms
```

I'm kind of happy that it worked, but confused. *Why* does it work? My impression is that one'd need either the condition system or continuations to get this working properly (or perhaps a generic method with something to run after it's called)... (?)

### DIFF
--- a/lib/load.stk
+++ b/lib/load.stk
@@ -515,13 +515,24 @@ doc>
         (old  (gensym)))
     `(begin
        ,@(map (lambda (x)
-                `(define ,x (lambda ,args
-                              (let ((,old ,x))
-                                (require ,file)
-                                (if (eq? ,old ,x)
-                                    (error 'autoload "~S has not been defined in ~S"
-                                           ',x ,file)
-                                    (apply ,x ,args))))))
+                (let ((x-found (gensym
+                                  (string-append "%%"
+                                                 (symbol->string x)
+                                                 "-found"))))
+                  (list 'begin
+                   `(define ,x-found #f)
+                   `(define ,x (lambda ,args
+                                 (let ((,old ,x))
+                                   (require ,file)
+                                   (if (eq? ,old ,x)
+                                       (error 'autoload "~S has not been defined in ~S"
+                                              ',x ,file)
+                                       (begin (set! ,x-found #t)
+                                              (apply ,x ,args))))))
+                   ;; If it was found, remove the previous definition of x
+                   ;; and bind it to its proper value in its module:
+                   `(when ,x-found
+                      (define ,x ,x-found)))))
               symbols))))
 
 ; LocalWords:  repl autoload prepended


jpellegrini:
Also, there is a bug: isntead of
```scheme
(set! ,x-found #t)
```
it should be
```scheme
(set! ,x-found ,x)
```
But even then it works! And I can see the gensym'd variable on the repl:

```
stklos> %%random-real-found479 
#f
stklos> (random-real)
0.786820954867802
stklos> %%random-real-found479 
#t
stklos> 
```

jpellegrini:
Oh wait!
It works even without my modification!
@egallesio - did you fix #386 with your recent changes? It looks so!

jpellegrini:
I see there were indeed changes to the autoload code. Cool, one more issue fixed! :)

Issue #379, opened 2022-05-16
stklos-config  -L not working?


jpellegrini:
Hello,

It looks like `stklos-config` works perfectly, except for the `-L` option:
```
$ stklos-config  -p
/usr/local
$ stklos-config  -v
1.70
$ stklos-config  -c
gcc -fpic -nostdlib -I/usr/local/include/stklos -I/usr/local/include/stklos/gc
$ stklos-config  -l
ld -shared -o

$ stklos-config  -L
@LIB_SUMMARY@
```

jpellegrini:
The variable is not being set anywhere...
```
$ grep -r LIB_SUM *
utils/stklos-config.in:            echo "@LIB_SUMMARY@"
utils/stklos-config:            echo "@LIB_SUMMARY@"
```

jpellegrini:
It works :)
```
$ stklos-config -L
libffi libpcre libgmp libgc
```
Problem fixed...

PR #406, opened 2022-06-26
Fix variable name in stklos-config.in

jpellegrini:
Hi @egallesio - I'm not sure this is what you indended (was it a typo/thinko? `LIB_SUMMARY` <-> `CONF_SUMMARY`?)

Could this be a resolution to #379 ?

Seems to work now, I'm not sure it's the exact output you wanted:

```
$ stklos-config -L
(:system (libffi libpcre libgmp libgc ) :compiled () :configure ( "CC=gcc" "CFLAGS=-g -O2"))
```

### DIFF
--- a/utils/stklos-config.in
+++ b/utils/stklos-config.in
@@ -88,7 +88,7 @@ while test $# -gt 0; do
             echo @THREADS@
             ;;
         --libraries|-L)
-            echo "@LIB_SUMMARY@"
+            echo "@CONF_SUMMARY@"
             ;;
         --help|-h|-\?)
             usage 0 1>&2


lassik:
This may be related:

```
$ stklos-config -L
@LIB_SUMMARY@
```

lassik:
LGTM

egallesio:
Hi @jpellegrini,
I have merged your PR and modified a bit the `stklos-config` command` after that: I have added a new option to distinguish the compiled and system libraries. Parsing the result will be simpler in shell-like programs. If a Scheme like form is more suitable for the user, I suppose he/she can use the result of `stklos -V`. 

This should also close issue #379.

PR #408, opened 2022-06-27
Add tests for hash tables

jpellegrini:
The hash tables are already tested with the tests for SRFI 69, but I though it would be interesting to test them without the changes introduced by the SRFI.


egallesio:
Thanks again for your work. It is merged now.
I had to displace the loading of the file, since `hash-table-map` is changed by SRFI-125.

PR #411, opened 2022-07-03
Fix `exact-integer-sqrt`, and add `integer-length`

jpellegrini:
Hi @egallesio !

I noticed that STklos was apparently too slow and timed out in the "pi" benchmart at https://github.com/ecraven/r7rs-benchmarks - so I did some debugging, and it turns out thaat the integer square root algorithm had two issues.
I fixed them, and in doing that I also implemented `integer-length` (different from `fx-length`, `integer-length` also works for bignums).

The two problems:

1. In the stop condition of the Newton-Raphson method, the implementation was checking if `(= xn xn1)`. It should be `(>= xn1 xn)`, and `xn` should be returned. In one case, the method oscillated between $xn > xn1$ and $xn < xn1$, never stopping (this happens when $n+1$ is a perfect square). 

1. The initial guess used for `xn` was `1`, which is too low. In the "pi" benchmark it makes a huge difference (from 4 minutes to 2 seconds!) We now use a better estimate,
$$2^{\lfloor (\log_2 n)/2+1\rfloor}$$ 
or  `(expt 2 (+ 1 (quotient (integer-length n) 2)))`, which is the least power of two bigger than $\sqrt{n}$.

For that we used the new `integer-length` primitive.

I implemented `integer-length` because just calculating the integer length with `(log n 2)` would not work for very large numbers, as the `log` will be computed as `+inf.0`. The `integer-length` primitive is fast and doesn't need to do computations that fit in doubles, such as `log`.

In short, these commits bring a much faster `exact-integer-sqrt` procedure (and fix the " $n+1$ is perfect square" case).


jpellegrini:
I forgot to mention: I am aware that `integer-length` is present in SRFI 60, but seems to exist in the core of in Chez, Gambit, Gauche, Guile, Kawa and Sagittarius. Also, the version in STklos SRFI-60 is not very efficient. Perhaps it could be replaced by this new primitive?

jpellegrini:
I have included one commit that changes SRFI 60 to use the new version of `integer-length`.

jpellegrini:
Hi @egallesio !

I think I went right into the details in the PR, without explaining what is currently not working.

To immediately see what the problem is:

```
stklos> (exact-integer-sqrt 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999)
```
loops forever, because the argument is a bignum that is also a square minus one.

egallesio:
Hi @jpellegrini,

Thanks a lot for the PR and for the explanations too. The speed improvement is impressive really (when it doesn't loop :wink: )

This is merged now.

jpellegrini:
> when it doesn't loop

It shouldn't anymore, this PR fixes it! :)

```
stklos> (exact-integer-sqrt 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999)
99999999999999999999999999999999999999999999999999
199999999999999999999999999999999999999999999999998
```

egallesio:
> It shouldn't anymore, this PR fixes it! :)

I've seen that. This is why I said it is way long faster, specially when the old implementation looped :rofl:

PR #413, opened 2022-07-06
Fix error message in `make-{polar,rectangular}`

jpellegrini:
When one calls
`(make-polar 1 2+1i)`
or
`(make-rectangular 2+1i 3)`

STklos says `2+1i is a bad number` - which is not very clear.
A better error message here would be
 `2+1i is not a real number`,
which we adjust here (a function that prints this error message already exists, but wasn't being used in these procedures).

This patch preserves the "bad number" message when an argument really isn't a number:

```
stklos> (make-rectangular "x" 2)
**** Error:
make-rectangular: `"x"' is a bad number
```


egallesio:
Merged. Thanks, @jpellegrini.

PR #414, opened 2022-07-14
Fix small glitch in `STk_utf8_grab_char`

jpellegrini:
When the first byte is $>=$ `0xf0`, the algorithm was decoding incorrectly:

```scheme
(list->string (list (integer->char 65537)))
 => "𐀁"        ;; Ok

(string->list (list->string (list (integer->char 65537))))
 => (#\Ё)      ;; Oops! list->string and string->list
               ;; do not work as inverses, because the
               ;; character is decoded incorrectly in
               ;; string->list (and string-ref too)
```

This fixes the problem. Also included some tests: the full Unicode range (excluding surrogates) is verified to encode and decode correctly, as per

```scheme
(equal? (string->list (list->string (list (integer->char i))))
                                    (list (integer->char i)))
```

### DIFF
--- a/src/utf8.c
+++ b/src/utf8.c
@@ -68,8 +68,8 @@ char *STk_utf8_grab_char(char *str, uint32_t *c) /* result = pos. after current 
     return str + 3;
   }
 
-  *c = ((buff[0] & 0x0f) << 16) +
-       ((buff[1] & 0x3f) <<  6) +
+  *c = ((buff[0] & 0x07) << 18) +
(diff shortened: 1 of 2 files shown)


jpellegrini:
hi @egallesio !
On top of this one I have an implementation of the R7RS Large `(scheme bytevectors)` library. It's not one I really like, but it's R7RS-Large... So I thought I'd implement it. So as soon as that small UTF-8 problem is fixed, I'd have another PR :)

PR #415, opened 2022-07-17
Add R7RS-Large bytevectors

jpellegrini:
These are the same as R6RS-bytevectors, which were chosen as part of R7RS-Large.

* As much as possible, operations are performed in C for performance (even the trivial `bytevector=?` was written in C).
* A known issue is that this introduces an incompatibility with R7RS (in `vector-copy!`), but since it's in a library, one can always rename when importing (which seems to be the consensus solution).
* Since this is not a SRFI, it includes the library in the `scheme` subdirectory (is this OK?)

~This depends on PR #414~ (that PR was already accepted)

### DIFF
--- a/lib/scheme/Makefile.am
+++ b/lib/scheme/Makefile.am
@@ -102,9 +102,9 @@ LARGE_OBJS = list.ostk       \
              division.ostk   \
              bitwise.ostk
 
-LARGE_C     = ilist.c     flonum.c     sort.c     vector.c
-LARGE_C_STK = ilist.stk   flonum.stk   sort.stk   vector.stk
-LARGE_SHOBJ = ilist.$(SO) flonum.$(SO) sort.$(SO) vector.$(SO)
+LARGE_C     = ilist.c     flonum.c     sort.c     vector.c     bytevector.c
+LARGE_C_STK = ilist.stk   flonum.stk   sort.stk   vector.stk   bytevector.stk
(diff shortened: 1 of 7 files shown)


jpellegrini:
Hi @egallesio -- I think I have adapter this correctly for the library reorganization. Tell me if it's not good.

egallesio:
Hi @jpellegrini,

>  I think I have adapter this correctly for the library reorganization. Tell me if it's not good.

It is perfect and tests are running OK. Thanks for this PR.

However, the compilation of `bytevector.c` produces numerous warnings when compiled with `-Wall -Wextra`. Most of the warnings are relative to comparisons of signed and unsigned values and are simple to correct. Some are more difficult, and I'm not sure of my corrections (Note:this file cannot be compiled with `clang` :disappointed: ). 

If you have time to look at it, it would be great. Otherwise, I will try to do it.
After the merge of your contribution, we will have nearly all the R7RS large libraries :smiley:

jpellegrini:
Oh, sorry about that. I'll take a look!

egallesio:
> Oh, sorry about that. I'll take a look!

no problem at all. Thanks for looking at it.

jpellegrini:
@egallesio I think it's all fixed now :)

jpellegrini:
I'm not yet sure about `LONG_FITS_INTEGER`... I'll take another look tomorrow,

jpellegrini:
I think it's ok now...

egallesio:
Thanks a lot @jpellegrini for your fixes.

I have also modified a bit your code since other warnings appeared when compiling with optimizations with `gcc`.

PR #416, opened 2022-07-20
Add SRFI 43 - Vector Library

jpellegrini:
This SRFI was being considered to inclusion in R7RS large when it was found that it was actually incompatible with R7RS small. Then SRFI 133 was designed, but it did not officially supersede SRFI 43.

I thought it would be interesting to have this one, just in case someone wants it (for running old code, for example).

Since there are only 6 incompatible procedures, and it's easy to rename symbols when importing and exporting R7RS libraries, there is no harm in having SRFI 43 available also.

(This one is available in Gauche, Gerbil, Guile, Larceny, Sagittarius and perhaps others)

This implementation:

* Takes as many procedures as possible from the `SCHEME` module;
* Takes others from SRFI-133;
* And finally, provides code for the 6 incompatible procedures, `vector-map`, `vector-map!`, `vector-for-each`, `vector-fold`, `vector-fold-right`, `vector-count`.

### DIFF
--- a/lib/srfi/43.stk
+++ b/lib/srfi/43.stk
@@ -1,0 +1,400 @@
+
+;; Incompatible procedures, which are taken from the SRFI-43
+;; reference implementation:
+;;
+;; * vector-map
+;; * vector-map!
+;; * vector-for-each
+;; * vector-fold
+;; * vector-fold-right
+;; * vector-count
+
+(define-module srfi/43
+  (import (prefix (srfi 133) s133-))
+  (export
+   ;;; Constructors
+   make-vector vector
+   vector-unfold                   vector-unfold-right
+   vector-copy                     vector-reverse-copy
+   vector-append                   vector-concatenate
+
+   ;;; Predicates
+   vector?
+   vector-empty?
+   vector=
+   
+   ;;; Selectors
+   vector-ref
+   vector-length
+
+   ;;; Iteration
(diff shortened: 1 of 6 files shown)


lib/srfi/43.stk
```diff
@@ -0,0 +1,400 @@
...
+   vector-binary-search
+   vector-any                      vector-every
+   
+   ;;; Mutators
+   vector-set!
+   vector-swap!
+   vector-fill!
+   vector-reverse!
+   vector-copy!                    vector-reverse-copy!
+   vector-reverse!
```
zdimension, line 48: By curiosity, why is `vector-reverse!` present twice?

jpellegrini: > By curiosity, why is `vector-reverse!` present twice?

Hm, that's because the PR author (me) didn't notice :)
I am unable to fix it now, but will do  tomorrow.

jpellegrini: Fixed now :)

egallesio:
Finally merge it.
Thanks.

egallesio:
@jpellegrini : Sorry for the long delay to merge this PR.

jpellegrini:
> @jpellegrini : Sorry for the long delay to merge this PR.

No problem @egallesio ! I'm also a bit slow to get things done, but I'm still working on some more libraries and SRFIs.

PR #417, opened 2022-07-22
Warn to not use `-DSTK_DEBUG` on Termux/Android

jpellegrini:
Becuase it won't compile (execinfo is not supported by Bionic libc)


egallesio:
This is fixed now (at the expanse of `%c-backtrace` which displays a message when used on Android).

PR #418, opened 2022-07-24
Don't use readline completion on android

jpellegrini:
Hi @egallesio !

I'm not sure about this one. While we don't find the reason for the readline completion failure on Android, maybe we could avoid using it?  If you feel like this is OK... There it is! :)


egallesio:
Hi @jpellegrini,

I have fixed it. This is (once again) a problem with the bugged RTL_GLOBAL of Android. The fix circumvents the issue.

PR #420, opened 2022-08-09
Mention editorconfig in hacking.adoc, and use indent_size for indentation size

jpellegrini:
Hi @egallesio !
Just mentioning `.editorconfig` in `hacking.adoc`.

By the way, the file has a generic
```
indent_size              = 4
```
and a specific
```
[*.{c,h}]
tab_width = 2
```
But the last one only changes the width of the tab character. Shouldn't it have been `indent_size = 2`?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,7 +12,7 @@ charset                  = utf-8
 trim_trailing_whitespace = true
 
 [*.{c,h}]
-tab_width = 2
+indent_size = 2
 
 [Makefile{,.am}]
 indent_style = tab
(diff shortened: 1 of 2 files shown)


lassik:
> Shouldn't it have been `indent_size = 2`?

Yes. `tab_width` refers only to the ASCII tab character. It does not affect "tabbing" with spaces.

Which files are using 4-space indents?

jpellegrini:
> Which files are using 4-space indents?

Makefiles. :)

egallesio:
Thanks for this PR. 
Merged.

PR #425, opened 2022-11-23
Add SRFI 236 - Evaluating expressions in an unspecified order

jpellegrini:
Hi @egallesio !
SRFI 236 went final, and since it's trivial, I've decided to make this PR. :)

The implementation I've included is different (and simpler) from the reference implementation: it just wraps the expressions in `begin`, but appending `(void)` as last expression.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -128,3 +128,4 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-229: Tagged Procedures
     - SRFI-230: Atomic Operations
     - SRFI-233: INI files
+    - SRFI-236: Evaluating expressions in an unspecified order
--- a/doc/HTML/stklos-ref.html
+++ b/doc/HTML/stklos-ref.html
@@ -29074,7 +29074,7 @@ their use in free software.</p>
 </div>
 <div id="footer">
 <div id="footer-text">
-Last updated 2022-11-23 21:47:27 +0100
+Last updated 2022-11-23 22:52:34 -0300
 </div>
 </div>
 </body>
--- a/doc/HTML/stklos-ref.html
+++ b/doc/HTML/stklos-ref.html
@@ -4,7 +4,7 @@
 <meta charset="UTF-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta name="generator" content="Asciidoctor 2.0.18">
+<meta name="generator" content="Asciidoctor 2.0.17">
 <meta name="author" content="Erick Gallesio">
 <title>STklos Reference Manual : (version 1.70)</title>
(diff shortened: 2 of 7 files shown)


jpellegrini:
> The SRFI's sample implementation wraps them in `let` because `let` bindings are evaluated in arbitrary order. If STklos evaluates all `let` bindings in left-to-right order, the effect would be the same as `begin`.

Yes... Anyway, from the specification I understand that it can be implemented as `(begin ... (void))`. Or, without a `(void)` procedure, `(begin ... (if #f #t))`. The execution order will be the same as that of `begin`, but the SRFI doesn't forbid that.  :grin:  The extra work to execute in reverse order doesn't seem necessary...

jpellegrini:
Now that I think of it... Since the SRFI says the result is "unspecified", and only requires that the results of the expressions are discarded, even this would fit:

```scheme
(define-syntax independently
      (syntax-rules ()
	((independently expr ...)
         (begin expr ... 'cool-symbol-returned-by-independently ))))
```
or:
```scheme
(define-syntax independently
      (syntax-rules ()
	((independently expr ...)
         (begin expr ... (random-integer) ))))
```
Because the result is "unspecified" (not specified by the SRFI)... Or did I miss something?

jpellegrini:
Force-pushed to resolve a conflict...

egallesio:
We can also shuffle the expressions of the `independently` to be sure that the user cannot know how evaluation will be done. Something like:

```scheme
(define-macro  (independently . body)
  (let ((shuffle (lambda (lst)
                   (map cdr
                        (sort (map (lambda (x) (cons (random-real) x)) lst)
                              (lambda (x y) (< (car x) (car y))))))))
    `(begin ,@(shuffle body) (void)))) 
```
```scheme

tklos> (define p display)
;; p
stklos> (independtly (p 1) (p 2) (p 3) (p 4))
2314stklos> (independtly (p 1) (p 2) (p 3) (p 4))
1234stklos> (independtly (p 1) (p 2) (p 3) (p 4))
1342stklos> 
```
:laughing:.

lassik:
> Yes... Anyway, from the specification I understand that it can be implemented as `(begin ... (void))`. Or, without a `(void)` procedure, `(begin ... (if #f #t))`. The execution order will be the same as that of `begin`, but the SRFI doesn't forbid that. 😁

Yes.

> The extra work to execute in reverse order doesn't seem necessary...

The point of unspecified order is that compilers can sometimes optimize more things when they know they can change the order. (The other point is for documentation purposes.)

> Since the SRFI says the result is "unspecified", and only requires that the results of the expressions are discarded, even this would fit: [...]
> Because the result is "unspecified" (not specified by the SRFI)...

Yes, exactly. Many R7RS procedures return unspecified values and you could do the same for them.

> We can also shuffle the expressions of the independently to be sure that the user cannot know how evaluation will be done.

That's a fun idea :) And a valid implementation.

jpellegrini:
So... Which version is better?
I think the shuffling one is really nice. If it's oK, I'll adapt the PR to include it (?)

jpellegrini:
Hi @egallesio !
So... Why not use your shuffling implementation? :smile:   I've changed the PR to use it -- and also included a missing `provide`.

egallesio:
Hi @jpellegrini, 
Merged.

PR #426, opened 2022-11-30
configure: use [ ] to compare values

amirouche:
ref: https://www.shellcheck.net/wiki/SC2283

### DIFF
--- a/configure
+++ b/configure
@@ -7509,7 +7509,7 @@ else $as_nop
   HAVE_LIBEDIT="no"
 fi
 
-                if "$HAVE_LIBEDIT" = "yes" ;then
+                if [ "$HAVE_LIBEDIT" = "yes" ] ;then
                    RDLINE="-ledit";
                 fi
             fi;;


amirouche:
There is another error reported by shellcheck, I was not sure how to fix it.

That PR does not fix my problem, see https://github.com/egallesio/STklos/issues/427

lassik:
Should probably be fixed in `configure.ac` and @egallesio can then re-run autoconf to fix `configure`.

jpellegrini:
> Should probably be fixed in `configure.ac`

Good catch! I hadn't noticed the PR changed `configure` and not `configure.ac`. @amirouche would you like to change and force-push it?

egallesio:
Hi @amirouche, 

Thanks for this PR. I have applied your patch in the `configure.ac`, since it will be used to rebuild the `configure` script. I have also changed the test with square brackets with an explicit call to "test" (just in case `configure` scripts runs on a machine without the symbolic link on `/bin/test`; this is very few unlikely now, but was current several years ago)

PR #428, opened 2022-11-30
Priorize stklos specific extensions

amirouche:


egallesio:
Hi @amirouche,

I have applied your PR, even if I plan soon to I intend to abandon the support of ScmPkg packages in STklos. Indeed, even if I think that the idea was interesting, I have to admit that it was not very successful :wink: and a better idea now would probably be to use a system like Akku or Snow.

PR #430, opened 2022-12-11
Adds SRFI 228: Composing Comparators

jpellegrini:
And also update the SRFI status in `srfis.stk`...


egallesio:
Wow @jpellegrini, this SRFI was just finalized! 
Thanks for this PR.

jpellegrini:
> Wow @jpellegrini, this SRFI was just finalized!

Yes - that was an easy one :grin: 

I'm still working on themappins SRFI, but it's a busy time now (end of year), so I'm still a bit slow. But it'll work.

PR #438, opened 2023-01-18
More SRFI 176 tweaks

lassik:
- Add `r7rs` to supported languages. Support seems good enough to declare it already.
- Change `stklos.threads` to `threads`. Added `threads` to [the Registry](https://registry.scheme.org/#version-properties).
- General refactoring.


lib/bonus.stk
```diff
@@ -2278,18 +2278,17 @@ doc>
...
-         (stklos-git (%stklos-git))
-         (commit (key-get stklos-git :commit ""))
-         (srfis  (filter-map %feature->srfi (features)))
-         (git-alist (if (= (string-length commit) 0)
-                        '()
-                        `((build.git.tag ,(key-get stklos-git :tag '()))
-                          (build.git.branch ,(key-get stklos-git :branch '()))
-                          (build.git.commit ,(key-get stklos-git :commit '()))
-                          (build.git.modified ,@(key-get stklos-git :modified '()))))))
+  (define (prop name plist plist-key)
```
jpellegrini, line 2281: The `bonus.stk` module contains bits from different SRFIs and also other extra Scheme procedures. `prop` won't be user-visible, because `bonus.stk` includes it inside  a module, but maybe it would make sense to name this `SRFI-176-prop`, just to be a bit clearer?  (Or maybe I'm being pedantic -- feel free to ignore me)

lassik: It's an internal define. Therefore it should only be visible inside the `version-alist` procedure, and not to users.

jpellegrini: Ops, sorry. From the github UI it wasn't clear.

egallesio:
Oops, forgot the `thread` property. Your code refactoring makes code cleaner.
Thanks @lassik for this PR.

PR #439, opened 2023-01-19
Implement SRFI 238 (Codesets)

lassik:
Here's a working implementation of SRFI 238 which was finalized this week. It converts STklos' existing errno and signal lists into a form that works with the SRFI, while also preserving STklos' existing behavior.

Examples from the REPL:

```Scheme
stklos> (codeset-list)
(signal errno)
stklos> (codeset-symbols 'signal)
(SIGHUP SIGINT SIGQUIT SIGILL SIGTRAP SIGABRT SIGIOT SIGEMT SIGFPE SIGKILL SIGBUS SIGSEGV SIGSYS ...)
stklos> SIGINT
2
stklos> (codeset-message 'signal SIGINT)
"Interrupt"
stklos> (codeset-message 'signal 'SIGINT)
"Interrupt"
stklos> (map (lambda (code) (codeset-message 'signal code)) (codeset-symbols 'signal))
("Hangup" "Interrupt" "Quit" "Illegal instruction" "Trace/BPT trap" "Abort trap" "Abort trap" "EMT trap"
"Floating point exception" "Killed" "Bus error" "Segmentation fault" "Bad system call" "Broken pipe" ...)
```

There are some TODO comment in the code about a few missing things.

* The string buffer for the messages should grow dynamically if it doesn't have enough space. Does STklos have a ready-made buffer we can use?

* A library hasn't yet been added, so `(import (srfi 238))` does not work.

### DIFF
--- a/src/signal.c
+++ b/src/signal.c
@@ -52,13 +52,26 @@ static SCM signals[NSIG];
 
 
 /* ====================================================================== */
-struct signal_info {
-  char *name;
-  int value;
-};
 
+#define HAVE_STRSIGNAL 1  /* TODO: Use configure */
 
-static struct signal_info signal_names[] = {
+static char *our_strsignal(int sig) {
+  char *message = NULL;
+
+#if HAVE_STRSIGNAL
+  message = strsignal(sig);
+  if (message) {
+    char *tail;
+    if ((tail = strchr(message, ':'))) {
+      /* MacOS repeats signal number in message, e.g. "Interrupt: 2" */
+      *tail = '\0';
+    }
+  }
+#endif
+  return message;
+}
+
+static struct codeset_code signal_codes[] = {
   /* Posix.1 signals */
   {"SIGHUP",    SIGHUP},
   {"SIGINT",    SIGINT},
@@ -137,6 +150,9 @@ static struct signal_info signal_names[] = {
   {NULL, 0}
 };
 
+struct codeset signal_codeset = {
+  "signal", &signal_codes[0], our_strsignal, 0, 0, 0
+};
 
 static void error_bad_signal_number(SCM sig)
 {
@@ -321,14 +337,19 @@ DEFINE_PRIMITIVE("send-signal", send_signal, subr12, (SCM sig, SCM process))
 
 int STk_init_signal()
 {
+  struct codeset_code *code;
+
   // Initialize the signals table
   for (int i=0; i < NSIG; i++) {
     signals[i] = STk_true;
   }
 
   // Define the symbols assiocated to signal names
-  for (struct signal_info *p = signal_names; p->name; p++) {
-    STk_define_variable(STk_intern(p->name), MAKE_INT(p->value), STk_STklos_module);
+  codeset_init(&signal_codeset);
+  for (code = signal_codeset.codes; code->number; code++) {
+    STk_define_variable(STk_intern((char *)code->name),
+                        MAKE_INT(code->number),
+                        STk_STklos_module);
   }
 
   ADD_PRIMITIVE(initialize_signals);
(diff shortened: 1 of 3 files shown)


jpellegrini:
Hi @lassik !
> * A library hasn't yet been added, so `(import (srfi 238))` does not work.

You implemented in a way that it'll always be included in the STklos module. If you want it to be a separate library, the primitives should be defined in a separate file, `lib/srfi/439.c`, and the `Makefile` in `lib/srfi` should be updated. Also, the SRFI list in `lib/srfis.stk` needs to be changed to include the new SRFI.

jpellegrini:
> You implemented in a way that it'll always be included in the STklos module. If you want it to be a separate library, the primitives should be defined in a separate file, `lib/srfi/238.c`, and the `Makefile` in `lib/srfi` should be updated. Also, the SRFI list in `lib/srfis.stk` needs to be changed to include the new SRFI.

*And* a wrapper `lib/srfi/238.stk` should be added, because it's what STklos will try to load. A one-liner, with just `(provide srfi/238)`. After that you'll be able to `(import (srfi 238))`.  (maybe @egallesio can correct me if I'm wrong).

lassik:
> You implemented in a way that it'll always be included in the STklos module. If you want it to be a separate library, the primitives should be defined in a separate file

@egallesio should decide which way to do it. However, the errno and signal tables probably have to be included in the main module. That would make it prudent to have at least the `struct codeset` data type built in.

egallesio:
Thanks for this PR @lassik.

> The string buffer for the messages should grow dynamically if it doesn't have enough space. Does STklos have a ready-made buffer we can use?

You can use string port for that and write your strings in it. 
However, I don't understand why you want to store the strings in *a priori* in the codeset object. This will duplicate strings which are already either in the `libc`, either given by tee user if we define new codesets in Scheme (for the `'windows` example you gave in the SRFI document, the generated string would be huge).  Furthermore, it brings some complexity to the implementation. Why not grab the message when needed?

About the codeset object, it could be defined as a new Scheme object (it is simple to add a new type of object to the interpreter). I can send you a kind of template for that if you want). I think it will also ease the definition of codeset in Scheme.

egallesio:
If you don't mind, I can adapt your code to use a new object type et make it only available when using SRFI-238. I have toyed with a kind of prototype for this SRFI that I can polish before publishing it. 
Let me know if you mind that I substantially change your contribution?

lassik:
> I don't understand why you want to store the strings in _a priori_ in the codeset object.

`strerror()` and `strsignal()` overwrite a global message buffer, so a lock is needed in multi-threaded situations. Fetching all the messages separately would require lock/unlock many times; fetching them all at once requires one lock/unlock upon the first fetch.

Some measurements:

* The `signal` messages are 0.5 kilobytes.
* The `errno` messages are 2 kilobytes.
* The `windows` messages are 15 kilobytes.

Given that the `signal` and `error` messages are so few, and the API is destructive, we might as well fetch all at once?

The `windows` case could go either way.

lassik:
> If you don't mind, I can adapt your code to use a new object type et make it only available when using SRFI-238. I have toyed with a kind of prototype for this SRFI that I can polish before publishing it. Let me know if you mind that I substantially change your contribution?

Sure, it's your implementation so you should of course be free to do whatever you like :)

egallesio:
Thanks @lassik for the explanation.
I didn't notice that the `strerror` and the `strsignal` were not thread safe (I should have seen that point). I have implemented something similar to the code you proposed in this PR. However, it is a dynamically loadable module, and it doesn't modify a lot the core (the only difference is that the `signal` and `errno` tables are public). I have also added a function which permits to define a codeset in Scheme.

```scheme
 (define cs
   (make-codeset 'foo
                 '((1   OK    "Everything is OK")
                   (1   YES   #f)                            ; Other name for OK
                   (2   KO    "We have a problem")
                   (2   NO    #f)                            ; Other name for KO
                   (3   MAYBE "To be determined")
                   (404 #f    "Not found")))          ; no symbol associated for 404 
```

Storage of messages is done by an A-list. This is less efficient than the storage you used, but easier to implement  (and it permits to have "_holes_" in the coding.

PR #443, opened 2023-02-05
Optimize calls to no-argument lambdas

jpellegrini:
Hi @egallesio ! What do you think of this?

We transform `((lambda () ...))` into `(begin ...)`.                                                                                                                                                                                

If we're applying a lambda without arguments, ((lambda () ...)), we may transform it into (begin ...), which is much more efficient. Given how common it is to pass thunks to procedures, this can speed up execution considerably in cases where code is generated by macros  (although full inlining would be much better, but also more complex):

    
```
(disassemble-expr '((lambda () a b)))

000:  PREPARE-CALL             <==
001:  ENTER-LET            0   <==
003:  GLOBAL-REF           0
005:  GLOBAL-REF           1
007:  LEAVE-LET                <==
```
 
The `PREPARE-CALL`, `ENTER-LET` and `LEAVE-LET` instructions are not necessary, and will not be generated if we actually transform the lambda into a begin.
    
We do not transform if the CDDR to be null, since `(lambda ())` is not allowed.

However, this fails:

```
(require "full-syntax")
(let-syntax ((foo (syntax-rules ()
                    ((_ expr) (+ expr 1)))))
  (let ((+ *))
    (foo 3)))
```

(expected 4 but got 3)

Likely because `"full-syntax"` was not provided/loaded, but I could't see exactly why.
Anyway, rewriting the syntax system is on the TODO list, so this may be fixed in the future with a new syntax system.

Also two other failure in the test suite, but I can't reproduce them:

```
test Test #3.1 on (let-syntax ((foo (syntax-rules () ((_ expr) (+ expr 1))))) (let ((+ *)) (foo 3))) expected 4 but got 3 
test Test #3.2 on (let-syntax ((foo (syntax-rules () ((_ var) (define var 1))))) (let ((x 2)) (begin (define foo +)) (cond (else (foo x))) x)) expected 2 but got 1 
test Test #8.1 on (let - ((n (- 1))) n) expected -1 but got 1
```

I'll mark this as not yet ready (draft), but I thought it would be nice to keep it here. Maybe soon it'll be possible to merge it?

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -1214,6 +1214,35 @@ doc>
         (compiler-error 'lambda epair "bad number of parameters ~S" actuals))))
 
 
+;; DEFINING-FORM? is used in COMPILE-CALL.
+;;
+;; We need to be able to tell if a defining for is at the beginning of
+;; a lambda form. We avoid converting
+;;
+;; ( (lambda () (define ...) ...) )
+;;
+;; into
+;;
+;; (begin (define ...) ...)
+;;
+;; because the first form is wrong (internal defines are not allowed in
+;; a lambda-form), and we'd be transforming it into a correct form.
+;;
+;; We also disallow (begin ...), since
+;;
+;; (begin (begin (define ...) ...) ...)
+;;
+;; would also be correct, although
+;;
+;; ( (lambda () (begin (define ...) ...) ... ) )
+;;
+;; is not.
+;;
+;; For that, we need the DEFINING-FORM? procedure:
+(define (defining-form? f)
+  ;; Problem: what if the user shadows "define" or similar?
+  (memq f '(define begin define-syntax define-macro)))
+
 (define (compile-call args env tail?)
   (let* ((fct     (car args))
          (actuals (cdr args))
@@ -1247,7 +1276,8 @@ doc>
           ;; However, we do not transform if the CDDR to be null, since (lambda ()) is not
           ;; allowed.
           (if (and (null? (cadr fct))
-                   (not (null? (cddr fct))))
+                   (not (null? (cddr fct)))
+                   (not (defining-form? (caddr fct))))
               (compile-body (cddr fct) env args tail?)
               (compile-lambda-call fct actuals len env ep tail?)))
 
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -1219,7 +1219,7 @@ doc>
          (actuals (cdr args))
          (len     (length actuals)))
     (if (and (pair? fct) (or (eq? (car fct) 'lambda)
-                          (eq? (car fct) '|λ|)))
+                             (eq? (car fct) '|λ|)))
         ;; fct is (lambda (...) ...)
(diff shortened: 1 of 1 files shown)


jpellegrini:
> (expected 4 but got 3)

Indeed. With this patch, `require` doesn't seem to load `full-syntax`, but a direct load works!

```
stklos> (let-syntax ((foo (syntax-rules ()
                    ((_ expr) (+ expr 1)))))
  (let ((+ *))
    (foo 3)))
3
stklos> (require "full-syntax")
WARNING: "full-syntax" was not provided
"full-syntax"
stklos> (let-syntax ((foo (syntax-rules ()
                    ((_ expr) (+ expr 1)))))
  (let ((+ *))
    (foo 3)))
3
stklos> (load "pkg/scheme/STklos-jpellegrini/lib/full-syntax.stk")
stklos> (let-syntax ((foo (syntax-rules ()
                    ((_ expr) (+ expr 1)))))
  (let ((+ *))
    (foo 3)))
4
```

(Why???)

egallesio:
Hi @jpellegrini ,

Not have really the time to look yet at this PR. However, we can not always substitute thi scall by a `begin` (in particular if there are definitions in the body

```scheme
stklos> ((lambda () (define a 1) a))
1
stklos> a
**** Error:
%execute: variable `a' unbound
```

whereas 

```scheme
stklos> (begin  (define a 1) a)
1
stklos> a
1
```

jpellegrini:
Hi @egallesio !
Ok, that's right. So perhaps if "_it's a lambda in which the first form is not a defining form_" could work...

jpellegrini:
I have added some small fix to the problem you pointed out  (which may be problematic, if the user redefines `define`).

egallesio:
> Hi @egallesio ! Ok, that's right. So perhaps if "_it's a lambda in which the first form is not a defining form_" could work...

and not a `begin`. For instance, `(begin (begin (define a 1)) a)` . Nobody will write that, but this could be easily produced by a macro expansion.

egallesio:
Ok just placing the symbol 'begin in the list of defining-form? should be OK.

jpellegrini:
> Ok just placing the symbol 'begin in the list of defining-form? should be OK.

Done! :)

jpellegrini:
Oooh, wait. I don't know where those errors come from, but I have checked out a clean copy of my branch, and it now passes all tests!   It actually works! :laughing: 

Ready to include in STklos, if you feel like it's good!

egallesio:
Hi @jpellegrini,

I was wrong: the function `compile-body` takes care of the internal definitions. So your initial commit was OK.  However, the PR code doesn't detects incorrect calls such as  `[(lambda() 1) 2 3 4]`.

Consequently, I have modified the compiler function compile-lambda-call directly to take care of such calls (and `let` without bindings).

Note that, the call `((lambda () ...))` are rare (only two occurrences when doing a complete bootstrap + all the tests). Generally these calls are written as `(let () ...)`, which are far more frequent.
Old code optimized the latter only.

PR #444, opened 2023-02-06
Declare support for SRFI 244

jpellegrini:
It's `define-values`, which is part of R7RS; the SRFI is intended for R6RS systems. But since R7RS systems do support it out of the box, advertising it would seem the right thing to do.


egallesio:
Merged!

egallesio:
Added lib/srfi/244.stk wot permit an explicit import of this library

Issue #447, opened 2023-02-10
Support cond-expand forms in R7RS libraries


Zambito1:
`cond-expand` should be supported as a top level form in R7RS libraries. It should behave similar to `include-library-declarations` in the sense that the body of the `cond-expand` is spliced directly into the library definition. This allows for things like conditional `(import ...)` expressions.

egallesio:
Hi @Zambito1, 

You are right. One point that I can delete from my TODO list :smiley:. Seen that you have two PRs. I'll have a look at it very soon. Thanks a lot.

Zambito1:
Of course! Let me know if there are changes you would like me to make or if you have any trouble with those PRs. Particularly the `cond-expand` one feels very hacky but I think it is functionally correct :D

egallesio:
I think I have corrected both issues with my last commits.
  * about `include-library-declarations`, The final code is more or less your code (I have just used the existing function `port->sexp-list` for reading the library code). 
  * about `cond-expand`, I have avoided the use of `eval` and have implemented it differently: all cond-expands are collected and expanded at first, as proposed by the R7RS document. 
 
Thanks again for your help.

Zambito1:
Great, thanks for this!

PR #448, opened 2023-02-10
Add support for cond-expand in library definitions

Zambito1:
This branches off of #446, and should be merged after that PR.


Zambito1:
Resolves #447

egallesio:
Implemented a bit differently (see comment in Issue #447).

PR #449, opened 2023-02-11
Fixes to SRFI-19

jpellegrini:
- The `copy-time` procedure would call `make-time` with parameters `#f #f #f`, but the `make-time` procedure only accepts proper integers for second and nanosecond

- `time-monotonic->date` should do the same as `time-tai->date`, but was not, so the results were wrong

- We now check the arguments to `make-date` (`make-time` already did this).

Added tests for several more cases.

### DIFF
--- a/lib/srfi/19.stk
+++ b/lib/srfi/19.stk
@@ -205,11 +205,10 @@
 ;;;
 
 (define (copy-time time)
-  (let ((ntime (make-time #f #f #f)))
-    (set-time-type! ntime (time-type time))
-    (set-time-second! ntime (time-second time))
-    (set-time-nanosecond! ntime (time-nanosecond time))
-    ntime))
+  (let ((type (time-type time))
+        (second (time-second time))
+        (nanosecond (time-nanosecond time)))
+    (make-time type nanosecond second)))
 
 ;;;
 ;;; CURRENT-TIME VARIATIONS
@@ -701,8 +700,11 @@ doc>
   (tm:time->date time tz-offset time-utc))
 
 ;; again, time-monotonic is the same as time tai
+;; we just use time-tai->date
 (define (time-monotonic->date time . tz-offset)
-  (tm:time->date time tz-offset time-monotonic))
+  (if (null? tz-offset)
+      (time-tai->date (time-monotonic->time-tai time))
+      (time-tai->date (time-monotonic->time-tai time) (car tz-offset))))
 
 
 ;;;
--- a/lib/srfi/19.stk
+++ b/lib/srfi/19.stk
@@ -205,10 +205,7 @@
 ;;;
 
 (define (copy-time time)
-  (let ((type (time-type time))
-        (second (time-second time))
-        (nanosecond (time-nanosecond time)))
-    (make-time type nanosecond second)))
+  (make-time (time-type time) (time-nanosecond time) (time-second time)))
 
 ;;;
 ;;; CURRENT-TIME VARIATIONS
--- a/lib/time.stk
+++ b/lib/time.stk
@@ -420,7 +420,22 @@ doc>
                         (day 1) (month 1) (year 1970) (zone-offset #f))
   (%make-date nanosecond second minute hour day month year zone-offset))
 
+(define (check name arg min max)
+  (unless (and (integer? arg)
+               (<= min arg max))
+    (error "bad ~s ~s" name arg)))
+
 (define (%make-date   nanosecond second minute hour day month year zone-offset)
+  (check 'nanosecond nanosecond 0 999999999)
+  (check 'second second 0 60)
+  (check 'minute minute 0 59)
+  (check 'hour hour 0 23)
+  (check 'day day 1 31)
+  (check 'month month 1 12)
+  (unless (integer? year) (error "bad year ~s" year))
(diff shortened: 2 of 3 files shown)


jpellegrini:
One more thing...

`copy-time` was copying its arguments, not using them directly:

```scheme
(define (copy-time time)
  (let ((type (time-type time))
        (second (time-second time))
        (nanosecond (time-nanosecond time)))
    (make-time type nanosecond second)))
```

This would make sense if the three arguments were not atomic objects (so a truly new copy would be made), but currently they're not (they're a symbol and two integers), so I've added a commit that optimizes this to

```scheme
(define (copy-time time)
  (make-time (time-type time)  (time-nanosecond time)  (time-second time)))
```

egallesio:
Thanks @jpellegrini. Merged.

egallesio:
Not really have the time to see the cause, but tests don't work for SRFI-19:

```
est srfi 19 date conversions 3 on (equal? d (time-monotonic->date m)) expected #t but got #f
test srfi 19 date conversions 6 on (equal? (time-tai->date (time-monotonic->time-tai m)) d) expected #t but got #f
test srfi 19 date conversions 9 on (equal? d (time-utc->date m)) expected #t but got #f
test srfi 19 date conversions 12 on (equal? (time-tai->date (time-utc->time-tai m)) d) expected #t but got #f
test srfi 19 date conversions 15 on (equal? d (time-tai->date m)) expected #t but got #f
test srfi 19 date conversions 18 on (equal? (time-utc->date (time-tai->time-utc m)) d) expected #t but got #f
```

jpellegrini:
> Not really have the time to see the cause, but tests don't work for SRFI-19:

Hm, the tests did pass here. I'm looking into it!

jpellegrini:
I see the automated build shows the errors in the tests. However, the tests pass on my machine, and:

```
stklos> (import (srfi 19))
stklos> (define d (make-date 05     ; nano
                    10     ; sec
                    15     ; min
                    20     ; hour
                    25     ; day
                    06     ; month
                    2000   ; year
                   -10800))
;; d
stklos> (define m (date->time-monotonic d))
;; m
stklos> (time-monotonic->date m)
#[date 2000-6-25 20:15:10]
stklos> d
#[date 2000-6-25 20:15:10]
stklos> (equal? d (time-monotonic->date m))
#t
```

I don't see why the automated build doesn't work... It gets different dates -- why?

jpellegrini:
@egallesio could you perhaps run the test suite on your machine? Maybe on a FreeBSD machine, or some other system?
I've ran it on two different PCs, and all tests pass...

jpellegrini:
Ok, since Github action  uses Ubuntu, I'll try installing it on a virtual machine and running it there. May take a while...

egallesio:
On my machine (Arch/Manjaro)

```scheme
stklos> (import (srfi 19))
stklos> (define d (make-date 05     ; nano
                    10     ; sec
                    15     ; min
                    20     ; hour
                    25     ; day
                    06     ; month
                    2000   ; year
                   -10800))
;; d
stklos> (define m (date->time-monotonic d))
;; m
stklos> (time-monotonic->date m)
#[date 2000-6-26 0:15:10]
stklos> d
#[date 2000-6-25 20:15:10]
stklos> (equal? d (time-monotonic->date m))
#f
```

The problem seems that the result of `time-monotonic->date` returns a date in the local timezone, as you can see below 

```scheme
stklos> (describe (time-monotonic->date m))
#[date 2000-6-26 0:15:10] is an instance of the structure type %date.
Slots are:
     nanosecond = 5
     second = 10
     minute = 15
     hour = 0
     day = 26
     month = 6
     year = 2000
     week-day = 1
     year-day = 178
     dst = 1
     tz = 3600
```

Of course if I define d with an offest of +3600 (my timezone) instead of  -10800, 
everything is OK.

So the test with `equal?` is not correct (even if the dates correspond to the same
instant, but in different timezones) 

We probably need a `date=?` function here (or that the `xxx->date` functions return a date expressed at GMT+0).

egallesio:
For now, I have commented the tests using `equal?` for this SRFI, to avoid failure in GIThub CI.

egallesio:
OK i have added tests for date comparisons  in the test file for SRFI-19

```scheme
(define (%date-test comparator)
  (lambda (d1 d2)
    (unless (date? d1) (error "bad date ~s" d1))
    (unless (date? d2) (error "bad date ~s" d2))
    (comparator (date->time-tai d1)
                (date->time-tai d2))))

(define date=?  (%date-test time=?))
(define date<?  (%date-test time<?))
(define date>?  (%date-test time>?))
(define date<=? (%date-test time<=?))
(define date>=? (%date-test time>=?))
```

Perhaps, these function could be exported by SRFI-19 as extensions. What do you think @jpellegrini ?

BTW, all your original tests work now. :smiley:

jpellegrini:
> Of course if I define d with an offest of +3600 (my timezone) instead of -10800,
> everything is OK.

Oops, sorry!  Really silly on my part. I should have tested with a timezone different from mine!

jpellegrini:
> Perhaps, these function could be exported by SRFI-19 as extensions. What do you think @jpellegrini ?

Yes! PR #454

PR #454, opened 2023-02-12
Add date comparison functions to SRFI-19

jpellegrini:
Now STklos offers, as an extension to SRFI-19, the procedures `date=?`, `date<?`, `date>?`, `date<=?` and `date>=?`. These will compare dates by first normalizing them to make the time zone offset irrelevant, so "2000 Nov 12 03:30:10 GMT-2" will be taken as equal to "2000 Nov 12 02:30:10 GMT-1".

@egallesio I'm not sure I got the documentation right...

### DIFF
--- a/doc/refman/srfi.adoc
+++ b/doc/refman/srfi.adoc
@@ -131,7 +131,7 @@ are given <<feature_identifiers>>.
 
 
 === Misc. Information
-//desactivate numbering 
+//desactivate numbering
 :sectnums!:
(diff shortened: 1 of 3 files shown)


jpellegrini:
(Are these 5 procedures not enough perhaps to propose as a new SRFI?)

egallesio:
Merged! Thanks.

> (Are these 5 procedures not enough perhaps to propose as a new SRFI?)

Not sure whether it's worth it.

PR #456, opened 2023-02-13
Expands and fixes VM documentation

jpellegrini:
There were some imprecise descriptions in the text. Also added some more explanation.


egallesio:
**Great**. Thanks  @jpellegrini  
Merged.

PR #458, opened 2023-02-15
`vm.c`: use `listify_top` in `adjust_arity`

jpellegrini:
This simplifies the code, and perhaps it's more interesting to do `vm->sp+=n` within `listify_top` than to call `pop()` (that is, do `vm->sp++`) several times. (?)

Also included some comments about the two functions.


egallesio:
Yep, this is simpler and more coherent now. Thanks @jpellegrini for this PR.

PR #459, opened 2023-02-15
Procedures optionally remember their sources (or something like that)

jpellegrini:
Hi @egallesio !
What do you think of this?
This patch does the same that was done to get docstrings and signatures in procedures.

```
stklos> (compiler:keep-source #t)
stklos> (define (f a b) (- a b))
;; f
stklos> (%procedure-source f)
(lambda (a b) (- a b))
```

For the future:

* Procedures do not remember their names when defined (not sure if it's worth implementing this).
* Code is not formatted. Maybe we can write a simple `stklos-source-format` procedure later.


jpellegrini:
I added some basic tests now.

One thing: I've made `%procedure-source` return `#f` for non-procedures, because that's what `%procedure-arity`, `%procedure-doc` and `%procedure-signature` do. Should they instead trigger an error?

jpellegrini:
Also included tests for `%procedure-arity`... And removed `(import STKLOS-COMPILER)` from the tests (they seem to run fine without it -- the parameter objects and procedures are visible in the STklos module anyway)

jpellegrini:
Now I see there's `%procedure-name`... This could be used to show the procedure name with the source!

egallesio:
Thanks for this PR @jpellegrini 

I have changed a bit the way things are represented to avoid the internal pointers in closure objects (since the source, as well as the signature are most of the time not used). So, source and signature are now stored in the procedure plist. 
Furthermore, the various compiler flags can be set now from the CLI (for stklos and stklos-compile commands).

PR #460, opened 2023-02-18
Add hyperbolic trigonometric functions

jpellegrini:
Specifically for `(atanh 1)` and `(atanh -1)` we trigger an error, as in Chez, Gambit and most Common Lisp implementations (as opposed to doing math with infinities).

Also included some (very few) tests.

A survey of which Schemes support these functions is available here: https://github.com/schemedoc/surveys/blob/master/surveys/hyperbolic-trigonometric.md

* I changed the placement of `#define transcendental(name)` to make things clearer (the hyperbolic trigs are defined after several others)

In general, `(asinh (sinh x))` and similar compositions should be equal to x, except for inexactness due to the internal floating point number approximation for real numbers.
```scheme
(sinh (asinh 0+1i)) => 0.0+1.0i
(cosh (acosh 0+1i)) => 8.65956056235493e-17+1.0i
(tanh (atanh 0+1i)) => 0.0+1.0i
```
These functions will always return an exact result for the following  arguments:
```scheme
(sinh 0.0)     => 0
(cosh 0.0)     => 1
(tanh 0.0)     => 0
(asinh 0.0)    => 0
(acosh 1.0)    => 0
(atanh 0.0)    => 0
``` 
The methods used are reasonably fast. Faster methods would make the code a bit too unclear, perhaps.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -52,6 +52,10 @@ static unsigned int log10_maxint;
 
 #define FINITE_REALP(n) isfinite(REAL_VAL(n))
 
+/* Complex i:
+   will be used as a constant when computing some functions. */
+static SCM complex_i;
+
 /* Forward declarations */
 static void integer_division(SCM x, SCM y, SCM *quotient, SCM* remainder);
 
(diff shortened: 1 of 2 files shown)


jpellegrini:
I think there's a lot of room for improvement here. Will get another version tomorrow.

jpellegrini:
Hi @egallesio !
I think this one is ready now. :)

egallesio:
Thanks for this R @jpellegrini. It is merged now/

PR #463, opened 2023-02-20
Add `void?`

jpellegrini:
For completeness. We have `null?`, `eof-object?`, `zero?`, and other procedures that are essentially convenience for
`(lambda (x) (eq? x some-special-object))`, so why not this one also?

This would be similar, I think, to the `unspecified?` procedure in MIT Scheme.


egallesio:
Good idea @jpellegrini. 
Merged.

PR #464, opened 2023-02-22
Remove unused file `src/parameter-test.c`

jpellegrini:
Hi @egallesio !
I suppose this is a test used in the past, and you forgot to remove? (Sorry if I'm wrong)
The file is not referenced in any of the `Makefile`s (or anywhere in the sources, at all).


egallesio:
Ouch you are right, this file should not be in the distribution. 
Thanks.

PR #466, opened 2023-03-01
Small fixes to documentation formatting

jpellegrini:
`|%f|` -> `#f` between backquotes (so we get highlighting), and some other trivial changes.


egallesio:
This PR weird: there is only one commit which modifies a Makefile, but there is nothing about text formatting. it seems that some commits were lost (maybe i'm wrong).

jpellegrini:
Indeed! There was a large diff changing small things in documentation. I don't get why it happened.
I'll check this.

jpellegrini:
Now it's fixed... (I force-pushed from my PC).

egallesio:
Perfect. Thanks for this PR. 
Merged

PR #468, opened 2023-03-01
Use GMP for bignum GCD (results in a really good speedup)

jpellegrini:
If at least one of the arguments is a bignum, call the GMP.

This brings a really nice speedup to gcd *and* to lcm (which uses gcd internally):

1. GCD bignum x bignum   4.960 -> 0.175
2. LCM bignum x bignum   4.771 -> 0.330
3. GCD bignum x fixnum   7.490 -> 2.106
4. LCM bignum x fixnum  11.648 -> 6.681

(Times in seconds)

1: calculate GCD of two bignums 100_000 times
2: calculate LCM of the same two bignums from (1), 100_000 times
3: calculate GCD of a bignum and a fixnum 10_000_000 times
4: calculate LCM of the same two numbers from (3), 10_000_000 times

Each benchmark was repeated 20 times.

**CAVEAT**: I did not test with different numbers, and this *may* make a difference (the GCD can be calculated quicker, depending on its arguments).

**EDIT**: this also fixes a problem I've seen:

```
(gcd 0.0 -5) => 5
```

With the current PR, STklos will return an inexact number, as expected.

**NOTE** While working on this, I also noticed a small glitch in `abs` (Issue #470) and proposed a fix (PR #471). So this PR will fail one test, but that is because STklos currently returns a negative result for `(abs fx_least)`. With that fix (#471) the new GCD will pass all tests.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2261,9 +2261,133 @@ DEFINE_PRIMITIVE("modulo", modulo, subr2, (SCM n1, SCM n2))
  * @end lisp
 doc>
  */
+
+static SCM gcd2_fixnums(SCM n1, SCM n2)
+{
+  /* If one of the arguments was inexact, we force the result
+     to be inexact also... */
+  if (zerop(n2)) {
+
+    return (isexactp(n1) && (!isexactp(n2)))
+      ? exact2inexact(n1)
+      : n1;
+
+  } else return gcd2_fixnums(n2, STk_modulo(n1, n2));
+}
+
+
+/* In the specific case we have bignums, GMP is absolutely faster than
+   doing it ourselves. gcd2_bignum computes the GCD of two numbers,
+   assuming that the first one is a bignum, while the second may be a
+   bignum or a fixnum (no type check is done here -- this is done in
+   the gcd2 function. */
+static SCM gcd2_bignum(SCM n1, SCM n2) {
+  int exact = 1;
+
+  SCM orig_n1 = n1;
+  SCM orig_n2 = n2;
+
+  mpz_t r;
+  mpz_init_set_si(r,0);
+
+
+  /****************************************************/
+  /* CONVERT REALS INTO INTEGERS (FIXNUMS OR BIGNUMS) */
+  /****************************************************/
+
+  if (REALP(n1)) {
+    exact = 0;
+    n1 = double2rational(REAL_VAL(n1));
+    if (! (INTP(n1) || BIGNUMP(n1) ))
+      STk_error("exact or inexact integer required, got ~s", orig_n1);
+  }
+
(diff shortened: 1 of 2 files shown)


jpellegrini:
In case you'd like to know what I used to do the timings, it's here.  Not really a great benchmark, but helps to see how much of a speedup we get.

```scheme
(define (check-fixnum x) (unless (fixnum? x) (error "bad fixnum ~s" x)))
(define (check-bignum x) (unless (bignum? x) (error "bad bignum ~s" x)))

;; GCD bignums
(let ((a 0)
      (bigX (* (expt 2 70)
               (expt 13 50)
               (expt 7 100)))
      (bigY (* (expt 2 20)
               (expt 11 60)
               (expt 7 50))))
  (check-bignum bigX)
  (check-bignum bigY)
  (time
   (repeat 100_000
     (set! a (gcd bigX bigY))))
  a)

;; LCM bignums
(let ((a 0)
      (bigX (* (expt 2 70)
               (expt 13 50)
               (expt 7 100)))
      (bigY (* (expt 2 20)
               (expt 11 60)
               (expt 7 50))))
  (check-bignum bigX)
  (check-bignum bigY)
  (time
   (repeat 100_000
     (set! a (lcm bigX bigY))))
  a)

;; GCD Bignum-Fixnum
(let ((a 0)
      (bigX (* (expt 2 70)
               (expt 13 50)
               (expt 7 100)))
      (Y (* (expt 2 10)
            (expt 11 5)
            (expt 13 5)
            (expt 7 3))))
  (check-bignum bigX)
  (check-fixnum Y)
  (time
   (repeat 10_000_000
     (set! a (gcd bigX Y))))
  a)

;; LCM Bignum-Fixnum
(let ((a 0)
      (bigX (* (expt 2 70)
               (expt 13 50)
               (expt 7 100)))
      (Y (* (expt 2 10)
            (expt 11 5)
            (expt 13 5)
            (expt 7 3))))
  (check-bignum bigX)
  (check-fixnum Y)
  (time
   (repeat 10_000_000
     (set! a (lcm bigX Y))))
  a)
```

jpellegrini:
However, I think I forgot one thing:

```
stklos> (gcd (+ (expt 7 60) 3) 4.0)
4
```

It should return an inexact, right? I'll fix this

jpellegrini:
This is ready @egallesio !
And I included some tests for GCD and LCM (I noticed that STklos didn't have any).

jpellegrini:
Obs: both PR #473 and this will speed up the GCD (the other PR does this implicitly by making the conversion to integer faster). The timings I got here are for the plain application of this PR only. I suppose that it will be faster if both PRs are applied. (Not sure)

PR #471, opened 2023-03-02
Fix `(abs fx-least)` being negative

### LINKED ISSUE #470: `(abs fx-least)` is negative
```
(abs fx-least) => -2305843009213693952
```

I's a corner case:

```
(abs (+ fx-least 1)) => 2305843009213693951
```

jpellegrini:
The problem is that `-INT_VAL(x)` is not a fixnum when `x` is `fx-least`, but the C code was assuming it was.

Fix #470


egallesio:
Oops. Thanks for this fix @jpellegrini.

PR #472, opened 2023-03-02
Make `double2rational` slightly more precise for integers

### LINKED ISSUE #469: STklos `inexact->exact` returns number slightly different from other Schemes (ALL Schemes lose precision here, but all of them -- except STklos -- agree on the result)
`(exact 73786976294838206473.0)` is usually  not `73786976294838206473` -- however, STklos seems to be the only one that returns a different result:

```
Chez        73786976294838206464                                                                                                                                                                                                   
Chibi       73786976294838206464                                                                                                                                                                                                   
Chicken     73786976294838206464                                                                                                                                                                                                   
Gambit      73786976294838206464                                                                                                                                                                                                   
Gauche      73786976294838206464                                                                                                                                                                                                   
MIT         73786976294838206464                                                                                                                                                                                                   
STklos      73786976294838218064     <== last 5 digits
Sagittarius 73786976294838206464                                                                                                                                                                                                   
Ypsilon     73786976294838206464

Python      73786976294838206464       (*)                                                                                                                                                                                     
```

(*) In Python, I did `int(73786976294838206473.0)`

Found this while working on PR #468 (when that number is passed to `gcd` as an inexact number, the GCD result is wrongly computed by all Schemes).

(This is done in `double2rational`, in `number.c`)

jpellegrini:
This patch makes STklos behave like other Schemes (and like Python for example).

Instead of calling double2integer(i), where i is the integer part identified by `modf` (which is a double itself), we call `scheme2number`. This makes STklos a bit more precise:

```
(exact 73786976294838206473.0)
 =>
73786976294838218064 (before this patch)
73786976294838206464 (after this patch)

(exact 11111111111111111111111.0)
 =>
11111111111111110406248 (before this patch)
11111111111111111344128 (after this patch)
```

Fixes #469

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -437,35 +437,6 @@ static SCM double2integer(double n)     /* small or big depending on n's size */
   return z;
 }
 
-static SCM double2rational(double d)
-{
-  double fraction, i;
-  SCM int_part, num, den, res;
-  int negative = 0;
-
-  if (d < 0.0) { negative = 1; d = -d; }
-  fraction = modf(d, &i);
-  int_part = double2integer(i);
-
-  if (!fraction) {
-    res = int_part;
-  } else {
-    num = MAKE_INT(0);
-    den = MAKE_INT(1);
-
-    while (fraction) {
-      num      = mul2(num, MAKE_INT(2));
-      den      = mul2(den, MAKE_INT(2));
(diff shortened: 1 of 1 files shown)


jpellegrini:
Wait, I think I have a better solution.

PR #473, opened 2023-03-03
Make `double2integer` more efficient and precise

### LINKED ISSUE #469: STklos `inexact->exact` returns number slightly different from other Schemes (ALL Schemes lose precision here, but all of them -- except STklos -- agree on the result)

jpellegrini:
By using `mpz_init_set_d` (which the mini-gmp also supports).

Fix #469 

In order for this to work, we also move the more fundamental functions to an earlier position in the code, before `double2integer`:
- `bignum2integer`
- `bignum2double`
- `scheme_bignum2double`
- `scheme_bignum2real`
- `bignum2scheme_bignum`
- `bignum2number`

A small experiment:

```
;;; BEFORE
;;;
(let ((a 0))
  (time (dotimes (i 10_000_000)
    (set! a (exact 1180591620717411303424.0)))) a)
Elapsed time: 34188.079 ms      <== 34 seconds
1180591620717411468064          <== loss of precision

;;; AFTER
;;;
(let ((a 0))
  (time (dotimes (i 10_000_000)
    (set! a (exact 1180591620717411303424.0)))) a)
Elapsed time: 2897.31 ms        <== less than 3 seconds!
1180591620717411303424          <== precisely the same number!
```

Of course, in the example above I used an integer that *is* representable as double, `1180591620717411303424`. I know This won't help with those that are not rerepresentable, as
```
(= (exact 123923192323291993.0) 123923192323291993)  => #f
```
But for those which are possible, this seems to bring more precision.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -401,6 +401,48 @@ static Inline SCM double2real(double x)
   return z;
 }
 
+static Inline SCM bignum2integer(mpz_t n)
+{
+  return MAKE_INT(mpz_get_si(n));
+}
+
+
+static Inline double bignum2double(mpz_t n)
+{
+  /* I do not use the function mpz_get_d since it gives an unspecified value
+   * when converting a number which is +inf or -inf
+   */
+ char *s = STk_must_malloc_atomic(mpz_sizeinbase(n, 10) + 2);
+ return atof(mpz_get_str(s, 10, n));
+}
+
+
+static Inline double scheme_bignum2double(SCM b)
+{
+  return bignum2double(BIGNUM_VAL(b));
+}
+
+
+static Inline SCM scheme_bignum2real(SCM bn)
+{
+  return double2real(scheme_bignum2double(bn));
+}
+
+static Inline SCM bignum2scheme_bignum(mpz_t n)
+{
+  SCM z;
+
+  NEWCELL(z, bignum);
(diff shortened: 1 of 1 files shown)


egallesio:
Oops. I love when deleting code makes things better :rofl:.
Thanks for this PR.

PR #474, opened 2023-03-03
Fix `(even? 1.5)`, `(even? "i am a number")` and others being `#t`

jpellegrini:
Hi @egallesio !

So... STklos was returning `#t` for the following:

```scheme
(even? 1/2)
(even? 0+1i)
(even? 0+1.0i)
(even? "i am a number")
```

With this commit, we have:

```scheme
(odd? 1.0)              => #t
(odd? 1.5)              => #f
(odd? 1/2)              => #f
(odd? 0+1i)             => #f
(odd? 0+1.0i)           => #f
(odd? "i am a number")  => error

(even? 1.0)             => #f
(even? 1.5)             => #f
(even? 1/2)             => #f
(even? 0+1i)            => #f
(even? 0+1.0i)          => #f
(even? "i am a number") => error
```

Most Schemes will trigger an error for `1.5`, `1/2` and complexes. The standards say nothing about this... If you want I can rework this to also result in error for anything that does not represent an integer.

I also included some tests...


egallesio:
Thanks for signaling it. I have finally fixed it a bit differently (it looped when calling `(even? +inf.0)`). 

> I also included some tests...

We have now more than 20k tests. Thanks @jpellegrini for your work on this point.

PR #475, opened 2023-03-04
Don't include declarations inside a switch label...

jpellegrini:
But a block an be included -- apply braces around the statements.

GCC allowed this, and only reported with `--pedantic`; and CLANG was stricter and did not allow it at all.

(I noticed when compiling on Android)

By the way, sorry -- this was my mistake when In made the PR with the hyperbolic trigonometric functions.


egallesio:
I'm surprised that gcc admits this. I'm even sure it had already pointed out an error in this case.  Didn't see it too.  
Merged.

PR #476, opened 2023-03-04
Don't compare long witn unsigned long

jpellegrini:
Sorry @egallesio  for even another small PR. This one fixes a thinko in the recent "fix abs" PR (#471)...
Without this, GCC (with `-Wextra` warns about a comparison between long and unsigned long.

Perhaps `-Wextra` should be default for compiling STklos? This would help catch these small issues in advance.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2090,7 +2090,7 @@ doc>
 DEFINE_PRIMITIVE("abs", abs, subr1, (SCM x))
 {
   switch (TYPEOF(x)) {
-    case tc_integer:  if (INT_VAL(x) == (unsigned long)INT_MIN_VAL)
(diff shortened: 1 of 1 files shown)


egallesio:
In fact, I have corrected this issue in last commit. 

> Perhaps -Wextra should be default for compiling STklos? This would help catch these small issues in advance.

I don't want to set it by default since it is dependent of GCC. Anyway, all my tests are done with an alias where CFLAGS is set to '"-O2 -Wall -Wextra"'. Except when I merge directly in the GitHub UI, I always do a complete compilation with these flags. From time to time, I also try to compile it with CLANG, since the warnings are not always the same. Anyway, under Emacs, I use `eglot`, which uses clang under the hoods, so it is a bit as if the files are precompiled with clang.

PR #478, opened 2023-03-05
Explain `UGLOBAL_{REF,SET}` and the checked global table in `vm.adoc`

jpellegrini:


egallesio:
This is perfectly explained (better than I could do). 
Thanks @jpellegrini.

PR #480, opened 2023-03-06
Report module or library name for unbounded variables

jpellegrini:
Hi @egallesio !
If this useful? :)

* In `env.c`, move the error functions to after the declaration of the module structure, so we can use `MODULE_NAME` in `error_unbound_variable`

* Change `error_unbound_variable` so it takes the module as argument  instead of the module name -- so it can decide wether to report  an unbound variable in "module x" or "library x"

* Use `error_unbound_variable` everywhere

* Change the VM in a similar way (`error_unbound_variable` will take a module as second argument and include its name in the error message).

```
stklos> a
**** Error:
%execute: variable `a' unbound in module `stklos'

stklos> (define-library empty)
;; empty
stklos> (in-module empty x)
**** Error:
symbol-value*: variable `x' unbound in library `empty'

stklos> (select-module STKLOS-COMPILER)
STKLOS-COMPILER> x
**** Error:
%execute: variable `x' unbound in module `STKLOS-COMPILER'

stklos> (symbol-value 'x (find-module 'STKLOS-COMPILER))
**** Error:
symbol-value: variable `x' unbound in module `STKLOS-COMPILER'
```

### DIFF
--- a/src/env.c
+++ b/src/env.c
@@ -29,30 +29,6 @@
 #include "vm.h"
 #include "thread-common.h"
 
-static void error_bad_module_name(SCM obj)
-{
-  STk_error("bad module name ~S", obj);
-}
-
-static void error_unbound_variable(SCM symbol, SCM modname)
-{
-  STk_error("variable ~S unbound in ~S", symbol, modname);
-}
-
-static void verify_symbol(SCM obj)
-{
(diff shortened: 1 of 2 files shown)


egallesio:
Thanks @jpellegrini,

Merged. I have modified it a bit to show library names as lists rather than symbols with a '/'.

PR #487, opened 2023-03-08
Use error specialized functions instead of `STk_error()`

jpellegrini:
This would make the messages more coherent, and would also make it easier to maintain those messages.

By the way, @egallesio -- what about having those in a specific file?

It would be nice to have a file `error-functions.c`   (with the commonly used error-reporting functions, like "bad list", "bad integer" and so on), and their prototypes in `error-functions.h`. If you say it's OK I can do that.
(Some very specific error messages would still use `STk_error` of course...)

### DIFF
--- a/src/list.c
+++ b/src/list.c
@@ -67,6 +67,10 @@ static void error_not_exact_positive(SCM x)
 {
   STk_error("index ~W is not an exact positive integer", x);
 }
+static void error_bad_comparison_function(SCM x)
+{
+  STk_error("bad comparison function ~S", x);
+}
 
 int STk_int_length(SCM l)
 {
@@ -346,7 +350,7 @@ SCM STk_append2(SCM l1, SCM l2)
   CDR(prev) = l2;
   return res;
 Error:
-  STk_error("argument ~S is not a list", l1);
+  error_bad_list(l1);
   return STk_void; /* never reached */
 }
 
@@ -413,13 +417,13 @@ DEFINE_PRIMITIVE("list-tail", list_tail, subr2, (SCM list, SCM k))
   x = STk_integer_value(k);
   if (x >= 0) {
     for (l=list; x > 0; x--) {
-      if (NULLP(l) || !CONSP(l)) STk_error("list ~S too short", list);
+      if (NULLP(l) || !CONSP(l)) error_too_short(list);
       l = CDR(l);
     }
     return l;
   }
 
-  STk_error("index ~S is not an exact positive integer", k);
+  error_not_exact_positive(k);
   return STk_void; /* never reached */
 }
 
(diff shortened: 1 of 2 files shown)


jpellegrini:
> It would be nice to have a file `error-functions.c`

If this is OK, there are two possibilities:

1. Don't change the checks (`if (!SYMBOLP(s)) error_bad_symbol(s);`, but only put the error messages in the separate file. This would not have a performance impact, since function calls would only be needed for reporting the errors, not detecting them:
```c
void error_bad_character(SCM c)
{
  STk_error("bad character ~S", c);
}
```

2. Instead, actually create *only* a header  a file `error-reporting.h` with `static inline` functions that also do the checking (as below), and then include this in the other `C` files. There would be a central place for managing the error messages, but the actual binary code for them would be duplicated (`error_bad_character` would be compiled into several places, for example).

```c
static inline
void check_character(SCM c)
{
  if (!CHARACTERP(c)) STk_error("bad character ~S", c);
}
```

What do you think?

egallesio:
Thanks @jpellegrini for this PR. I'll merge it. 


> It would be nice to have a file error-functions.c (with the commonly used error-reporting functions, like "bad list", "bad integer" and so on), and their prototypes in error-functions.h.

The problem with placing all the error functions in a unique file is that we miss the opportunity to have inlining of the function by the compiler. When you declare error functions in the file as static functions, they may be inlined by the compiler (even without using the `inline` keyword).

Of course, we can place all functions in a `.h` file as `static inline error...`, but I really hate this C++ism, which permits you to have code in `.h` files. I know that this style is considered as OK now in C, but I really dislike this style. Furthermore, we'll have probably a lot of error functions which are used in only one or two files, and I prefer to have them near the place where they are used.

Tere are 119 error functions for now, and the only duplicates are (very quickly tested):

- error_bad_character (also defined as error_bad_char)
- error_bad_cpointer  (2 definitions)
- error_bad_index (3 definitions) + 1 with a 2nd parameter
- error_bad_length (2 definitions)
- error_bad_list (6 definitions)
- error_bad_procedure (2 definitions) + 1 definition as error_bad_proc
- error_bad_sequence (2 definitions)
- error_bad_string (8 definitions)
- error_bad_symbol (2 definitions)
- error_bad_type (2 definitions)
- error_bad_vector (2 definition + 1 with 2 parameters)
- error_const_cell (2 definitions)
- error_division_by_0 which is also named error_divide_by_0
- error_unbound_variable 1 definition with 1 parameter and 1 with 2 parameters


So duplication, except bad_list, bad_string, bad_index, and perhaps maybe bad_vector are used more than 3 times. Perhaps, these functions could be exported by `list.c`, `str.c` and `vector.c`. We'll lose the implicit inlining, but it should be OK. 

Otherwise, you are right, using `static void check_xxx(SCM o)` would probably make code clearer.

PR #488, opened 2023-03-08
Change `%procedure-signature` to `procedure-formals`, describe closures and primitives better, and also describe bytevectors

jpellegrini:
In `help.stk` and `describe.stk`

Also describe closures and primitives separately in `describe.stk`.

And don't try to call `procedure-signature` on primitives...

Some examples:

```
stklos> (describe +)
 #[primitive +] is a primitive procedure.
stklos> (define (f a b) (+ a b))
;; f

stklos> (compiler:keep-source #t)

stklos> (define (g a b) (- a b))
;; g

stklos> (compiler:keep-formals #t)

stklos> (define (h a b) (/ a b))
;; h

stklos> (compiler:keep-source #f)

stklos> (define (i a b) (* a b))
;; i

stklos> (describe f)
 #[closure f] is a procedure.

stklos> (describe g)          ;; keep-source also keep formals:
 #[closure g (a b)] is a procedure.
 The source code is available -- call `procedure-source` on it

stklos> (describe h)
 #[closure h (a b)] is a procedure.
 The source code is available -- call `procedure-source` on it

stklos> (describe i)
 #[closure i (a b)] is a procedure.
```

### DIFF
--- a/lib/describe.stk
+++ b/lib/describe.stk
@@ -131,10 +131,15 @@ doc>
                             (format port "an empty vector")
                             (format port "a vector of length ~s"
                                     (vector-length x))))
-     ((procedure? x)    (if (%procedure-signature x)
-                            (format port "a procedure, whose arguments are ~s"
-                                    (%procedure-signature x))
-                            (format port "a procedure")))
+     ((closure? x)      (format port "a procedure")
+                        ;; We don't need to show the formals, because they're already
+                        ;; displayed in the object. For example, [closure g (a b)].
+                        ;; But we offer the user a way to read the source, if it is
+                        ;; available.
+                        (when (procedure-source x)
+                          (format port
+                                  ".~%The source code is available -- call `procedure-source` on it")))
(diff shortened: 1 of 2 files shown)


jpellegrini:
I think it would be nice to also describe uvectors, but the changes to `uvector.c` are not obvious (in prticular, the need to pass the uvector type as first argument). I'll leave that for another time.

egallesio:
Thanks for this PR, @jpellegrini.

> I think it would be nice to also describe uvectors, but the changes to uvector.c are not obvious (in prticular, the need to pass the uvector type as first argument). I'll leave that for another time.

I have added the classes `<uvector>` and `<bytevector>` (a subclass and `<uvector>`), and the function `%uvector-type` which returns a symbol (`s8`, `u8`, ...). It should suffice for the description of uvectors.

jpellegrini:
Oh, we just got `%uvector-type`! :)
Thank you! I'll add SRFI-4 vectors to `describe` later today.

Issue #493, opened 2023-03-10
Parallel build problem


ryandesign:
With the latest code in master there's an intermittent problem with parallel builds and a problem where the build system doesn't realize when the build has failed.

```
make clean && \
CC=cc CPPFLAGS=-isystem/opt/local/include LDFLAGS=-L/opt/local/lib ./configure --prefix=/opt/local && \
make -j9 && \
make test
```

Output: [j9.txt](https://github.com/egallesio/STklos/files/10944009/j9.txt)

```
make clean && \
CC=cc CPPFLAGS=-isystem/opt/local/include LDFLAGS=-L/opt/local/lib ./configure --prefix=/opt/local && \
make && \
make test
```

Output: [j1.txt](https://github.com/egallesio/STklos/files/10944005/j1.txt)

jpellegrini:
From `j9.txt`, it seems that the missing dependency is of the two components of SRFI 41 on SRFI 1.

ryandesign:
Thanks for taking a look so quickly! That doesn't appear to have fixed it however.

jpellegrini:
> That doesn't appear to have fixed it however.

But is the error you get the same as before?

jpellegrini:
Ooooh -also, I forgot to commit `Makefile.in`. Can you run `autoreconf`, then `./configure`, and see if that works?

ryandesign:
I did run `autoreconf` already.

Yes, the first error is still:
```
**** Error while executing command ("../../utils/stklos-compile.stk" "--no-time" "--output=derived.ostk" "derived.stk")
	 Where: in error
	Reason: module/library (scheme list) does not exist
```
So derived.stk doesn't have a dependency on "scheme list"? But it seems like it does:

https://github.com/egallesio/STklos/blob/8191f02c75cc58693d3564168604a8918236ff30/lib/streams/Makefile.am#L73

jpellegrini:
Ok, I think I found what I did wrong (see #494). Can you check again?

jpellegrini:
(Not sure if I got it right. I added an explicit dependency on `1.ostk` instead of `list.ostk`. We'll see if this works...)

jpellegrini:
Hm, the dependency actually works. In `j9.txt` we have this line:

```
../../utils/tmpcomp -o list.ostk list.stk
```

Long before the error.

What I suspect is happening:

There are two places where the `list.ostk` file can be compiled:
1. The `Makefile` in `lib/scheme` compiles it
[...]

PR #494, opened 2023-03-10
Declare dependency of SRFI 41 on SRFI 1

jpellegrini:
SRFI 41 `(in derived.stk)` imports `(scheme list)`
See #493


jpellegrini:
Ouch. It's `derived.ostk` that depends on SRFI-1, not `srfi-41.ostk`... Will fix.

ryandesign:
I think this can be closed in favor of #495.

jpellegrini:
> I think this can be closed in favor of #495.

Agreed. Even if #495 is not the best option, this one certainly isn't the solution

PR #495, opened 2023-03-10
Exit if compilation error occurs

ryandesign:
See #493


utils/tmpcomp
```diff
@@ -84,6 +84,13 @@ ${prefix}/src/stklos -c -q -b ${prefix}/src/boot.img \
      -f ${prefix}/utils/stklos-compile.stk -- \
      --no-time $Copt --output=$out $in
 
+status=$?
+if test $status -gt 0
+then
+    rm -f "$out"
+    exit $status
+fi
+
```
ryandesign, line 93: I wanted to write:

```suggestion
     --no-time $Copt --output=$out $in || exit $?
```

but that left behind a 0-byte file which was enough to make `make` think the target had been successfully built.

It would be better if `stklos` did not leave behind an empty (or in any other ways incorrect) file when compilation fails but I didn't know how to do that.

ryandesign:
I still think it would be better if this were done in `stklos` (or, I'm realizing now, in utils/stklos-compile.stk?) instead, but I did it in utils/tmpcomp for now because I know how to write shell scripts but I don't know scheme.

jpellegrini:
> I still think it would be better if this were done in `stklos` (or, I'm realizing now, in utils/stklos-compile.stk?)

@egallesio - in this case we'd add a guard in `compfile.stk`, and in case an error happens, we report and delete the file. Would that  work?

jpellegrini:
> but I don't know scheme

I can tell you it's really worth learning! :)

ryandesign:
> in this case we'd add a guard in `compfile.stk`, and in case an error happens, we report and delete the file. Would that work?

That would be one half of the fix. The other half would be to create the file under a temporary name and, upon successful compile, move the temporary file to the real filename.

jpellegrini:
> That would be one half of the fix

Can you see if #502 helps?

ryandesign:
> Can you see if #502 helps?

That works for me, thanks!

egallesio:
Hello @ryandesign and @jpellegrini ,
 
I think that this resolved now. I have modified `stklos-compile` function to use a temporary file and delete it when the compilation fails. The use of a temporary file should solve some problems I had with `make -j` (without specifying a number of processors). Note, that I had no problems when specifying a value like 12 or 16 on my 4 core processor (x2 threads). 
This should close (I hope) this PR and PR #502. Tell me if you have problems.
Thanks.

jpellegrini:
Hi @egallesio !
It seems to have fixed it for me...

egallesio:
No news from @ryandesign. I suppose that it"s OK. Feel free to reopen this issue if it is not the case.
Thanks for signaling the issue.

PR #497, opened 2023-03-10
Describe keywords and characters more nicely

jpellegrini:
@egallesio - does this make sense?


egallesio:
Great. Merging it. 
Thanks @jpellegrini.

PR #498, opened 2023-03-11
Identify all Darwin OS versions

ryandesign:
This fixes the configure summary so that it doesn't identify most Darwin versions by the nickname "DARWIN_OLD" which was strange to see on current OS versions. Previously the only Darwin versions not identified as "OLD" were 9, 10, and 11, even though they were released over 15, 13, and 11 years ago, respectively. Whereas before the summary on my system would show:

```
SUMMARY
*******
               System:  Darwin-21.4.0
              OS nick:  DARWIN_OLD
```

it now shows:

```
SUMMARY
*******
               System:  Darwin-21.4.0
              OS nick:  DARWIN_21
```

I realize this changes the preprocessor macros. Previously a macro called `DARWIN_OLD` would have been defined. But I don't see anywhere in the code that made use of that macro.

### DIFF
--- a/configure
+++ b/configure
@@ -6816,19 +6816,15 @@ BUILD_ARCH=`uname -m`
 OS_NAME=$OS
 case $OS in
    LINUX)
-      v=`echo $version | sed -e 's/\([0-9]*\)\.\([0-9\]*\).*/\1_\2/'`
-      OS_NAME_VERSION=LINUX_$v
+      v=`echo $version | sed -e 's/^\([0-9]*\)\.\([0-9\]*\).*/\1_\2/'`
       ;;
    DARWIN)
-      case $version in
-         9*)  OS_NAME_VERSION=DARWIN_9;;
-         *)   OS_NAME_VERSION=DARWIN_OLD;;
-      esac
+      v=`echo $version | sed -e 's/^\([0-9]*\).*/\1/'`
       ;;
   *)
       v=`echo $version | sed -e 's/[^0-9]/_/g'`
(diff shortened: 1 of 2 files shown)


ryandesign:
I'll just add that although there is no harm in defining a macro for the Darwin version (like `DARWIN_21`), I would urge you never to use it for anything. Perhaps *osname_osversion* macros are useful on other operating systems, but on macOS, there can be a difference between the OS version you are building *on* (which your macro is based on) and the minimum OS version you're building *for*. The latter is called the macOS deployment target, and if you ever find yourself wanting to write code conditionally depending on the macOS version, you'll almost certainly want to be doing it based on the deployment target, not based on the build machine's OS version, and Apple provides the `MAC_OS_X_VERSION_MIN_REQUIRED` macro that you can use to do that.

For example, let's say you wanted to use one block of code if compiling for Darwin 21 and later and a different block for earlier OS versions. Darwin 21 is macOS 12 so you could write:

```c
#include <AvailabilityMacros.h> // for MAC_OS_X_VERSION_MIN_REQUIRED
#if MAC_OS_X_VERSION_MIN_REQUIRED >= 120000
// code for macOS 12 and later (Darwin 21 and later)
#else
// code for macOS versions less than 12 (Darwin versions less than 21)
#endif
```

But in a project like STklos which isn't a macOS GUI application I wouldn't expect this need to arise at all.

egallesio:
Thanks for this PR and for your explanation about `MAC_OS_X_VERSION_MIN_REQUIRED`. 
Merged.

PR #500, opened 2023-03-11
Don't allow non-integer stack sizes

jpellegrini:
`%make-thread` does not check if the stack size is integer, so this:

```
(thread-start!
  (make-thread
    (lambda ()
      (print (stack-size (current-thread))))
      "name"
      'a-symbol))
```

will likely crash.


egallesio:
Thanks @jpellegrini for this PR. I have corrected it differently. The problem came from the size variable is declared as an int, instead of a long (STk_integer_value return LONG_MIN, if the parameter is not an int)

PR #502, opened 2023-03-11
Remove output file if compilation fails

jpellegrini:
If anything goes wrong during compilation, print the error or condition message, remove the output file and exit with a -1 status.

Test: create a file `wrong.stk`,

```scheme
(import nothing)
(define x eally-nothing)
```

And compile:

```
stklos-compile -o wrong wrong.stk
```

The error message and the backtrace will be shown, *and* the output file `wrong` will be deleted.

Should this be optional, so the user would be able to keep the partial output file, maybe?


jpellegrini:
@egallesio -- I put the `guard` after `open-output-file` because its main purpose is to delete the output file when something goes wrong. If anything bad happens *while* opening the file, then `open-output-file` will signal an error, and there won't be an output file to be deleted.

jpellegrini:
Perhaps the compilation should always stop after the error, but "deleting the file and exiting" should be optional, so `stklos-compile` would use it?

lib/compfile.stk
```diff
@@ -98,59 +108,72 @@ doc>
...
-
-(export compile-file)
\ No newline at end of file
+    ;; If anything goes wrong during compilation, print the error or condition
+    ;; message, remove the output file and exit with a -1 status.
+    (guard (cnd (else (begin
+                        (eprintf "*** Error while compiling: ~a~%" (condition-message cnd))
+                        (eprintf "Backtrace:~%")
+                        (%display-backtrace (condition-ref cnd 'backtrace) 10)
+                        (when (delete-and-exit-on-failure)
```
ryandesign, line 117: I get an error:
```
**** Error while executing command ("./utils/stklos-compile.stk" "--no-time" "--output=wrong" "wrong.stk")
	 Where: in compile-file
	Reason: bad function `#t'. Cannot be applied
```
I don't know scheme but based on [this post](https://stackoverflow.com/questions/61860245/the-object-f-or-t-is-not-applicable) saying this error can be caused by too many brackets, I tried:
```suggestion
                        (when delete-and-exit-on-failure
```
which seems to help.

jpellegrini:
Sorry, I see you have made a suggestion, but I had force-pushed already.

egallesio:
See comment on PR #495

egallesio:
BTW, in my commit I don't display a backtrace, but an error message in GNU format for a better integration with other Unix tools (and Emacs of course). We'll see if it's sufficient, but I could add the backtrace as you did after the message if this seems useful.

PR #506, opened 2023-03-12
Small enhancements

jpellegrini:
The guard is around `STk_init_utf8()`, but that function, in
`utf8.c`, already has the same guard:

```c
int STk_init_utf8(void)
{
 #ifdef STK_DEBUG
  ADD_PRIMITIVE(char_utf8_encoding);
  ADD_PRIMITIVE(dump_string);
 #endif
  return TRUE;
}
```

It seems cleaner to call all init functions in lib.c, and just
include the debug guards around the relevant code chunks, as
is done in `utf8.c`, shown above.

Also fixed a typo in `mutex-common.c`

-----------------

Ok - I also included a lot of small trivial changes after the commit described above...

### DIFF
--- a/doc/hacking/hacking.adoc
+++ b/doc/hacking/hacking.adoc
@@ -37,7 +37,7 @@ on the same project across various editors and IDEs).
 
 The subdirectories in the STklos source tree are:
 
-* `doc` – documentation, written mostly in Skribe
+* `doc` – documentation, written mostly in asciidoctor
 * `etc` – various sample files for specific needs
 * `examples` – examples (oh, who could tell?)
 * `ffi` – `libffi` (a local copy)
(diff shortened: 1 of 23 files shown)


ryandesign:
I wonder why this PR contains a commit I wrote earlier and which has already appeared on master (under a different hash)...

jpellegrini:
Ouch. Wrongly rebased, clearly. I'll fix it. Sorry!

jpellegrini:
There is on more commit I'm trying to push here, but GitHub seems to have some problem today. From the command line, it refuses the push (even with `-f`), and the web interface also seems a bit unstable.

egallesio:
Thanks @jpellegrini, for all these enhancements. PR merged.

PR #507, opened 2023-03-12
Optimize `dotimes`, `repeat`, `while` and `until`, and also do some loop unrolling in `repeat`

jpellegrini:
### 1. DOTIMES

`dotimes` creates a `LET` even when not necessary:

```scheme
(macro-expand '(dotimes (i 10) 'work))
  => (let ((G1 10))
       (do ((i 0 (+ i 1)))
           ((>= i G1)
            (void))
         'work))
```

The new variable `G1` is not necessary -- since 10 is constant, it could have been used in the body instead of `G1`, and the `LET` is superfluous. For nested `dotimes`, this can make a difference.

With this patch we check (at macro expansion time) if `count` is a number, and when it is, we generate a begin instead of a let.

For example:
```scheme
(macro-expand '(dotimes (i MAX) 'work))
    => (let ((G77 MAX)) (do ((i 0 (+ i 1))) ((>= i G77) (void)) 'work))

(macro-expand '(dotimes (i 10) 'work))
    => (begin (do ((i 0 (fx+ i 1))) ((fx>= i 10) (void)) 'work))
```

The second case will be faster for nested `DOTIME`s, because it avoids access to a local variable *and* entering-and-leaving a `LET`.

### 2 REPEAT

We do loop unrolling! For `repeat` this is very easy, since it does not use variables that can be changed by the user.
    
* The parameter `compiler:unroll-iterations` controls the number  of iterations to unroll
* We have an internal function `%multiply-list` that is used to make copies of the body: `(%multiply-list '(x y) 3) => (x y x y x y)`
* `repeat` will divide the repetition count by compiler:unroll-iterations, and produce `(%repeat ,quot ,(multiply-list body))` followed by the remainder times the body:

```scheme
    (macro-expand '(repeat 5 a b c))
    => (begin
         (%repeat 1
           a b c
           a b c
           a b c) ;; 5 / 3 is one, so one iteration of three copies of body
         a b c
         a b c) ;; The remainder of 5 / 3 is two: these are those two copies.
```

**Some timings**. I have ran the following:

```scheme
(compiler:unroll-iterations K)
(let ((a 1))
  (time
    (repeat 1_000_000_000 
      (set! a (+ a 1)))) 
  a)
```

For different `K`, these are the running times:
| K | average time  (ms) |
|-----|---|
| 100 | 9461.12 |
| 10 | 12407.993 |
| 3 | 15015.482 |
| 1 | 26796.797 |

(I also checked the original `repeat`, and the time is similar to unrolling with K=1 -- that is, the new method has no overhead (as expected).

So it looks like it's really worth it.

**Possible issue**: ~we use `list-copy` to make the copies of the body. This *could* maybe have an undesirable effect? Maybe one would expect `(eq? x '(...))` somewhere in the body to be true, and it won't anymore because we copied the literal list? Should we not use `list-copy`?~ Hm, I actually don't think this is an isue...

### 3. REPEAT *and* DOTIMES

In `(repeat N body)` and in `(dotimes (var N) body)`, when `N` is constant *and* a fixnum we can use `fx-` and `fx>`, even though the speed difference is not that great.

### 4. ADDITION (new commits included) -- `while` and `until`

Rewrite while and until using `tagbody`. After that, disassembling  `(while test -1)` we see that the generated code went from 31 instructions to 9 instructions!

Also,

```scheme
(let ((a 1))
  (time
    (while (< a 10_000_000)
      (set! a (fx+ 1 a))))
  a)
```
    
The timing for that code goes from 524.973 ms to 384.457 ms.

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -40,7 +40,7 @@
         hash-table-keys hash-table-values hash-table-fold
         hash-table-merge! hash-table-copy
         fluid-let time tagbody
-        dotimes repeat while until
+        dotimes %repeat repeat while until
         call/ec
         base64-encode-string base64-decode-string
         md5sum-file
@@ -1032,13 +1032,40 @@ doc>
 doc>
 |#
 (define-macro (dotimes bindings . body)
+  ;; In the lambda passed to the apply, below:
+  ;; If count is a number (not a variable), don't include a
+  ;; LET for it -- it won't change anyway!
+  ;;
+  ;; For example:
+  ;; (macro-expand '(dotimes (i MAX) 'work))
+  ;;    => (let ((G77 MAX)) (do ((i 0 (+ i 1))) ((>= i G77) (void)) 'work))
+  ;;
+  ;; (macro-expand '(dotimes2 (i 10) 'work))
+  ;;    => (begin (do ((i 0 (fx+ i 1))) ((fx>= i 10) (void)) 'work))
+  ;;
+  ;; The second case will be faster for nested DOTIMEs, because it
+  ;; avoids access to a local variable *and* entering-and-leaving
+  ;; a LET.
+
+  ;; Also - if count is constant *and* a fixnum, we can use
+  ;; fx- and fx>, even though the speed difference is not
+  ;; that great.
+  ;;
+  ;; When count is not a fixnum, we use + and >= :
+  ;; (macro-expand '(dotimes (i 1180591620717411303424) 'work))
+  ;;   => (begin (do ((i 0 (+ i 1))) ((>= i 1180591620717411303424) (void)) 'work))
+  ;;
   (apply (lambda (var count . result)
-           (let ((limit  (gensym))
-                 (result (if (null? result) (list '(void)) result)))
-             `(let ((,limit ,count))
-                (do ((,var 0 (+ ,var 1)))
-                    ((>= ,var ,limit) ,@result)
-                  ,@body))))
+           (let* ((result (if (null? result) (list '(void)) result))
+                  (limit (if (number? count) count (gensym)))
+                  (head  (if (number? count)
+                             '(begin)
+                             `(let ((,limit ,count)))))
+                  (plus (if (fixnum? count) 'fx+ '+))
+                  (ge   (if (fixnum? count) 'fx>= '>=)))
+             `(,@head (do ((,var 0 (,plus ,var 1)))
+                          ((,ge ,var ,limit) ,@result)
+                        ,@body))))
          bindings))
 
 #|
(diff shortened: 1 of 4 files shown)


jpellegrini:
This can be enhanced... Wait a bit

jpellegrini:
I think we can also optimize `dotimes`, with a little extra care. But I'll leave that for later.

egallesio:
Thanks a lot @jpellegrini for this PR.

This optimizes a lot. I have changed a bit your code to permit also the unrolling if the first parameter is not a constant.
 We produce two loops (that can be not produced if `remainder` or `quotient` is 0).
 
 
We have

```scheme
stklos> (compiler:unroll-iterations 4)
stklos> (pp (macro-expand  '(repeat 18 (void))))   ;; 4 "big" loops, 2 small
(begin
  (let ((G18 4))
    (tagbody
      #:top
      (when (fx> G18 0)
        (set! G18 (fx- G18 1))
        (void)
        (void)
        (void)
        (void)
        (-> #:top))))
  (let ((G19 2))
    (tagbody
      #:top
      (when (fx> G19 0)
        (set! G19 (fx- G19 1))
        (void)
        (-> #:top)))))
stklos> (pp (macro-expand  '(repeat 12 (void))))  ;; 3 "big" loops only
(begin
  (let ((G21 3))
    (tagbody
      #:top
      (when (fx> G21 0)
        (set! G21 (fx- G21 1))
        (void)
        (void)
        (void)
        (void)
        (-> #:top))))
  (void))
stklos> (pp (macro-expand  '(repeat 3 (void))))     ;; 3 small loops only
(begin
  (void)
  (let ((G25 3))
    (tagbody
      #:top
      (when (fx> G25 0)
        (set! G25 (fx- G25 1))
        (void)
        (-> #:top)))))
stklos> (pp (macro-expand  '(repeat 0 (void))))    ;; special case. We should just produce a void (but should not happen often)
(begin (void) (void))
stklos> (pp (macro-expand  '(repeat N (void))))    ;; Not constant. compute quotient and remainder at runtime
(let* ((G29 N)
       (G30 (quotient G29 4))
       (G31 (remainder G29 4)))
  (tagbody
    #:top
    (when (> G30 0)
      (set! G30 (- G30 1))
      (void)
      (void)
      (void)
      (void)
      (-> #:top)))
  (tagbody
    #:top
    (when (> G31 0)
      (set! G31 (- G31 1))
      (void)
      (-> #:top))))
stklos>
```

Note, I have spent some time to use a `repeat` when expanding a `dotimes`. It was really great ... until I saw that it was not thread safe. In effect, if the code in the body creates a thread which uses the current value of the iteration variable, we need to create a copy of this variable (and we lost all the benefits of unrolling).

PR #509, opened 2023-03-13
Set some procedure names

jpellegrini:
Some procedure are defined inside closures using `set!`. Some of them did not have their names set, and this patch does that.

The procedures are:
`base64-encode-string`, `base64-decode-string`, `ansi-color-protect`, `require/provide`, `provide`, `provided?`, `try-load`, `equal?`.

Also, `ansi-color` had the name of `do-color`.

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -1150,7 +1150,10 @@ doc>
                            (get-output-string out))))))
 
   (set! base64-encode-string (encode/decode base64-encode))
-  (set! base64-decode-string (encode/decode base64-decode)))
+  (set! base64-decode-string (encode/decode base64-decode))
+
+  (%set-procedure-name! base64-encode-string 'base64-encode-string)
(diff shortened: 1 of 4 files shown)


egallesio:
Hi @jpellegrini,

Merged. 
Note that I have added a new way to set procedure name.                                        
                                                                                
A procedure can set its name with a keyword at the start of the body (like documentation strings).                                                     

```scheme                                                                                
    (define foo #f)                                                             
                                                                                
    (set! foo                                                       
                (lambda ()                                                      
                   #:foo     ;; Set the name of the lambda to foo               
                   "Documentation for foo"                                      
                   (+ a 1)))                                                  
                                                                                
    > foo                   =>  #[closure foo]                                  
    > (%procedure-doc foo)  => "Documentation for foo"                          
```                                                                                
Documentation and name can be in any order. They are evaluated at closure creation only.

All the calls to `%set-procedure-name` have been suppressed in current source. IMO, code is less cluttered with  keywords.

PR #511, opened 2023-03-16
Fix formatting problems

ryandesign:
The markdown does not render properly when the backticks are at the wrong indentation.

### DIFF
--- a/QUICK-INSTALL.md
+++ b/QUICK-INSTALL.md
@@ -21,8 +21,8 @@ four) step process:
 
     You can also choose the compiler and the compiler options to use for building **STklos** with the `CC` and `CFLAGS` parameters:
 
-     ```bash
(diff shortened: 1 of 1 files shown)


jpellegrini:
I wasn't aware that indentation was needed when using the backticks!  Thanks for telling us.

ryandesign:
I wasn't either! I normally start a triple-backtick block at the beginning of the line preceded by no whitespace. But I noticed the formatting problems at the bottom of the document* (list items 2 and later render in a pre-formatted block) and I played around with it in the web-based editor and it seems that since this triple-backtick block occurs within a list item, it must be preceded by the same amount of whitespace as the rest of the list item.

*Compare [old](https://github.com/egallesio/STklos/blob/d50196d75e2b5d47c781b9614711926af11629ea/QUICK-INSTALL.md#readme) with [new](https://github.com/egallesio/STklos/blob/57eb8ce2e67f59b7a6b5a4a873e4168904622d18/QUICK-INSTALL.md#readme).

Issue #515, opened 2023-03-16
Inaccurate stable version designation when not building in a git clone


ryandesign:
Based on this output you'd think I was running the stable 1.70 version of stklos:

```
% stklos
  \    STklos version 1.70  (Id: stable-1.70)
   \   Copyright (C) 1999-2023 Erick Gallesio <eg@stklos.net>
  / \  [Darwin-21.4.0-x86_64/pthreads/libedit/utf8]
 /   \ Type ',h' for help
stklos> ^D
% 
```

but I'm not. I'm running nearly the latest code from the git master. It's just that I didn't build this from within a git clone; I build it from a tarball.

From the main page of this repository, click the green Code button, then click Download ZIP to get an archive of the latest code on master. (You can construct URLs to download archives of any commit or tag.) Extracting this archive and building it, stklos will identify itself as stable even though it's some random development version.

If you want to distinguish stable and unstable versions in your UI, then there has to be some place in your source code that states whether the code is stable or unstable. You can't rely on external factors like the output of the `git` program.

jpellegrini:
Hi @ryandesign !
Actually, each build of STklos contains the relevant information. It's just not being shown in the banner:

```
$ stklos
  \    STklos version 1.70  (Id: 590c7270)
   \   Copyright (C) 1999-2023 Erick Gallesio <eg@stklos.net>
  / \  [Linux-6.1.0-5-amd64-x86_64/pthreads/readline/utf8]
 /   \ Type ',h' for help
stklos> (%stklos-git)
(#:branch "master" #:commit "590c7270" #:tag "stklos-1.70" #:modified ("doc/Makefile.in" "src/Makefile.in"))
```

In Scheme, it is the `%stklos-git` procedure; in C, it is the `STk_stklos_git` function...

@egallesio - what about changing it in the banner?

jpellegrini:
About `%stklos-git` and other procedures...
Do you have readline? If you do, you can  start stklos, type a letter, (or `%`) then hit tab... Readline will suggest completions.

ryandesign:
This is the line of the output I was objecting to:

> ```
> \    STklos version 1.70  (Id: stable-1.70)
> ```

jpellegrini:
> This is the line of the output I was objecting to:

Yes. We can easily change the banner. But releases are just tags from git, so we'd need a way to tell that we're building that specific commit (tagged 1.80 for example)  and use it in the banner.

jpellegrini:
[...]

PR #516, opened 2023-03-17
Use git information to show stable versions in banner

jpellegrini:
See issue #515 

* When compiling from a git checkout, a stable version will be detected and included in `git-info.h`
* In `runtime.stk` there is a new procedure, `stklos-stable?`, that will tell if this STklos binary was compiled from a stable release
* In `repl.stk`, we use this to make the banner more informative: After creating a "10.1" tag for the current commit and reconfiguring and compiling STklos, we have:

```
  \    STklos version 10.1
   \   Copyright (C) 1999-2023 Erick Gallesio <eg@stklos.net>
  / \  [Linux-6.1.0-5-amd64-x86_64/pthreads/readline/utf8]
 /   \ Type ',h' for help
```

If we delete the tag, reconfigure and compile again,

```
  \    STklos version post-1.70 (git commit fefabbde)
   \   Copyright (C) 1999-2023 Erick Gallesio <eg@stklos.net>
  / \  [Linux-6.1.0-5-amd64-x86_64/pthreads/readline/utf8]
 /   \ Type ',h' for help
```

### DIFF
--- a/lib/repl.stk
+++ b/lib/repl.stk
@@ -330,7 +330,10 @@ doc>
       (repl-theme 'classic))
     ;; Show banner
     (let* ((utf8?  (key-get *%system-state-plist* :use-utf8 #f))
-           (line1   (format "STklos version ~A  (Id: ~A)\n" (version) (%push-id)))
+           (version-string (if (stklos-stable?)
+                               (format "version ~a" (version))
+                               (format "version post-~a (git commit ~A)" (version) (%push-id))))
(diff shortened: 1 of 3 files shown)


jpellegrini:
The `stklos-stable?` procedure can be cleaned up  a bit - but the idea is there...

jpellegrini:
@egallesio this one is not necessary anymore - right?

egallesio:
> @egallesio this one is not necessary anymore - right?

You're right. Closing it.

PR #522, opened 2023-03-18
Remove sentence about STklos library from manpage

ryandesign:
See 86b6c11c where the same sentence was removed from the README.


egallesio:
Oops. Merged.

PR #523, opened 2023-03-19
Fix bytevector portability (failing tests on Android)

jpellegrini:
Some (six) tests were failing on Android because we were trusting that the sign of `char` would be the same everywhere...  So in Android, `bytevector-sint-ref` was not returning signed integers, as well as `bytevector-s8-ref`.

Changed several occurrences of `char` to `int8_t`. Passes all tests:

- x86_64 GCC
- x86_64 Clang
- Android Clang (which was failing!)

(@egallesio - sorry the buggy code was mine... I should have been more careful.)


egallesio:
Thanks @jpellgrini for the fix. Merging it.

PR #531, opened 2023-04-07
Make `expt` work with complex exponents, and adjust to one R7RS requirement

jpellegrini:
The following condition in the primitive definition for `expt`:
```c
  if (negativep(y))
    return div2(MAKE_INT(1),
                my_expt(x, sub2(MAKE_INT(0), y)));
```
was preventing it from working with complex exponents, because `negativep` will trigger an error for complexes. Adding `!COMPLEXP(y)` fixes it.

However, there is another issue that we also fix with this PR: R7RS says that

> "The value of $0^z$ is $1$ if `(zero? z)`, $0$ if `(real-part z)` is positive, and an error otherwise.  Similarly for $0.0^z$, with inexact results."

But the current code would not do the "zero when (real-part z) is positive", so we also fix this in `my_expt`.

Then, we add some tests. And adjust one test from SRFI 70:

```scheme
(test "expr.4" 1    (expt (sqrt 3) 0))
```

This won't work with R7RS, because `(sqrt 3)` is inexact, and then `expt` is *required* to return an inexact number. So we do this instead:

```scheme
(test "expr.4" 1.0    (expt (sqrt 3) 0)) ; because sqrt(3) is inexact
(test "expr.5" 1      (expt (sqrt 4) 0)) ; because sqrt(4) is exact
```

(And actually, as far as I can see, SRFI 70 doesn't really require that the result from `(expt (sqrt 3) 0)` be exact).

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3181,7 +3181,10 @@ static Inline SCM exact_exponent_expt(SCM x, SCM y)
 {
   mpz_t res;
 
-  if (zerop(y)) return MAKE_INT(1);
+  if (zerop(y))
+      return (isexactp(x) && isexactp(y))
+          ? MAKE_INT(1)
+          : double2real(1.0);
   if (zerop(x) || (x == MAKE_INT(1))) return x;
 
   if (TYPEOF(y) == tc_bignum)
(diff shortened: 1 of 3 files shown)


egallesio:
> @egallesio what do you prefer?

1 is probably better, since the value is exactly known.

jpellegrini:
Ok @egallesio - changed that!

egallesio:
Great fix. Thanks.
It is merged now.

PR #534, opened 2023-04-08
Optimize `double2rational` (and `inexact->exact`)

jpellegrini:
And consequently, `exact`.

When the full GMP is available, we just turn the double into an `mpq_t` variable and get back the numerator and denominator.

Much faster than the current code:

```
(let ((a 3.1)
      (b #f))
  (time
   (repeat 5_000_000
           (set! b (exact a))))
  b)

Current code: 14535.114 ms
This patch:    2461.519 ms  ==> Actually, 2105 ms, after using mpz_clear
```

This uses conditional compilation, since the mini-GMP does not have rationals.

There is another PR (#524) which also adds a primitive `%stklos-has-gmp?`

**edit**: I also added `mpz_clear` to several places in `number.c`.  It's of course required (we were leaking some memory), and also seems to make the code faster.

**edit, again**: I also noticed a speedup in `expt` with exact exponent.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -504,7 +504,7 @@ static SCM double2integer(double n)     /* small or big depending on n's size */
 static SCM double2rational(double d)
 {
   double fraction, i;
-  SCM int_part, num, den, res;
+  SCM int_part, res;
   int negative = 0;
 
   if (d < 0.0) { negative = 1; d = -d; }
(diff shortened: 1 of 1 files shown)


jpellegrini:
The results from `double2rational`, and from `exact`, are the same as before, and  the same as most Schemes. There is only one Scheme which does something different, which is Gauche, using a continued fraction expansion of the number:
```
3.1 == most Schemes exact ==> 6980579422424269/2251799813685248
3.1 == Gauche exact ==> 31/10
```
Maybe we can try to do that too, later. But I think this PR is already a nice enhancement, isn't it?

egallesio:
Merged this one too.The last one for today :smile: 

Thanks.

egallesio:
Not merged, in fact, since I have sync problems between machine. Will merge later.

PR #535, opened 2023-04-12
Clean up STk_errno_names

lassik:
- Sort in alphabetical order.

- EAGAIN and EDEADLK were listed twice.

- The following entry was incorrect:

    #ifdef ENOTSUP
        CODESET_ERRNO(EOPNOTSUPP)
    #endif

- Note: EOPNOTSUPP and ENOTSUP are sometimes the same number, sometimes different. We should list both names.


lassik:
Added another commit to fix a typo.

egallesio:
Oops. Thanks, @lassik, for this PR.
Merged.

PR #537, opened 2023-04-12
Better use of GMP in `expt` (fixes crash)

jpellegrini:
### 1. Include missing calls to `mpz_init`
 

In `exact_exponent_expt`, some calls were missing, and STklos would crash on this, for example:

```scheme
(expt (expt 2 100) -2)
```

This patch also includes a test for that.

If an exponent is reeealy too large:
```scheme
(expt fx-greatest 100000)
```

STklos will report a `SIGABRT` from the GMP.

Anyway, when the exponent is too large, several Scheme implementations either crash (Chez, Guile, Sagittarius), or report an error like "our of memory" or "exponent too large" (Gosh, MIT). Kawa and Bigloo return an imprecise infinity -- so the behavior of STklos should be OK.

### 2. Don't create and initialize a new `mpz_t` number unnecessarily 

* Use `mpz_ui_pow_ui` instead of creating and setting the value of a new mpz number to then call `mpz_pow_ui`.

  Since that function only accpets *unsigned* integers, we check the sign, pass the number as unsigned, then if the base was negative *and* the exponent was odd, we multiply by -1.

* Call `mpz_init(res)` for the `tc_bignum` case.

The first change seems to make exponentiation with fixnum base slightly faster:

```scheme
;; fixnum result:
(let ((a 0))
  (time
   (repeat 10000000
     (set! a (expt (expt 2 5) (expt 2 3)))))
  a)
;; 6545.442 => 5446.367
```

```scheme
;; bignum result:
(let ((a 0))
  (time
   (repeat 10000000
     (set! a (expt (expt 2 10) (expt 2 5)))))
  a)
;; 8859.299 => 7478.944
```

(Average of 20 runs)

The sign change didn't make much of a difference in timings (That is, timings seem similar for iterated computation of `(expt -32 9)` and `(expt 32 9)`.

### 3. Call `mpz_clear` before returning from `expt`

This actually makes it faster:

```scheme
(let ((a (* (expt 2 60)))
      (b 30)
      (c #f))
  (time
   (repeat 5_000_000
           (set! c (expt a b))))
  c)
    
4660.171 ms => 2900.074 ms
```

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3327,8 +3327,9 @@ static Inline SCM exact_exponent_expt(SCM x, SCM y)
 
       /* Put back the sign, if needed (that is, if sign (of x) < 0 and the exponent is odd): */
       if (sign<0 && ((INT_VAL(y)) & 1UL)) mpz_neg(res,res);
-
-      return bignum2number(res);
+      SCM scm_res = bignum2number(res);
+      mpz_clear(res);
+      return scm_res;
     case tc_bignum:
       mpz_init(res);
       mpz_pow_ui(res, BIGNUM_VAL(x), INT_VAL(y));
--- a/src/number.c
+++ b/src/number.c
@@ -3306,6 +3306,7 @@ doc>
 static Inline SCM exact_exponent_expt(SCM x, SCM y)
 {
   mpz_t res;
+  SCM scm_res;
 
   /* y is already known to be exact; so if it is zero,
      return exact one. */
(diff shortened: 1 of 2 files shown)


jpellegrini:
Force-pushed to fix conflicts due to merging of other PRs. Still passes all tests! :)

egallesio:
Great, great, great. 

PR #538, opened 2023-04-13
Fix help docs path

jpellegrini:
Don't use `major.minor.PATCH` for looking up the database; it's just `major.minor` (`short-path`).

Otherwise, we get this:
```
stklos> (help memq)
Warning: cannot open "/usr/local/share/stklos/1.70.1153/DOCDB".
```

### DIFF
--- a/lib/help.stk
+++ b/lib/help.stk
@@ -44,7 +44,7 @@
           '()))))
 
 (define *the-doc*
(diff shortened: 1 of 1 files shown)


egallesio:
Oops...
Thanks, @jpellegrini.

PR #539, opened 2023-04-23
Add `radians->degrees` and `degrees->radians`

jpellegrini:
Inspired by Gauche...
Two trivial functions, but actually useful. :grin: 

(@egallesio I may need to slow down a bit in the next weeks, but I already have a couple of other PRs ready, so I'll send them tomorrow...)


egallesio:
> Two trivial functions, but actually useful. :grin:

Yep. Merged.
Thanks  @jpellegrini

PR #540, opened 2023-04-23
Compute (log -0.0) as per R7RS

jpellegrini:
R7RS requires that

```scheme
(log  0.0) => -inf.0
(log -0.0) => -inf.0+(PI)i
```

STklos was returning `-inf.0` for both cases. This patch corrects that.


egallesio:
Oops. Didn't see this subtlety in R⁷RS. Thanks.

PR #543, opened 2023-04-23
Optimize (inline) the `c_r` aliases to compositions of `car` and `cdr`

jpellegrini:
## 1. Inline the `cxr` accessors

Currently STklos has all `cxr` aliases implemented as procedures:

```scheme
(disassemble-expr '(caddar x) #t)

000:  PREPARE-CALL
001:  GLOBAL-REF-PUSH      0
003:  GREF-INVOKE          1 1
006:

Constants:
0: x
1: caddar
```

This patch optimizes this, opening their code:

```scheme
(disassemble-expr '(caddar x) #t)

000:  GLOBAL-REF           0
002:  IN-CAR
003:  IN-CDR
004:  IN-CDR
005:  IN-CAR
006:

Constants:
0: x
```

Some timings:

```scheme
(let ((a #f) (L '(a b c d e f g))) (time (repeat 50_000_000 (set! a (cadddr L)))) a)
before patch: 4174.988 ms
after patch:  2786.303 ms
```

```scheme
(let ((a #f) (L '(a b c d e f g))) (time (repeat 50_000_000 (set! a (cadr L)))) a)
before patch: 3703.763 ms
after patch:  2479.088 ms
```

```scheme
(let ((a #f) (L '((((a))) b c d e f g))) (time (repeat 50_000_000 (set! a (caaaar L)))) a)
before patch: 4387.891
after patch:  2802.235
```

Running the STklos tests seems to become *very slightly* faster (around 1%, average of 20 repetitions) - probably the speedup isn't great here because the cxr functions are not used that often in libraries, and also because there are slow tests (such as the Unicode tests) that don't use them.

## 2. Optimize `list-ref` for small constant indices

Since the `cxr` accessors have been optimized, we can use them
to optimize some cases in list-ref. For example,

```scheme
(list-ref L 2) => (caddr L)
```

`list-ref` becomes indeed faster:

```scheme
(let ((a #f) (L '(a b c d e f g))) (time (repeat 50_000_000 (set! a (list-ref L 3)))) a)
before: 3828.612 ms
after:  2836.350 ms

(let ((a #f) (L '(a b c d e f g))) (time (repeat 50_000_000 (set! a (list-ref L 0)))) a)
before: 3562.374 ms
before: 2235.379 ms
```

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -87,7 +87,35 @@
                          (eq?            . ,eq?)
                          (eqv?           . ,eqv?)
                          (equal?         . ,equal?)
-                         (void           . ,void)))
+                         (void           . ,void)
+                         (caar           . ,caar)
+                         (cdar           . ,cdar)
+                         (cadr           . ,cadr)
+                         (cddr           . ,cddr)
+                         (caaar          . ,caaar)
+                         (cdaar          . ,cdaar)
+                         (cadar          . ,cadar)
+                         (cddar          . ,cddar)
+                         (caadr          . ,caadr)
+                         (cdadr          . ,cdadr)
+                         (caddr          . ,caddr)
+                         (cdddr          . ,cdddr)
+                         (caaaar         . ,caaaar)
+                         (cdaaar         . ,cdaaar)
+                         (cadaar         . ,cadaar)
+                         (cddaar         . ,cddaar)
+                         (caadar         . ,caadar)
+                         (cdadar         . ,cdadar)
+                         (caddar         . ,caddar)
+                         (cdddar         . ,cdddar)
+                         (caaadr         . ,caaadr)
+                         (cdaadr         . ,cdaadr)
+                         (cadadr         . ,cadadr)
+                         (cddadr         . ,cddadr)
+                         (caaddr         . ,caaddr)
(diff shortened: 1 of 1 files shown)


jpellegrini:
Yay! :)

egallesio:
Hi @jpellegrini, 

Thanks for this PR.  I have merged it and modified a bit the generated code. I have added an instruction in the VM for cxr functions. Performances are better (but not a lot, and very less than I expected). The good news is that it permits to have better error messages. For instance, 

```scheme
stklos> (cadadr '(1 (2 . 3)))
**** Error:
cadadr: wrong type of argument `3' for car in (cadadr '(1 (2 . 3)))
```

jpellegrini:
I see you have included a new instruction. I thought of doing it, but was afraid of adding too many instructions to the VM.

Anyway, the implementation of the CXR  functions looks great now! :)

egallesio:
>  but was afraid of adding too many instructions to the VM.

Yep. I try to avoid the inclusion of new instructions as far as possible. As I said, I really expected better performances with this new instruction. This was not the case, and I'm a bit disappointed. However, since it permits better error messages, I kept it

PR #544, opened 2023-04-23
Enhancements and optimizations for SRFI-1

jpellegrini:
Hi @egallesio !

Last one today, I promise :)

There are three commits in this PR:

## 1. ~Inline~ Make `%car+cdrs` and friends primitives

The reference implementation of SRFI-1 recommends that the following procedures be inlined:

```
%cars+, %cdrs, %cars+cdrs, %cars+cdrs+, %cars+cdrs/no-test
```

~We just did that.~ -- (actually, we made them primitives. We *could* inline them, but then we'd need to get them into the core, and make the compiler inline even more primitives...)

There is one single C function `cars_cdrs` that does all the work, and five new internal Scheme primitives.

All procedures in the `fold` family are faster now.

## 2. Other general enhancements

* In `redefine`, get the standard procedures from module `SCHEME`, not `STklos`.
* Don't define `proper-list?`, just make it an alias to `list?`, since they do the same thing, and `list?` is a fast primitive already in STklos core.

## 3.  Optimize `fold-right`

The single list case has been changed: to avoid exhausting the stack *AND* to obtain better performance, instead of a recursive procedure we call list->vector and process the vector from the end towards the beginning.

This avoids wrong results due to stack overflow in the following case:

```scheme
(define L (iota 1_000_000))
(fold - 0 L)
```
With the default STklos stack size it would overwrite local variables and signal an error. With this patch, it works (because the stack does not grow at all, and all work is done inside a single vector).

It also makes `fold-right` faster:

```scheme
(let ((L (iota 1_000))
      (a 0))
  (time
   (repeat  30000
     (set! a (fold-right - 0 L))))
  a)
previous version: 5527.42 ms
new version:      3996.49 ms
```

We only change the case of one single list -- this optimization would be less efficient for the N-ary cse, since we'd have to compute the minimum size of all lists (or vectors), potentially traversing the larger list (that could perhaps be a million times larger than the others). It would also probably require some more uses of apply.

### DIFF
--- a/lib/scheme/Makefile.am
+++ b/lib/scheme/Makefile.am
@@ -70,8 +70,7 @@ R7_OBJS =  base.ostk            \
 # ----------------------------------------------------------------------
 # R7RS Large Support
 # ----------------------------------------------------------------------
-LARGE_STK  = list.stk       \
-             charset.stk    \
+LARGE_STK  = charset.stk    \
              stream.stk     \
              box.stk        \
              set.stk        \
@@ -86,8 +85,7 @@ LARGE_STK  = list.stk       \
              bitwise.stk
 
 
-LARGE_OBJS = list.ostk       \
-             charset.ostk    \
+LARGE_OBJS = charset.ostk    \
              stream.ostk     \
              box.ostk        \
              set.ostk        \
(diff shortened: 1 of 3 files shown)


egallesio:
Merged this PR. It definitively makes things better (in particular the function which uses `call/ec`). 
Thnaks a lot @jpellegrini

PR #546, opened 2023-04-30
Fix crash in `(scheme bytevector)`, and adjust the procedures to the spec.

jpellegrini:
I've found two bugs in the `(scheme bytevector)` library.

## How the bug was found -- `test/error` was wrong!

I have noticed that the `test/error` macro in `tests/test.stk` doesn't actually do anything: it ignores errors, but doesn't complain when **no** error is produced!   So I fixed it (will include in a later PR), and... Found several problems.

This is one of them -- now the test suite complains that we didn't error when the wrong index (non-aligned) was used. 

Later I'll send a PR with a new test macro, and several fixes in SRFIs, in the core, and in tests.

## 1. Check arguments in `bytevector-ieee-*-set!`

Otherwise STklos crashes!

Also, convert all numbers to imprecise (except complexes) before storing them in the bytevector (only complexes can't be turned into reals -- for bignums, if they don't fit a double, we store an infinity; for rationals, if they're too small, they become zero.)

## 2. Check indices in some of the IEEE bytevector procedures

The spec (the only part of $R^7RS$ Large that was taken from $R^6RS$) says we need to access with proper alignment (multiples of 4 for single precision, and multiples of 8 for double precision). Fixed.

### DIFF
--- a/lib/scheme/bytevector.c
+++ b/lib/scheme/bytevector.c
@@ -1065,6 +1065,7 @@ DEFINE_PRIMITIVE("bytevector-ieee-single-native-ref", bytevector_ieee_single_nat
   check_bytevector(b);
   check_integer(i);
   unsigned long idx = INT_VAL(i);
+  if (idx%4) STk_error("index for IEEE single precision bytevector not multiple of 4: ~S", i);
 
   switch (sizeof(float)) {
   case 4: {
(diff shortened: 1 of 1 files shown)


egallesio:
> I have noticed that the `test/error` macro in `tests/test.stk` doesn't actually do anything

Funny. This is a way to not have problem :wink:

egallesio:
Merged. Thanks.

PR #547, opened 2023-04-30
Make `expt` work again for inexact exponent + negative base

jpellegrini:
```
stklos>  (expt -2.0 0.5)
-nan.0
stklos>  (expt -2.0 1/2)
8.65956056235493e-17+1.41421356237309i
```

Oops. This was my recent PR.

We should not call `pow` for negative base and inexact exponent.


egallesio:
Merged. Thanks @jpellegrini

PR #548, opened 2023-04-30
Fix `test/error`, and fix a (long) sequence of bugs that were uncovered...

jpellegrini:
@egallesio --
In `tests/test.stk`, the macro `test/error` was ignoring errors, as it should. However, it was also ignoring the code to be tested, completely!

So I wrote a new macro (using `define-macro`) that uses `with-handler` to capture errors. Then I ran the test suit, and... Well, there were a lot of new errors!

One of the problems I have already sent separately (PR #546 -- the bytevector problem), but then there were so many others that I thought I'd keep them in one single PR.

Besides the several simple error fixes, I have three comments, below.

I'll be away for some days, probably...

## 1. Shouldn't `a+0.0i` be considered complex?

STklos treats it as the real number `a`, but... What if that `0.0` imaginary part came from an inexact computation that underflowed? I see several Schemes do:
* `a+0i = a` (real)
* `a+0.0i = a+0.0i` (complex)
So I changed this. If it's not OK, I can revert it. (tell me)

**_EDIT_**: Ok, I found it in R7RS (section 6.2.6):

```scheme
(real? -2.5+0.0i) => #f
```

So it seems like my intuition was in accordance with the standard...

## 2. We cannot test syntax errors

`with-handle`, `guard`, etc will not catch syntax errors, or compile-time errors at all. So I commented out several of those.

## 3. Misterious bug

One test fails when I run the test suit, but works fine on the REPL. (It's the guard test adapted from Gauche).
Do we comment it out?


egallesio:
Hi @jpellegrini,

I'm working on this PR. Correcting all the problems gave you a lot of work. Thanks. 
BTW, I have added a `test/syntax-error` for the tests you had to comment.

egallesio:
I have finally finished it.

> Shouldn't a+0.0i be considered complex?

Yes!  I have taken as is your code.  

> We cannot test syntax errors.

Now, we can :wink: In fact I use `eval` to force a compilation and test if it yields an error. BTW, the initial version of `test/error` was almot OK. It was: 

```scheme
(define-syntax test/error
  (syntax-rules ()
    ((_ str code)
     (test str *test-failed* result))))
```
 
The last parameter should be code in place of result . Consquently, I have not used your fix.

>  Misterious bug

Not seen it. :smiley:

jpellegrini:
Great @egallesio !

PR #549, opened 2023-05-04
Redefining a symbol should clear its immutbale bit

jpellegrini:
`set!` is not allowed on a mutable binding; but a 'define' would actually redefine the binding, so we should clear the immutability bit.
In order to forbid `define`s, one should make the *module* immutable (not the symbol) -- which is already possible using `module-immutable!`. Right?


egallesio:
You are right! 
Thanks for the fixed. Merged

PR #550, opened 2023-05-05
More power to constant folding :)

jpellegrini:
Hi @egallesio !
Still kind of erratic (got no time for long hacking sessions), but I managed to do this...

* This patch does constant folding to all math expressions, even using transcendentals etc. 
* Creates a compiler option to turn this on and off (`compiler:open-math` -- not a good name... Is there a better one?)
* It also creates a more compact way to create the inline table
* And, well, it's not large at all :)

Do you think it's interesting?

`(exp (sqrt 5))` will end up generating:

```
000:  CONSTANT             0

Constants:
0: 9.35646901660115
```

Another example, with a nested expression:
```
stklos> (disassemble-expr '(cos (+ 3 (tan 2))) #t)
000:  CONSTANT             0

Constants:
0: 0.685897402427631
```

I did this because it's not unusual to write `(expt 2 10)` or other powers of two... But then, why only optimize powers of two? We can do that to all math functions! :)

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -73,26 +73,36 @@
       `(set! ,*table* (cons (cons (quote ,name) ,name) ,*table*))))
 
 (define-macro (add-list-to-inline-table *table* . names)
+  (when (null? names) (error 'add-list-to-inline-table
+                             "need at least two arguments, onle one given."))
   (cons 'begin
         (map (lambda (n) `(add-entry-to-inline-table ,*table* ,n))
              names)))
 ;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define *inline-table* '())
-(add-list-to-inline-table *inline-table*
-                          + - * /
+(define *inline-math-commutative* '())
+(add-list-to-inline-table *inline-math-commutative*
+                          + * gcd lcm)
+(define *inline-math-general* '())
+(add-list-to-inline-table *inline-math-general*
+                          - /
                           fx+ fx- fx* fxquotient
-                          = < <= > >=
-                          fx=? fx<? fx<=? fx>? fx>=?
-                          ;; The following line associates symbols with values
-                          ;; different from their own (they're aliases):
-                          (fx= fx=?) (fx< fx<?) (fx<= fx=?) (fx> fx>?) (fx>= fx>=?)
                           sin  cos  tan  asin  acos  atan
                           sinh cosh tanh asinh acosh atanh
                           exp log expt
                           sqrt ;; square => square is not available when compiling...
-                          gcd lcm
+                          )
+
+(define *inline-math* (append *inline-math-commutative*
+                              *inline-math-general*))
+
+(define *inline-general* '())
+(add-list-to-inline-table *inline-general*
+                          fx=? fx<? fx<=? fx>? fx>=?
+                          ;; The following line associates symbols with values
+                          ;; different from their own (they're aliases):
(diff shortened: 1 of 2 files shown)


jpellegrini:
(This will conflict with #543, since it reformats the inline table. Easy to fix.)

jpellegrini:
Maybe in the future we can also do more general code transformations? Like this:

```scheme
    (when (and (compiler:some-compiler-option)
               (eq? fct 'log)
               (pair? (car actuals))       ; (log (...) ...)
               (eq? (caar actuals) 'expt)) ; (log (expt ...) ...)

      ;; (log (expt b e) z) => (/ e (log b z))
      (let ((e (caddar actuals))       ; (log (expt b e) ...) => e
            (b (cadar actuals)))       ; (log (expt b e) ...) => b
        (case len
          ((1) (compile `(/ ,e (log ,b ,z)) env epair #f))
          ((2) (let ((z (cadr actuals)))        ; (log (expt b e) z) => z
                 (compile `(/ ,e (log ,b ,z)) env epair #f)))))
```
(because computing
$$\frac{e}{log _z b}$$
 is faster than 
$$log_z(b^e)$$ 
-- one `log` and one division is faster than one `log` and one `expt` -- at least for non-bignums)

jpellegrini:
This PR can get much better. I'm working on it.

jpellegrini:
I think it makes sense now.

There are some cases where this won't do the complete constant folding I was hoping to get done, but it already does a lot:

```scheme
stklos> (disassemble-expr '(+ (sin 2) (log (expt 2 3) 4)) #T)

000:  CONSTANT             0
002:

Constants:
0: 2.40929742682568
```

jpellegrini:
One thing that this PR does is to change uses of `memq` into `assq`. This is because it simplifies the code, and both are primitives, so there doesn't seem to be too much of a difference in speed (differently from `assoc`, which is not primitive...)

egallesio:
I have not yet looked at the details, but this seems really impressive!

jpellegrini:
It's not complex at all... Just implemented some rules (the two new procedures in the compiler): `maybe-fold-constant` and `fold-commutative-args`. I'll include the code comments for them here, reformatted for clarity:

`maybe-fold-constant` takes the current environment, a list of arguments and tries to math-fold them into constants.

Example:

```scheme
(maybe-fold-constant '( 2.0 (log (sin 2)) (f 5) )) => ( 2.0  -0.0950830360951606  (f 5) )
```

* The number 2.0 remains unchanged;
* The expression (log (sin 2)) is evaluated;
* The expression (f 5) remains unchanged.

```scheme
(maybe-fold-constant #f '(expt 2 (expt 2 3))) => 256
(maybe-fold-constant #f '(expt x (expt 2 3))) => (expt x 8)
(maybe-fold-constant #f '(expt 2 (expt 2 x))) => (expt 2 (expt 2 x))
```

`fold-commutative-args` will do constant folding in the arguments as much as possible, and return a new argument list. This of course should be used only for commutative operations.

```scheme
(fold-commutative-args + '(a 2 3 b -1)) => (4 a b)
(fold-commutative-args * '(a 2 3 b -1)) => (-6 a b)
```

egallesio:
Hi @jpellegrini,

I have finally managed to merge this PR. I have also simplified the inline tables, now that the SCHEME module is immutable.  
Sorry for the long delay. I hope to be more productive soon. 
Anyway, thanks a lot for this PR.

egallesio:
> And... This, along with your recent enhancements to the equivalence predicates, seems to have boosted STklos' speed considerably. The test suite last week took ~8.2 seconds to run, and not it runs in ~7.1 seconds! 😄

Oh, I had great expectations about that but didn't see such improvements (on an old i5). Happy, if it works for you

PR #551, opened 2023-05-12
Fix typo in `gcd2` which was allowing arbitrary types (not just integer numbers) as arguments. (One-liner)

jpellegrini:
Hi @egallesio !

```
stklos> (gcd 10 'a)
0
stklos> (gcd 'a 10)
**** Error:
gcd: exact or inexact integer required, got a
```

Oh, that is because:

```c
  if (STk_integerp(n1) == STk_false) error_not_an_integer(n1);
  if (STk_integerp(n1) == STk_false) error_not_an_integer(n2);
                   ^^
                   ^^
```

Oops.. Fixed!


egallesio:
Thanks @jpellegrini.

PR #553, opened 2023-05-13
some notes on the gtk requirement

bentaca:
As a mac user I was positively surprised that I was able to build the examples in the extensions/gtk folder. Maybe these notes make it bit easier for the next user.

### DIFF
--- a/extensions/gtklos/README.md
+++ b/extensions/gtklos/README.md
@@ -10,13 +10,25 @@ To compile GTklos:
 $ make all      # compile GTklos and build the demos
 ```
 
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @bentaca,

Glad to see that you can use GTK on macOS. Thanks a lot for this PR.

BTW, in the `README.md` file, you said that the `STKLOS_GTK_DIR` variable needs to be set before the `configure && make` phase. Normally, this should be necessary only before running the demos (I will not be able to verify by myself for a while, but I'm always unsure with macOS :wink:). The loading of GTK libraries is done dynamically and is hence not necessary at compile time.

PR #558, opened 2023-07-02
Add compiler documentation

jpellegrini:
In a separate section.

We should probably also get the dos for the compiler flags in the same section with this...


egallesio:
Hi @jpellegrini ,

Flags were already documented in Appendix A. As a consequence, I have added the text of this PR in front of  compiler flags documentation in this appendix. 
Thanks

PR #559, opened 2023-07-09
Add `,version`, `,date` and other commands to the REPL

jpellegrini:
Hi @egallesio !
What do you think of this one?

### DIFF
--- a/lib/boot.stk
+++ b/lib/boot.stk
@@ -248,6 +248,10 @@
 ;; ----------------------------------------------------------------------
 ;;      option analysis and REPL launching
 ;; ----------------------------------------------------------------------
+
+(define repl-last-values (make-parameter (make-vector 3)))
+(define repl-last-expr   (make-parameter (make-vector 3)))
+
 (let ((no-init (key-get *%system-state-plist* :no-init-file #f))
       (ld      (key-get *%system-state-plist* :load  #f))
       (s176    (key-get *%system-state-plist* :srfi-176 #f))
--- a/lib/repl.stk
+++ b/lib/repl.stk
@@ -115,6 +115,16 @@ doc>
 (define (pad-with-zero str)
   (regexp-replace " " str "0"))
 
+(define (repl-update e v)
+  (let ((vec-v (repl-last-values))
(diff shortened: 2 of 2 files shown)


jpellegrini:
Maybe some others would be interesting?

```
- ,describe ,d  Describe an object
- ,time         Run the next expression while evaluating the time to do so
- ,require ,r   Require a feature
- ,expand ,e    Macroexpand a form
```

Those are things I usually do and would be nice to have quick access to, but maybe they're redundant, since one can also type them as Scheme forms?

jpellegrini:
By the way... It seems that the shell  command `,! stklos` can be nested an arbitrarily large number of times. Quite nice! :)
And when comparing output with other Schemes, we can do things like

```
stklos> (+ 2 3)
5
stklos> ,! gosh
gosh$ (+ 2 3)
5
gosh$ ^D
stklos> 
```

:grin:

jpellegrini:
I added more commands, including `,+` and `,*` (inspired by CL's `+` and `*`).

I only thing the `time` command is not actually good, since it discards the result of the expression.

jpellegrini:
Hmm, @egallesio -- the REPL uses an output port and an error port:

```scheme
(define (repl :key (in  default-in)
                   (out default-out)
                   (err default-err))
  .
  .
  .
                           (for-each (lambda (x) (write-shared x out) (newline out))
                                     v))
```

But the REPL commands don't:

```scheme
(define (do-repl-command l)
  .
  .
  .
    ((pwd)              (printf "~S\n" (getcwd)))
```

Shouldn't the REPL commands also be accepting error and output ports as arguments?

jpellegrini:
* The `describe` command may not be interesting the way it was implemented:

```scheme
stklos> (define x 20)
;; x
stklos> ,d x
x is a symbol, with mutable binding.
```

Not sure what do do here... `eval` it, maybe? Would that be ok?

egallesio:
I have mostly applied all your propositions, but in a different manner as explained in PR #598. 
Thanks for all your ideas.

jpellegrini:
It looks like this PR is also not needed! :)

PR #560, opened 2023-07-09
Add `%vm-config` for VM introspection

jpellegrini:
Hi @egallesio !

This is one thing I am using to help debug the VM. I am reeeeeally not sure if it's worth including, but well... Here it is, if you think it'd be nice to  have in STklos.

```
stklos> (%vm-config)
(computed-goto)
```

It would (when #528 is fixed) also include in the list the symbols `debug-vm` and `stat-vm` when appropriate.


egallesio:
Hi @jpellegrini,
I'm sorry for being so long to work on your PRs. I'm just in the process of purging most of the items on my TODO list, targeting the long awaited 2.00 version (yep!). I plan to add them just after 2.00, otherwise we will never have a stable release.

jpellegrini:
Hi @egallesio !
Of course, it's just an idea I had, and it's not that important. The new release is more important!

egallesio:
This is not for this specific PR, but all the ones that are still open.

egallesio:
I have merged your PR and modified it a bit. Instead of returning a list, `%vm-config` returns now a property list. 

```scheme
stklos> (%vm-config)
(#:computed-goto #t #:debug-vm #f #:stat-vm #t)
```

Thanks.

PR #565, opened 2023-07-23
Don't include extra `+` when printing complexes

### LINKED ISSUE #563: When the imaginary part of a number is `+inf.0`, STklos prints an extra `+` sign -- for example, `+inf.0++inf.0i`
Hi @egallesio !

```
stklos>  (* 1+1i +inf.0)
+inf.0++inf.0i

stklos> (* 0+1i +inf.0)
-nan.0++inf.0i

stklos>  (* 2+1i +inf.0)
+inf.0++inf.0i
```
Oops... Looks like there's an extra `+` being printed there...

jpellegrini:
Hi @egallesio !

I believe this fixes #563 ...  Some tests included. :)


src/number.c
```diff
@@ -384,7 +384,9 @@ static void print_complex(SCM n, SCM port, int mode)
   SCM imag = COMPLEX_IMAG(n);
 
   STk_print(COMPLEX_REAL(n), port, mode);
-  if (positivep(imag) || (REALP(imag) && !signbit(REAL_VAL(imag))))
```
egallesio, line 387: I read and re-read this code and don't understand why it works also for  NaNs (but obviously it works).  There is no call to `isnan`. Why does it works?

src/number.c
```diff
@@ -759,16 +762,22 @@ static char *number2Cstr(SCM n, long base, char buffer[], size_t bufflen)
...
+           glue them together in cplx. tmp is just a temporary buffer for
+           the imaginary part. */
+        char *re, *im, *cplx, tmp[100];
         size_t len;
 
-        s1  = number2Cstr(COMPLEX_REAL(n), base, buffer, bufflen);
-        s2  = number2Cstr(COMPLEX_IMAG(n), base, tmp, sizeof(tmp));
-        len  = strlen(s1) + strlen(s2) + 3;
-        s3 = STk_must_malloc_atomic(len);
-        snprintf(s3, len, "%s%s%si", s1, ((*s2 == '-') ? "": "+"), s2);
```
egallesio, line 769: This is clearer with `re` and `im,` but we lose the symmetry with what is done for rational. Perhaps, use `left`, `right`, `res` or both. What do you think @jpellegrini?

src/number.c
jpellegrini: Hi @egallesio !
I suppose it's because NaNs and infinities do have a signbit and they satisfy `REALP`, so

* `+nan.0` passes the first line in the condition (sign bit is zero). But for the extra plus to be printed, the other two lines require:
- Either zero (it's not), OR:
- Positive AND not infinity -- but NaNs so not satisfy the STklos C predicate `positivep`. :)

* `-nan.0` fails the first line, because its sign bit is 1 (negated = 0). Then the `+` is not printed, and the `-nan.0` shows up.

I'll include a comment in the code, I think it's kind of obscure...

src/number.c
jpellegrini: Ah, yes. I thought it was kind of away from the style in the code. I'll redo that part.

jpellegrini: Is it OK now?  :)

src/number.c
jpellegrini: (The sign bit for NaNs and infinities is described in the manpage for `signbit`, and as far as I know it's standard)

egallesio:
Everything is perfect. Merged.

Thanks a lot @jpellegrini

PR #566, opened 2023-07-25
Avoid some crashes with `#p`

jpellegrini:
Hi @egallesio !

I'm not sure this patch makes sense, but... Well, it avoids some crashes, and it's very simple (but perhaps there is a better way to handle this?)

When `#p......` is used, STklos may crash if the number read does not represent a valid STklos `SCM` object. 

This patch fixes it by changing `read_address` in `read.c`. When the number that is read ends in:

```
00: this is a pointer. we can't guarantee it's good. :( 
01: this is an integer. whatever the other bits are, it's OK
10: this is a small object (currently characters only).
    we *check* if this is really a character, and signal an
    error it it's not
11: this is a small constant. we check if it's larger than the
    last known small constant, and also trigger an error when
    it's appropriate.
```

Before (in both cases STklos crashes):

```
stklos> #pa                    ;; 1010, ends in 10 but not char
Received a SIGSEGV signal.
```

```
stklos> #p1f       ;; tagged small constant, but too large

**** PANIC:
Bad small constant 31
ABORT.
```

After (now STklos won't crash):

```
stklos>  #pa
**** Error:
%read: bad small object address #pa
```

```
stklos> #p1f
**** Error:
%read: bad small constant address #p1f
```

Unfortunately, I could not yet find a way to validate an arbitrary number as a Boehm-GC-managed address, so this will fail:

```
stklos> #p12000   ;; ends in 00
Received a SIGSEGV signal.
```


jpellegrini:
Included some tests... :)

These are very interesting:

```scheme
(test "address-of and read back, string"
      "xyz"
      (read-from-string (format "#p~x" (address-of "xyz"))))

(test "address-of and read back, date"
      #t
      (date? (read-from-string (format "#p~x" (address-of (current-date))))))
```

jpellegrini:
* Should there be a `address-ref`, which does the opposite of `address-of`?

```scheme
(address-ref (address-of x)) => x
```

I'm not sure it's useful, and it's kind of dangerous... (Unless we find a way to tell wether a pointer is valid).

jpellegrini:
Hmmmmm. The garbage collector actually does what we need!

```c
/* Return a pointer to the base (lowest address) of an object given     */
/* a pointer to a location within the object.                           */
/* I.e., map an interior pointer to the corresponding base pointer.     */
/* Note that with debugging allocation, this returns a pointer to the   */
/* actual base of the object, i.e. the debug information, not to        */
/* the base of the user object.                                         */
/* Return 0 if displaced_pointer doesn't point to within a valid        */
/* object.                                                              */
/* Note that a deallocated object in the garbage collected heap         */
/* may be considered valid, even if it has been deallocated with        */
/* GC_free.                                                             */
GC_API void * GC_CALL GC_base(void * /* displaced_pointer */);
```

I'll update this PR later.

jpellegrini:
Ok, done! :)

With this PR (unless I'm mistaken) `#p` is completely safe!

jpellegrini:
Except for small constants `#p17` = `#b10111` and  `#x1b` = `#b11011`. These are the two small constants:

```c
#define STk_dot         ((SCM) MAKE_SCONST(5)) /* special pupose value see read.c */
#define STk_close_par   ((SCM) MAKE_SCONST(6)) /* special pupose value see read.c */
```
and for some reason, they crash:
```
stklos> #p17

**** PANIC:
Bad small constant 23
ABORT.
```
(Why?)

jpellegrini:
I've added `address-ref`, for completeness. It it's too silly, I can remove it from the PR.

jpellegrini:
> Except for small constants `#p17` = `#b10111` and `#x1b` = `#b11011`. These are the two small constants:

Ah. I found it -- `MAKE_SCONST(5)` and `MAKE_SCONST(6)` should not be allowed. Fixed in the last commit.

jpellegrini:
This is really nice -- see the garbage collector in action:

```scheme
stklos> (address-of (make-vector 1000))
140094260412416
stklos> (address-ref 140094260412416)
#(#void #void ... #void)
stklos> (gc)
stklos> (address-ref 140094260412416)
**** Error:
address-ref: bad object address 140094260412416
```

jpellegrini:
Great! :)

```scheme
(dotimes (i  (expt 2 18))  (with-handler (lambda (a) -1) (eval-from-string (format "#p~a" i ))))
```

This finishes without errors or crashes now! :partying_face:

PR #567, opened 2023-07-28
Read complexes with infs and nans

### LINKED ISSUE #564: The STklos reader does not understand `+inf.0+1i` and `-inf.0+1i`
Hi @egallesio !
Found a tiny issue:
```
+inf.0+1i
%execute: symbol `+inf.0+1i' unbound in module `stklos'

-inf.0+1i
%execute: symbol `-inf.0+1i' unbound in module `stklos'
```

jpellegrini:
`STk_Cstr2number` now calls `Cstr2simple_number` twice in order to read the two parts of a complex number...

`Cstr2simple_number` needs to modify some of its arguments, so we pass some pointers to it -- mostly to keep the basis and exactness consistent in the two parts of the number.

This should fix #564

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -1303,7 +1303,10 @@ static SCM read_rational(SCM num, char *str, long base, char exact_flag, char **
   return STk_false;             /* never reached */
 }
 
-SCM Cstr2simple_number(char **str2, char *str, char *exact, long *base)
+/* STk_Cstr2simple_number will read from str a non-complex number.
+   The function STk_Cstr2number, which reads complexes, uses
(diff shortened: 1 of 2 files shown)


jpellegrini:
Should I change `Cstr2simple_number` to `STk_Cstr2simple_number`, and perhaps declare it inline?

jpellegrini:
> Should I change `Cstr2simple_number` to `STk_Cstr2simple_number`, and perhaps declare it inline?

I just did... :grin:

egallesio:
That's really great!
I have merged it.

PR #569, opened 2023-07-29
`(log -inf.0)` = `+inf + i*pi`

jpellegrini:
Because the limit when $x \to -\infty$ is actually $+\infty + i\pi$.

Current behavior:

```scheme
(log -0.0)    => -inf.0+3.14159265358979i
(log -10e50)  => 117.431839742696+3.14159265358979i
(log -10e100) => 232.561094392399+3.14159265358979i
(log -10e400) => -nan.0            ;; oops, should be inf + pi.
```

With this patch:

```scheme
(log -10e400) => +inf.0+3.14159265358979i
(log -inf.0)  => +inf.0+3.14159265358979i
```

A test is included.

This patch also corrects the test for `(log -0.0)`, which was ignoring the outcome from two predicates.


egallesio:
Thanks.

PR #574, opened 2023-08-02
More detailed description of NaNs

jpellegrini:
Nothing really important... Just be more informative when describing NaNs.


egallesio:
Merged! Thanks

PR #575, opened 2023-08-03
Be more careful when testing NaNs (and fix a failing test on Android).

jpellegrini:
`(* 0+1i +inf.0)` can be `-nan.0+inf.0i` (on Linux/amd64 for example) and also `+nan.0+inf.0i` (on Android, for example). So we don't test the sign of the NaN.

### DIFF
--- a/tests/test-number.stk
+++ b/tests/test-number.stk
@@ -431,17 +431,26 @@
 (test-subsection "complex writer syntax")
 
 (test "complex printing"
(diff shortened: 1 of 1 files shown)


jpellegrini:
(This morning I realized that this test was failing on Android)

jpellegrini:
@egallesio I think this is not necessary anymore, since you have commited a much better fix! :)

By the way, all tests pass on Android now. :smiley:

egallesio:
Oops. Forgot to tell you that it is corrected now, but you have seen it :smiley:

PR #578, opened 2023-08-07
More documentation to the compiler and `vm.adoc`

jpellegrini:
Hi @egallesio !
These are mostly patches that add comments to `compiler.stk`, and one which also expands `vm.adoc` a little bit.


egallesio:
Merged! Thanks.

PR #580, opened 2023-08-07
Fixes hyperbolic trigs for lim(x -> infinity)

jpellegrini:
These patches fix some limits... I only noticed this when I saw a commit to Gambit fixing something similar (although in a very different way).


egallesio:
Merged! Thanks.

PR #581, opened 2023-08-07
Fix behavior of `vector-copy` in SRFI 43

jpellegrini:
The procedure should augment the vector as necessary, even if there's no fill argument...

```scheme
(define v (make-vector 5 -1))
v => #(-1 -1 -1 -1 -1)

(vector-copy v 0 8)
```

The above currently triggers an error, but SRFI 43 says it should augment the vector with an unspecified argument...

This patch fixes this:

```scheme
(vector-copy v 0 8) => #(-1 -1 -1 -1 -1 #void #void #void)
```

### DIFF
--- a/lib/srfi/43.stk
+++ b/lib/srfi/43.stk
@@ -22,7 +22,7 @@
    vector?
    vector-empty?
    vector=
-   
+
    ;;; Selectors
    vector-ref
    vector-length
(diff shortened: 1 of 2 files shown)


egallesio:
Thanks @jpellegrini. Merged.

PR #583, opened 2023-08-10
Add a parameter to control the peephole optimizer

jpellegrini:
Hi @egallesio !
I know you're busy working on the 2.0 release -- I think this one could be done later. But I thought I'd make the PR anyway...

This makes the peephole-optimization pass optional -- this could be interesting for debugging and profiling, perhaps. :)

```
stklos> (compiler:peephole-optimizer )
#t
stklos> (disassemble-expr '(+ a #f) #t )

000:  GLOBAL-REF-PUSH      0
002:  IM-FALSE            
003:  IN-ADD2             
004:

Constants:
0: a
stklos> (compiler:peephole-optimizer #f)
stklos> (disassemble-expr '(+ a #f) #t )

000:  GLOBAL-REF           0
002:  PUSH                
003:  IM-FALSE            
004:  IN-ADD2             
005:

Constants:
0: a
```


jpellegrini:
Sorry for closing and reopening. That was not intentional.

jpellegrini:
Ah, The peephole optimizer does change

```
000:  IM-TRUE             
001:  IN-NOT              
002:  JUMP-FALSE           3	;; ==> 007
```

into

```
000:  IM-TRUE             
001:  JUMP-TRUE            3	;; ==> 006
```

but then it's too late to change the `JUMP-TRUE`, because it's not looking at the `IM-TRUE` before it anymore...

egallesio:
Also merged this one (not detected by GitHub). 

I also added the new parameter to the `stklos` and `stklos-compile` commands as a new flag. Not sure if this is a good idea, as I don't see any reason not to use it all the time (and it makes the commands more complex). What do you think?    

Thanks for the PR.

PR #585, opened 2023-08-12
`(- 0.0)` should be `-0.0`

jpellegrini:
Hi @egallesio !

Because of the way subtraction is performed, the expression `(- 0.0)` turns out to evaluate as `0.0`, because

* `(- 0.0)` becomes `(- 0 0.0)` in `"-"`
* `(- 0 0.0` becomes `(- 0.0 0.0)` in `convert`
* the result is `0.0`

This patch fixes this, so `(- 0.0)` is now `-0.0` (as all Schemes do)

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2157,6 +2157,12 @@ doc>
  */
 SCM STk_sub2(SCM o1, SCM o2)
 {
+  /* Special case:
+     (- 0.0) is calculated as (- 0 0.0)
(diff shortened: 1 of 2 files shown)


egallesio:
Thanks @jpellegrini. Merged 

I've also corrected some other problems with NaN and -0.0 elsewhere. It seems that problems with NaNs, infinities and negative zeros will never be finished :cry:

PR #588, opened 2023-08-14
Make `abs` work for complexes

jpellegrini:
R7RS says it "computes the absolute value of its argument", but doesn't imply it should be real...
So we do compute the norm,
` (abs a+bi)` $`= \sqrt{a^2+b^2} `$
as usual.

A very small patch... :)

```
(abs -3+4i)             => 5
(abs -3.0-4i)           => 5.0
```

Here's a list of how other lisps deal with `(abs -3-4i)`:

| System      | result       | remark           |
|-------------|--------------|------------------|
| Bigloo      | no complexes |                  |
| Biwa        | no complexes |                  |
| Chez        | error        |                  |
| Chibi       | -3-4i        | leaves unchanged |
| Chicken     | error        |                  |
| Cyclone     | error          |  |
| Gambit      | error        |                  |
| Gauche      | 5.0          |                  |
| Guile       | error        |                  |
| Kawa        | 5.0          |                  |
| LIPS        | error        |                  |
| MIT         | error        |                  |
| Racket      | error        |                  |
| Sagittarius | 5            | keeps exactness  |
| STklos      | error        |                  |
| Unsyntax    | -3-4i        | leaves unchanged |
|             |              |                  |
| ABCL        | 5.0          |                  |
| CCL         | 5            | keeps exactness  |
| Clisp       | 5            | keeps exactness  |
| ECL         | 5.0          |                  |
| GCL         | 5.0          |                  |
| SBCL        | 5.0          |                  |

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2284,6 +2284,8 @@ DEFINE_PRIMITIVE("/", division, vsubr, (int argc, SCM *argv))
  * @lisp
  * (abs -7)                =>  7
  * (abs -inf.0)            => +inf.0
+ * (abs -3+4i)             => 5
+ * (abs -3.0-4i)           => 5.0
  * @end lisp
 doc>
  */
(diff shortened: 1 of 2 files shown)


lassik:
R7RS says the signature is `(abs x)`. For the number procedures, `x` means a real number. `z` ("zahl") means any number.

Nevertheless, R7RS doesn't forbid procedures that work with more data types than is standard.

jpellegrini:
> R7RS says the signature is `(abs x)`. For the number procedures, `x` means a real number. `z` ("zahl") means any number.

I haven't paid attention to the signature... Indeed!
Anyway, @egallesio, here's the PR. Any decision seems to be fine! :)

jpellegrini:
Hm, I think I hadn't had coffee when I proposed this PR.   If one wants to make `abs` work for complexes, it would be simpler to do something like
```scheme
(define abs magnitude)
```
No?

egallesio:
I have applied your patch (and simplified the writing of magnitude, BTW)

PR #591, opened 2023-08-18
`(exact 1+0.0i)` should be an exact integer

jpellegrini:
Currently, STklos does this:

```scheme
(integer? (exact 65+0.0i)) => #f
```

Internally, a complex number is created with an exact zero imaginary part... But since the zero is exact, it would be strange to have this number as complex: it doesn't satisfy the `integer?` predicate, while 0+0i does:

```scheme
(integer? 0+0i) => #t
```

An easy solution is to change `STk_inex2ex` to call `make_complex` instead of `Cmake_complex`.


egallesio:
Merged.

PR #593, opened 2023-08-21
Add convenience macros and procedures...

jpellegrini:
More macros and procedures for `bonus.stk`:

* `dolist` (the extremely convenient Common Lisp macro)
* `push!`, `pop!` (Gauche has them, and Common Lisp has `PUSH`)
* `1+`, `1-` (as in Common Lisp, several Schemes have these)
* `inc!`, `dec!` (from Chez and Gauche)

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -59,6 +59,9 @@
         void?
         receive case-lambda
         radians->degrees degrees->radians
+        1+ 1-
+        pop! push!
+        dolist
 
         %define-condition-type-accessors
         message-condition? condition-message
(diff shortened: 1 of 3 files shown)


lib/bonus.stk
```diff
@@ -1124,6 +1127,109 @@ doc>
...
+  (let ((x (gensym 'res-)))
+    `(let ((,x (car ,stack)))
+       (set! ,stack (cdr ,stack))
+       ,x)))
+
+
+#|
+<doc EXT dolist
+ * (dolist (x lst) body)
+ *
```
lassik, line 1204: Common Lisp has an optional result form which you may or may not want to have in STklos: `(dolist (x lst [result]) body)`. Same with `dotimes`.

lib/bonus.stk
```diff
@@ -1124,6 +1127,109 @@ doc>
...
+ * (define x 10)
+ * (inc! x)
+ * x               => 11
+ * (dec! x)
+ * x               => 10
+ * @end lisp
+doc>
+|#
+(define-macro (inc! x)
+  `(set! ,x (+ ,x 1)))
```
lassik, line 1168: This will cause trouble if you have generalized `set!` in the future because `x` is evaluted twice. Unfortunately, dealing with generalized `set!` efficiently is hard; see "setf expansions" in Common Lisp.

lib/bonus.stk
jpellegrini: Yes - but since STklos' `dotimes` (and Gauche's) do not return a result, I thought this would keep the system consistent.

lib/bonus.stk
jpellegrini: Oh, you're right! In fact, STklos *does* have generalized `set!`, and I forgot about that. I'll remove this part for now.

jpellegrini:
Ok, removed `inc!` and `dec!`. Thanks @lassik for spotting this.

lib/bonus.stk
lassik: `push!` and `pop!` evaluate the list twice, too.

jpellegrini: Yes - the expansion should be more careful. I'll work more on this later...

jpellegrini: For example, Gauche expects either a symbol or an s-expression as the "stack" argument, and expands it as this:
```scheme
(macroexpand '(push! (list 1 2 3) 10))
  =>
(let
 ((getter.0 f) 
  (#:G347 a) 
  (#:G348 b) 
  (#:G349 c))
 ((setter getter.0) #:G347 #:G348 #:G349
  (cons 10 (getter.0 #:G347 #:G348 #:G349))))
```
which is very similar to what Clisp, for example, does:
```lisp
(macroexpand '(push 10 (f a b c)))
  =>
(LET* ((#:G2953 10)) 
  (LET* ((#:TEMP-2952 A)
         (#:TEMP-2951 B)
         (#:TEMP-2950 C)) 
    (LET* ((#:G2954 (F #:TEMP-2952 #:TEMP-2951 #:TEMP-2950)))
      (LET* ((#:NEW-2949 (CONS #:G2953 #:G2954)))
        (FUNCALL #'(SETF F) 
                 #:NEW-2949
                 #:TEMP-2952 
                 #:TEMP-2951
                 #:TEMP-2950)))))
```
And this is ECL (almost the same as that of Clisp):
```lisp
(LET* ((#:G1 A) 
       (#:G2 B)
       (#:G3 C)
       (#:G4 (CONS 10 (F #:G1 #:G2 #:G3))))
  (DECLARE (:READ-ONLY #:G1 #:G2 #:G3))
  (FUNCALL #'(SETF F) #:G4 #:G1 #:G2 #:G3))
```

When the stack is a symbol, they all revert to the simple expansion

```lisp
(macroexpand '(push 10 x)) => (SETQ X (CONS 10 X))
```

lassik: Yes. See [CLHS section 5.1: Generalized Reference](http://clhs.lisp.se/Body/05_a.htm). That's where the temp variables come from.

lassik: In particular [5.1.1.2 Setf Expansions](http://clhs.lisp.se/Body/05_aab.htm).

egallesio:
Hi @jpellegrini,

I have finally merged this PR. I have modified your code since, as remarked by @lassik, we have generalized `set!`. This has to be taken into account for `inc!` and `dec!` but also for `push!`. For instance,

```scheme
stklos> (define v (vector  1 2 3))
;; v
stklos> (push! (vector-ref v 0) 10)
stklos> v
#((10 1 2 3))
```
Since the implementation of `push!` is similar to the implementation of `inc!` and `dec!`, I have reintroduced them in this PR merging. 

For `dolist`, I have permitted to have a result as we have in `dotimes`:

```scheme
(let ((sum 0))
  (dolist (x '(1 2 3 4 5) sum)
    (inc! sum (square x))))       => 55
```

lassik:
NB: In Common Lisp the argument order is `(push element list)`

lassik:
I don't understand the `push!` example. Is `#((10 1 2 3))` really the correct result?

egallesio:
> NB: In Common Lisp the argument order is `(push element list)`

Yes, but in Scheme, functions on structured objects takes the object as first arg (`list-ref`, `vector-ref`, `vector-set!` ...). 
This is also the order used by Gauche. 

> I don't understand the push! example. Is #((10 1 2 3)) really the correct result?

Nope!  
In this case, it is `#((10 . 1) 2 3)`. I wanted to avoid the dotted pair and have redefined. The example should have been 

```scheme
stklos>  (define v (vector  (list 1 2 3)))
;; v
stklos> (push! (vector-ref v 0) 10)
stklos> v
#((10 1 2 3))
stklos> 
```

lassik:
> > NB: In Common Lisp the argument order is `(push element list)`
> 
> Yes, but in Scheme, functions on structured objects takes the object as first arg (`list-ref`, `vector-ref`, `vector-set!` ...). This is also the order used by Gauche.

Fair enough.

> The example should have been
> 
> ```scheme
> stklos>  (define v (vector  (list 1 2 3)))
> ;; v
> stklos> (push! (vector-ref v 0) 10)
> stklos> v
> #((10 1 2 3))
> stklos> 
> ```

Thanks. That makes sense.

Issue #434, opened 2023-01-01
Refactoring the sharp # reader


lassik:
The `#` and `#!` reader in `read.c` has at least two bugs, which seem to be oversights due to the messy code.

Looking at the code, it could be clarified if we used:

* a dispatch table instead of a big switch
* lookup based on the whole word, instead of one letter at a time (e.g. `#u32` instead of `u`)

Here's a quick sketch of how it could look:

```C
static SCM read_fold_case(SCM port, struct read_context *ctx,
                          const char *word)
{
    if (word[0] == 'n') {
        PORT_FLAGS(port) &= ~PORT_CASE_SENSITIVE;
        ctx->case_significant = FALSE;
    } else {
        PORT_FLAGS(port) |= PORT_CASE_SENSITIVE;
        ctx->case_significant = TRUE;
    }
}

static SCM read_keyword_position(SCM port, struct read_context *ctx,
                                 const char *word)
{
    PORT_KW_COL_POS(port) =
        colon_position_value(&s[strlen("keyword-colon-position-")]);
}

static SCM read_keyword(SCM port, struct read_context *ctx,
                        const char *word)
{
    (void)port;
    (void)ctx;
    return STk_makekey(word);
}

static SCM read_numeric_vector(SCM port, struct read_context *ctx,
                               const char *word)
{
    int tag = STk_uniform_vector_tag(tok);
    if (tag >= 0) {
        int konst = ctx->constant;
        
        /* Ok that's seems correct read the list of values (this IS a constant) */
        ctx->constant = TRUE;
        v =  STk_list2uvector(tag, read_list(port, ')', ctx));
        ctx->constant = konst;
        BOXED_INFO(v) |= VECTOR_CONST;
        return v;
    } else {
        signal_error(port,
                     "bad uniform vector specification ~A",
                     STk_Cstring2string(tok));
        return STk_void;
    }
}

struct sharp {
    const char *word;
    SCM read_func(SCM, struct read_context *, const char *);
};

static const struct sharp sharp_readers[] = {
    {"#!fold-case", read_fold_case},
    {"#!key", read_keyword},
    {"#!keyword-colon-position-after", read_keyword_position},
    {"#!keyword-colon-position-before", read_keyword_position},
    {"#!keyword-colon-position-both", read_keyword_position},
    {"#!keyword-colon-position-none", read_keyword_position},
    {"#!no-fold-case", read_fold_case},
    {"#!optional", read_keyword},
    {"#!rest", read_keyword},
[...]

PR #596, opened 2023-08-28
Add SRFI 178 - Bitvector library

jpellegrini:
Hi @egallesio !

Well...  I thought I'd implement one more SRFI... :grin: 

This is SRFI 178 - Bitvector library.

* Written mostly in C, but with some parts in Scheme

## MISSING:

* The reader syntax, '#*', is not implemented, because the reader has no way to know if the SRFI has been loaded.

Regarding this, I see some possibilities:

* Adopt the `bitvector` type as native, and keep only `bitvector`, `bitvector-set!`, `bitvector-ref` in the core, and if the user wants, the rest may be brought with `(import (srfi 178))`.
* Use a global variable `loaded_srfi_178` that tells the reader wether it can read bitvectors. *However* -- the bitvector data structure would not be known to the reader, and the SRFI would at least have to change a function pointer in the reader.
* Don't implement the read syntax (but it's very convenient; `#*10011` instead of `(bitvector 1 0 0 1 1)`)...

## POSSIBLE OPTIMIZATIONS FOR THE FUTURE:

* Some list reversals may be unnecessary
* Some procedures go bit-by-bit, when they could iterate on bytes

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -110,6 +110,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-174: POSIX Timespecs
     - SRFI-175: ASCII character library
     - SRFI-176: Version flag
+    - SRFI-178: Bitvector library
     - SRFI-180: JSON
     - SRFI-185: Linear adjustable-length strings
     - SRFI-189: Maybe and Either: optional container types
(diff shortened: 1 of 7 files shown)


jpellegrini:
Also... The code to read from a string is already there (see `string->bitvector`).

egallesio:
Yep, code of the reader has been refactored for `#xxx` objects and can be extended in Scheme (see last comment in Issue #434). I hope it will be useful for future extensions.

Anyway, thanks a lot @jpellegrini for this great PR.

PR #598, opened 2023-08-30
Adds a browseable manual from the REPL

jpellegrini:
Hi @egallesio !

This is probably not good, since it includes some fluff in the boot file, and I have written it without thinking about portability or error handling at all... But it's an idea for the future (post-2.0). :)  Maybe we can enhance this.

Some Scheme implementations (Gauche and Gambit) will call an external pager for the user to read the manual. This patch brings something similar to STklos.

Added parameters:

* `(default-browser)` is a string with the full path of the browser to be used.

* `(available-browsers)` is a list of possible browsers (strings with executable names), that will be searched in the system `PATH` at boot time.

* `(manual-path)` is the full path to the STklos manual. At boot time, this is searched in
```
   (string-append
     (%library-prefix)
     "/share/doc/stklos/"
     (short-version)
     "/html/stklos-ref.html")
```

Two procedures are added in `bonus.stk`:

* `(open-in-browser url)`
Opens the URL given by the string `url` in the default browser, which is determined by the string contained in the parameter `default-browser`.

* `(manual)`
Opens the STklos manual in a browser.

### DIFF
--- a/doc/refman/stdproc.adoc
+++ b/doc/refman/stdproc.adoc
@@ -966,6 +966,8 @@ See SRFI document for more information.
 {{insertdoc 'apropos}}
 {{insertdoc 'help}}
 {{insertdoc 'describe}}
+{{insertdoc 'manual}}
+{{insertdoc 'open-in-browser}}
 
 {{insertdoc 'trace}}
 {{insertdoc 'untrace}}
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -48,6 +48,7 @@
         print printerr eprintf printf fprintf
         declare-new-error
         exec exec-list
+        open-in-browser manual
         apropos die
         decompose-file-name dirname basename file-separator
         make-path file-suffix file-prefix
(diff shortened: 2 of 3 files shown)


jpellegrini:
By the way, it doesn't compile. I can copy and paste the new code and it works, but  building fails (something related to the boot process, I suppose).

jpellegrini:
Hmm, @egallesio -- the REPL uses an output port and an error port:

```scheme
(define (repl :key (in  default-in)
                   (out default-out)
                   (err default-err))
  .
  .
  .
                           (for-each (lambda (x) (write-shared x out) (newline out))
                                     v))
```

But the REPL commands don't:

```scheme
(define (do-repl-command l)
  .
  .
  .
    ((pwd)              (printf "~S\n" (getcwd)))
```

Shouldn't the REPL commands also be accepting error and output ports as arguments?

egallesio:
Hi @jpellegrini,

I'm sorry to be so slow to integrate your PRs.
To be able to browse the manual is really a goof d idea. I'm working (too slowly) on it. I have added custom IDs to the documentation, so that you can possibly add a parameter/ Hence,  `,m cadr` permits to open the manual on  the `cadr` function.

jpellegrini:
> I'm sorry to be so slow to integrate your PRs.

No problem! :)   I'm also swamped currently...

> To be able to browse the manual is really a goof d idea. I'm working (too slowly) on it. I have added custom IDs to the documentation, so that you can possibly add a parameter/ Hence, `,m cadr` permits to open the manual on the `cadr` function.

Ah, that will be great! I'll wait for it to be included, and adjust this PR.

jpellegrini:
Hi @egallesio !

It looks like there will be a lot of great enhancements! This is awesome!
I have some work and personal problems currently, so I couldn't do much in STklos, but sure, I can take a look at SRFI 27 if needed. I'll wait for your commit.

egallesio:
>  I have some work and personal problems currently, so I couldn't do much in STklos, but sure, I can take a look at SRFI 27 if needed. I'll wait for your commit.

I hope that you can resolve these problems swiftly. 

Anyway, I'll try to do my best to resolve the issue by myself. In any case, we can live with old implementation for a while.

jpellegrini:
> SRFI-27

Maybe it's an extra thing I added to the SRFI (the shared array counter, which uses a mutex and libgc's finalizer). It's unnecessary complexity, and I think it can be removed. I'll do that.

jpellegrini:
> Maybe it's an extra thing I added to the SRFI (the shared array counter, which uses a mutex and libgc's finalizer). It's unnecessary complexity, and I think it can be removed. I'll do that.

Sorry, I wasn' thinking straight. The array counter is in SRFI-25, not 27.
Anyway, I'll take a look in SRFI-27 when you commit the new code.

jpellegrini:
Okay, it looks like we have a really nice help system now!

```scheme
stklos> (default-browser)
"xdg-open"
stklos> (default-browser "lynx")
stklos> (open-in-browser "stklos.net")             ;; Opens stklos.net
0
stklos> ,m display                           ;; opens manual entry for display
stklos> ,m class-of                          ;; opens manual entry for class-of
```

Seems perfect to me! I think this PR is not needed anymore. :)

PR #600, opened 2023-08-31
(tiny fix) set names of two procedures

jpellegrini:
`base64-decode-string` and `base64-encode-string` had no names.  We can't use the `(lambda (...) #:name ...)` method, since it's a single lambda defining both procedures, so we force the name with `%set-procedure-name!`.


egallesio:
As explained in PR #598, this is automatically done now when setting or defining a variable to an anonymous lambda. 
So I close this issue (to give me the impression that the PR list shrink :smile: )

PR #602, opened 2023-09-01
`syntax-rules`: accept ellipsis with improper lists -- enhances pattern-matching (brings easy path to incorporate at least two SRFIs)

jpellegrini:
`mbe:matches-pattern?` and `mbe:get-bindings` did not account for the presence of an ellipsis pattern in an improper list, as in `(a b ... c . d)`. It would call `length` on the list tail after the ellipsis, which would signal an error.

So, we:

* Create `split-improper-tail`, which splits the list into lst-CDR and everything but it.
  `(split-improper-tail '())`        => values `'(), '()`
  `(split-improper-tail '(1 2 3))`   => values `'(), '(1 2 3)`
  `(split-improper-tail '(1 2 . 3))` => values `'(3), '(1 2)`

* Change `mbe:matches-pattern?` and `mbe:get-bindings` to use `split-improper-tail`

* Since we were there, document those procedures a bit. :)

* And, of course, include some tests...

### DIFF
--- a/lib/mbe.stk
+++ b/lib/mbe.stk
@@ -131,7 +131,7 @@ doc>
     (error 'letrec-syntax
       "cannot be used here. You must load the file \"full-syntax\" to access it:"
            (cons 'letrec-syntax args)))
-  
+
 
 ;;;
 ;;; Scheme utilities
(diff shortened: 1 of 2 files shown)


jpellegrini:
> The reference implementations of SRFIs 227 (optional lambda) and 232 (currying) use ellipsis patterns in improper lists, 
Ha ha! I just realized you already implemented SRFI 227. Anyway, 232 is ready!

egallesio:
Great!!!
Thanks @jpellegrini. I'm merging it immediately.

PR #603, opened 2023-09-01
Add SRFI 232 - Flexible Curried Procedures

jpellegrini:
Hi @egallesio !
As I mentioned in PR #602, with a simple modification in `syntax-rules`, the reference implementation of SRFI-232 works!
Passes all tests!

*** This is based on my other branch, from #602, so it's easier to see the changes there first, I suppose.

It exports all its symbols, but currently that's the only thing we can do with macros. 
(In the future, when we make macros reference variables in their defining environment, not their expansion environment, then we could only export the right symbols)


egallesio:
Hi @jpellegrini,

Happy new year! 
I have merged this PR. 
Thanks.

jpellegrini:
Happy new year @egallesio ! :)
That's great! And I see you have also implemented SRFI-115.   Wonderful!

Issue #604, opened 2023-09-03
`let-syntax` always signals an error


jpellegrini:
Hi @egallesio !

This is probably in your TODO list, but if it isn't, I'm opening an issue: the current `let-syntax` implementation doesn't work because it calls `%find-macro-clause` with 4 parameters, and a 5th was added.

```scheme
stklos> (let-syntax ((f (syntax-rules () ((f) 10)))) (f))
**** Error:
find-clause: 5 arguments required in call to `#[closure find-clause]' (4 provided)
```
Here is the cause:

```scheme
(define-macro (let-syntax bindings . body)
  `(%let-syntax
    ,(map (lambda (x)
            (let* ((macro-name (car x))
                   (syn-rules  (cadr x))
                   (keywords   (cons macro-name (cadr syn-rules)))
                   (clauses    (cddr syn-rules)))
              `(,macro-name (lambda args
                              (%find-macro-clause ',macro-name
                                                  args
                                                  ',keywords
                                                  ',clauses)))))  ;; Here! It should pass the ellipsis symbol too!
          bindings)
    ,@body))
```

The fix would require some rewriting of that macro...

jpellegrini:
Fixed! :)

PR #605, opened 2023-09-03
Fix a small glitch in `let-syntax`

jpellegrini:
Hi @egallesio 

This is about issue #604 --

The current `let-syntax` implementation doesn't work because it calls `%find-macro-clause` with 4 parameters, and a 5th was added.

```
stklos> (let-syntax ((f (syntax-rules () ((f) 10)))) (f))
**** Error:
find-clause: 5 arguments required in call to `#[closure find-clause]' (4 provided)
```

This patch changes `let-syntax` so it takes into account custom ellipsis symbols.

### DIFF
--- a/lib/mbe.stk
+++ b/lib/mbe.stk
@@ -131,7 +131,7 @@ doc>
     (error 'letrec-syntax
       "cannot be used here. You must load the file \"full-syntax\" to access it:"
            (cons 'letrec-syntax args)))
-  
+
 
 ;;;
 ;;; Scheme utilities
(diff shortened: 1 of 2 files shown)


jpellegrini:
So this now works:
```scheme
stklos>  (let-syntax ((f (syntax-rules () ((f) 10)))) (f))
10
```

jpellegrini:
One remark: perhaps we could also have, in the future, `let-syntax` for unhygienic macros? Something like
```scheme
(let-syntax ((f (lambda (a b) 
                  ...)))
   (f 2 3))
```
Which would be similar to an internal `define-macro`... (For the future...)

PR #606, opened 2023-09-05
Include missing argument to `compiler-error`

jpellegrini:
A tiny fix...


jpellegrini:
Doesn't really change STklos in terms of functionality, but it brings up a more informative error message than the current one:
```
stklos> (%%in-scheme)
**** Error:
compiler-error: `#[closure compiler-error]' requires at least 3 argument (2 provided)
```
:)

egallesio:
Merged for 2.00 release. Thanks @jpellegrini.

PR #607, opened 2023-09-05
(tiny fix) Fix plural in error message

jpellegrini:
```
STKLOS-COMPILER> (compiler-error )
**** Error:
compiler-error: `#[closure compiler-error]' requires at least 3 argument (0 provided)
```

Oops! That should have been plural, "arguments". Turns out it was a thinko when comparing arity.


egallesio:
Merged for 2.00 release. Thanks @jpellegrini.

PR #608, opened 2023-09-06
Use `~W` for general objects, not `~S`

jpellegrini:
Hi @egallesio !

Error messages do not know what the object is, so they shouldn't assume it's not circular:

```scheme
(define c (list 1 2 3 4))
(set-cdr! (cdddr c) (cdr c))
(+ c)
  => loops
```

### DIFF
--- a/src/fixnum.c
+++ b/src/fixnum.c
@@ -39,12 +39,12 @@ static void error_fx_at_least_1(void)
 
 static void error_negative_fixnum(SCM obj)
 {
-  STk_error("expected non-negative fixnum, found ~S", obj);
+  STk_error("expected non-negative fixnum, found ~W", obj);
 }
 
 static void error_bad_fxlength(SCM obj)
 {
-    STk_error("bad fixnum length ~S (should be < ~S)", obj, INT_LENGTH);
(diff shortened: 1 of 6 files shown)


jpellegrini:
Maybe this also happens in other files. I just checked `number.c`...

jpellegrini:
I have added patches for other files (not all of them)...

But I wonder if it wouldn't make sense to have a single directive, `~S`, that would already deal properly with circular structures?

egallesio:
You are right @jpellegrini, 

I should have seen this. 

> But I wonder if it wouldn't make sense to have a single directive, `~S`, that would already deal properly with circular structures?

That would be probably a better solution. We could make `~S` and `~W` be equivalent in error messages. For errors, the format parsing is done in the specific function ` print_format` in file `src/error.c`.  This should be easier than passing on  all files (and when we have an error, the overhead implied by  `~W` doesn't matter).

egallesio:
Hi @jpellegrini,
 I have modified the formatting function for errors. Now, both `~S` and `~W` use `write*` to print the culprit in an error message. 
Thanks for signaling the problem.
Since, it should resolve the problem, I close this PR, but you can reopen it, if needed.

jpellegrini:
Great news! :smile:

PR #609, opened 2023-09-11
Make `log` return exact result when possible

jpellegrini:
If the arguments to log are exact *and* the result is an integer, return it as an exact int.

```scheme
(log 16   2) => 4
(log 16.0 2) => 4.0

(log 1/4 2  ) => -2
(log 1/4 2.0) => -2.0
```


jpellegrini:
Not sure if this is correct. Would we possibly give an exact result when it's not really supposed to be exact?
STklos uses the C `log` function... Does it ever return an integer wrongly?

* Clisp does return an exact `4` for `(log 81 3)`...

jpellegrini:
Nope... Here are two different numbers with the same log, using this patch:
```
(log 179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368 2)
1024
stklos> (log 179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858367 2)
1024
```
So this would only work if both the first argument and the result are less than the first non-representable integer. I'll work on this later.

jpellegrini:
We'd actually need to do an exponentiation to check the result, and I think that'd be slow. Closing this... Sorry about the noise!

PR #610, opened 2023-09-28
configure: report if `asciidoctor` was found

jpellegrini:
Hi @egallesio !
I'm not an expert in autotools, so maybe there's a more elegant way to do this. Anyway, it would be nice to tell the user in the configure summary if `asciidoctor` was not found...

It won't complain or say anything if ruby-rouge is not installed, but it's already better than not saying anything at all...

```
SUMMARY
*******
               System:  Linux-6.5.0-1-amd64
              OS nick:  LINUX_6_5
              OS type:  unix
       Install prefix:  /usr/local
           C compiler:  gcc
    Compilation flags:  -g -O2
               Loader:  ld
       Thread support:  pthreads
Control fx parameters:  yes
System libraries used:  libffi libpcre2 libgmp libgc
   Compiled libraries:
          asciidoctor:  no (documentation will NOT be rebuilt)
```

### DIFF
--- a/configure
+++ b/configure
@@ -5072,7 +5072,7 @@ fi
 
 if test "$ASCIIDOCTOR" = "asciidoctor"
 then
-    REPORT_ASCIIDOCTOR="asciidoctor"
+    REPORT_ASCIIDOCTOR="yes"
 else
     REPORT_ASCIIDOCTOR="no (documentation will NOT be rebuilt)"
 fi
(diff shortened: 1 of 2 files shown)


egallesio:
Hi @jpellegrini ,

Merged this one. I have changed the message (I prefer to talk about documentation, instead of asciidoctor, since this could be meaningless for some) . However, I'm not really happy with the current formulation. I f you have a better idea , feel free to reopen this PR. 
Thanks

PR #613, opened 2023-11-20
Enhancements to REPL commands

jpellegrini:
* Mention that `ls` accepts usual arguments
* Make `,o` use the `default-browser` procedure (otherwise, at least on my Linux system, it won't work, although I can't see exactly why)

### DIFF
--- a/lib/repl.stk
+++ b/lib/repl.stk
@@ -231,7 +231,7 @@ doc>
 
   ;; ==== ls
   (repl-add-command 'ls
-                    "List directory content"
+                    "List directory content (accepts usual arguments to ls such as -l etc)"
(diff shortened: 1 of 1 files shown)


egallesio:
Thanks @jpellegrini for this PR. I have adapted it a bit:

* shorten the message for `ls` to fit in the 80-character limit when using help
* added the `,b` which calls the default browser and keep the `,o` command as is (since they are not the same). For instance, using  `,o foo.stk` will launch `emacs` on file `foo.stk` (if `xdg-open` is configured), or `o .` will open the file browser in the current directory. On the other hand, `,b foo.stk` will probably launch `emacs`, but **after**  having opening your browser.

PR #615, opened 2023-11-21
Bring back newer SRFI-27

jpellegrini:
It looks like we were missing an `in-module` in the definition of `random-integer` and `random-real`.


egallesio:
> It looks like we were missing an `in-module` in the definition of `random-integer` and `random-real`.

I'm quite sure I have done that, and it didn't work :disappointed: 
Anyway, thanks @jpellegrini for your fast (but clean :smile:) fix.

PR #616, opened 2023-11-27
Compute exact `log` when base is a power of 2 (also faster)

jpellegrini:
Create a `my_log2` C function, that deals with logs of exact powers of 2. This will give us exact results when the number is exact and the base is a small power of two:

```scheme
(log (expt 8 21) 8) => 21
(log (expt 2 10) 2) => 10
(log 1/4 2)         => -2
(log -4 2)          => 2+4.53236014182719i
```

In the last example, we got an _exact_ real part.

Roughly,
* For fixnums, compare to the known possible powers of two (there are not too many of them anyway)
* For bignums, use `mpz_sizeinbase`
* For rationals, do $\log(x/y) = \log(x) - \log(y)$. This allows us to give exact answers to `(log 1/64 4)`.

But additionally,

* If the base itself is a small power of two, we try to get an exact result anyway: use it in `mpz_sizeinbase` for bignums and use it (the base) to check for known powers for fixnums
* If the number is negative, return a complex with an imaginary part equal to $\pi/\log(base)$

This seems faster than the previous code, in particular when the first argument is a fixnum (but there seems to also be some speedup for bignums). Some timings:

```scheme
(define a  #f)

(define y (expt 4 50))
(time (repeat 50000000 (set! a (log y 4))))
;; current code  7019.777 ms
;; new code      4276.111 ms

(define y (expt 16 100))
(time (repeat 50000000 (set! a (log y 16))))
;; current code 6961.37 ms
;; new code     4384.379 ms

(define y (expt 2 500))
(time (repeat 50000000 (set! a (log y 2))))
;; current code     6929.235 ms
;; new code         4639.744 ms

(define x (expt 4 20))
(time (repeat 50000000 (set! a (log x 4))))
;; current code 6385.002 ms
;; new code     1645.658 ms

(define x (expt 2 50))
(time (repeat 50000000 (set! a (log x 2))))
;; current code 6365.162 ms
;; new code     2401.009 ms
```

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3603,8 +3603,8 @@ SCM my_log2(SCM x, SCM b) {
         unsigned long pwr = base;
 
         long xx = INT_VAL(x);
-        if (power_of_2_p(xx) && xx < base)
-          return div2(MAKE_INT(1), my_log2(b,x));
+        if (power_of_2_p(xx) && base > 2)
+          return div2(my_log2(x,MAKE_INT(2)), my_log2(b,MAKE_INT(2)));
 
         int pos = (xx > 0);
 
--- a/src/number.c
+++ b/src/number.c
@@ -74,7 +74,8 @@ struct bignum_obj {
 
 /*==============================================================================*/
 
-#define MY_PI           3.1415926535897932384626433832795029L  /* pi */
+//#define MY_PI           3.1415926535897932384626433832795029L  /* pi */
+#define MY_PI      3.1415926535897932384626433832795028841971693993751058209749445923078164062862090L
 
 #define BIGNUM_FITS_INTEGER(_bn) (mpz_cmp_si((_bn), INT_MIN_VAL) >= 0 &&        \
                                   mpz_cmp_si((_bn), INT_MAX_VAL) <= 0)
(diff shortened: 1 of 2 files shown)


jpellegrini:
It would be possible to extend this to bases up to `10`, for example, but it would require some more computation at the beginning of `my_log2`. I'm not sure it's worth it. Gambit, for example, only gives exact answers when the base is a power of 2, and returns inexact answers for all other bases. All other schemes seem to not preserve exactness for `log`. So if we keep this PR as it is, we'd have something similar to Gambit (and faster than what STklos has today).

jpellegrini:
> Not an error, but it would be nice in the future to enhance this to get `1/2` there.

Just did it. :)

Now the only thing missing is when the result is an exact rational with numerator different from one:

```scheme
(log 8 4)  => 1.5
```

I'd like it to answer `3/2`, but that requires more work. Maybe in a future PR...

jpellegrini:
> I'd like it to answer `3/2`, but that requires more work. Maybe in a future PR...

Okay... So just because I said that, I immediately realized it's actually a two-line change. implemented, and now it's ready. :)

```scheme
    (log 4 8) => 2/3
    (log 8 4) => 3/2
```

egallesio:
Hi @jpellegrini,

That's really really great. I'll integrate, it asap. 
Thanks.

egallesio:
Finally merged this PR. 
Thanks a lot @jpellegrini.

PR #618, opened 2024-01-15
Respect LDFLAGS

hattya:
The Makefile has a rule for `.c.@SH_SUFFIX@`, which includes `@LDFLAGS@`.


egallesio:
You are right. This is simpler and better. 
Merged. Thanks for your help @hattya.

PR #619, opened 2024-01-19
Customize `class-of` and `describe` for bitvectors

jpellegrini:
```
stklos> (import (srfi 178))
stklos> #*11
stklos> (class-of #*11)

Program received signal SIGSEGV, Segmentation fault.
```

This is because `#*11` is of a type which is only defined in `srfi-178.c`, and `object.c` can't tell what it is.

We do as we did in SRFI 25, and customize `class-of` and `describe` for bitvectors.

### DIFF
--- a/lib/srfi/178.stk
+++ b/lib/srfi/178.stk
@@ -109,6 +109,18 @@
                       list->bitvector
                       reverse-list->bitvector)
 
+;; Customize "describe"
+(%user-type-proc-set! 'bitvector 'describe
+                      (lambda (x port)
+                        (format port "a bitvector of length ~A"
+                                (bitvector-length x))))
+;; Customize "class-of"
+(define-class <bitvector> (<top>) ())
+(export <bitvector>)
+
+
+(%user-type-proc-set! 'bitvector 'class-of <bitvector>)
+
 (define (bitvector . args)
   (list->bitvector args))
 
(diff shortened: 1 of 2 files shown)


jpellegrini:
But it's still strange that it segfaults without this patch.

```c
DEFINE_PRIMITIVE("class-of", class_of, subr1, (SCM obj))
{
...
  if (HAS_USER_TYPEP(obj)) {
    SCM tmp = STk_extended_class_of(obj);
    if (tmp) return tmp;
  }
...
  return UnknownClass;
}
```
So it should return `UnknownClass`.

Anyway, the patch seems to work well for bitvectors.

egallesio:
> But it's still strange that it segfaults without this patch.
> 
> ```c
> DEFINE_PRIMITIVE("class-of", class_of, subr1, (SCM obj))
> {
> ...
>   if (HAS_USER_TYPEP(obj)) {
>     SCM tmp = STk_extended_class_of(obj);
>     if (tmp) return tmp;
>   }
> ...
>   return UnknownClass;
> }
> ```
> 
> So it should return `UnknownClass`.
> 
> Anyway, the patch seems to work well for bitvectors.

This is my fault. I have extended `STk_extended_class_of`: we can now have a class or a function which returns a class. I forgot to test if the value is `NULL`. I correct this immediately and  I'll merge your PR. 
Thanks for seeing this @jpellegrini .

PR #621, opened 2024-03-27
Make `(acos 1)` return exact zero

jpellegrini:
Previously we had

```
stklos> (acos 1)
0.0
stklos> (asin 0)
0
```

Hm, we have anice precise result for `(asin 0)`, why not for
`(acos 1)`?   It's because for asin, we have:

```
    case tc_integer:  if (z == MAKE_INT(0)) return MAKE_INT(0);
                      return asin_real(INT_VAL(z));
```

and for acos, we were also testing for exact zero, and then we returned inexact pi/2:

```
    case tc_integer:  if (z == MAKE_INT(0)) return div2(double2real(MY_PI),
                                                        MAKE_INT(2));
```

But that's not necessary, because `acos_real` will do that anyway. And we were not testing for the nice case of `(acos 1)`. Now we do that.

One test was also adjusted.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3270,8 +3270,7 @@ static SCM acos_real(double d)
 static SCM my_acos(SCM z)
 {
   switch (TYPEOF(z)) {
-    case tc_integer:  if (z == MAKE_INT(0)) return div2(double2real(MY_PI),
-                                                        MAKE_INT(2));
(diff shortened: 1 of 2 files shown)


egallesio:
Hi @jpellegrini,

That's perfect. Thanks.
Merged.

PR #622, opened 2024-04-17
Fix `(atan 0.0 0)` => +pi/2

jpellegrini:
And also `(atan -0.0 0)` => $-\pi/2$

This is explicitly given as a corner case in R7RS (in the table in the section for trigonometric procedures).

We also make `(angle 0)` trigger an error, since this would have to be exactly the same as `(atan (/ 0 0))`, which can't be computed.

Interestingly, only Chez does trigger the same error (for "`(angle 0)`"), but it is conceptually the right thing to do.

Tests included.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3311,6 +3311,19 @@ static SCM my_atan2(SCM y, SCM x)
 {
   if (STk_realp(y) == STk_false) error_bad_number(y);
   if (STk_realp(x) == STk_false) error_bad_number(x);
+  if (x == MAKE_INT(0))
+    /* For (atan y 0), if y is y is inexact zero, the result is either
+       -pi/2 or +pi/2, according to R7RS (there is a table in the standard
+       with several specific cases, including these). */
+    if (REALP(y)) {
+      double yval = REAL_VAL(y);
+      if (yval ==  0.0)
+        return signbit(yval)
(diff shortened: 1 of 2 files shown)


egallesio:
Merged. Thanks.

I have seen that your return a NaN for `(atan 0 0)` whereas it is an error for Chez. Don't know what is better (I left your code).

jpellegrini:
> I have seen that your return a NaN for `(atan 0 0)` whereas it is an error for Chez. Don't know what is better (I left your code).

Hmm, indeed. This is what others do:

|System | (atan 0 0) |
|-----------|---------|
|Bigloo  |error |
|Chibi  | 0.0 |
|Chez    |error |
|Cyclone |PI |
|Gambit  |0 |
|Gauche | +nan.0 |
|Guile   |0.0 |
|Kawa   | 0.0 |
|Loko  |  error |
|MIT    | 0 |
|Sagittarius | 0 |
|Unsyntax | 0.0 |

behavior seems to vary a lot.

My opinion: exact argumetns should not result in ineactness (not even NaN), so it would make more sense to do what Bigloo, Chez and Loko do.  What do you think @egallesio ?

egallesio:
You are right @jpellegrini , giving an error is a better option. I have changed the code to yield an error in this case.

PR #623, opened 2024-04-18
Make `(log 1 x)` return exact zero

jpellegrini:
Hi @egallesio !
One more small change:  for `(log 1 x)`, if x is neither 0 nor 1, we could just return exact zero...


egallesio:
Thanks @jpellegrini for this PR (and PR #622 too).
I'll merge them as soon as I have access to my computer.

PR #625, opened 2024-05-02
Small enhancement to message

jpellegrini:
Now we also use plural for zero: "zero bytes" and "zero allocation calls"...

```
stklos> (time 1)
;    Elapsed time: 0.016 ms
;    Allocations: 0 bytes in 0 allocation calls (GC: 0)
1
stklos> (time (list 1 2 3 4))
;    Elapsed time: 0.007 ms
;    Allocations: 96 bytes in 4 allocation calls (GC: 0)
(1 2 3 4)
stklos> (time (list 1))
;    Elapsed time: 0.004 ms
;    Allocations: 24 bytes in 1 allocation call (GC: 0)
(1)
```

But we use singular for one, as in the last example (one "allocation call").


egallesio:
Oops, I'm confused. You are right.

PR #626, opened 2024-05-03
Enhancement to error message

jpellegrini:
Before, we had:

```
stklos> (inc! "abc")
**** Error:
map: malformed list `"abc"'
```

Now we have:

```
stklos> (inc! "abc")
**** Error:
error: bad place "abc": it is neither a symbol (variable name) nor a list (generalized variable)
```

Of course this is not enough to make everything clear all the times, but it's better...


egallesio:
Once again, GitHub didn't close this PR. I do it manually.

PR #628, opened 2024-05-05
Add `improper-list?`

jpellegrini:
Not sure if this reeeealy is interesting, but here it is: it can be useful to make code clearer -- instead of

```scheme
(and (pair? x) (not (list? x)))
```

we can write

```scheme
(improper-list? x)
```


egallesio:
Hi @jpellegrini, 
As you said, not really useful. But why not?  :smile: 
Merged.

PR #629, opened 2024-05-05
Fix `valid-let-bindings?`

jpellegrini:
We had this:
```
(let 1 2)
=> #[primitive %execute]
```
because `valid-let-bindings?` would return `#f` and not trigger an error when `bindings` was not a list.

This patch fixes that error, and passes all tests.

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -1454,6 +1454,19 @@ doc>
 ;;;;
 ;;;; LET / LET* / LETREC
 ;;;;
+
+;; valid-let-bindings? returns #t if:
+;;
+;; - its first argument must be a valid LET binding
+;; - when the second argument is #t, there may not be duplicate bindings
+;;
+;; When these conditions are not satisfied, the procedure will NOT return
+;; #f; it will signal an error instead.
+;;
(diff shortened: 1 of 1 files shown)


jpellegrini:
By the way, I'm also working on SRFI 71 (Extended LET-syntax for multiple values), and that's how I found this problem. 
71 is not a popular SRFI, and I know we already have `let-values`, but... It's formally a standard, and actually not that hard to implement. :)

jpellegrini:
By the way, I didn't include tests because the error raised cannot be caught... It would break the test procedure.
Is there a way to catch *any* possible error, including those during compilation?

jpellegrini:
Perhaps it would also make sense to change the procedure name to `check-let-bindings`, since after this patch it would not be a predicate anymore?

egallesio:
Thanks @jpellegrini, for this PR.

>  Is there a way to catch any possible error, including those during compilation?

Yes, use `test/compile-error`.

> Perhaps it would also make sense to change the procedure name to check-let-bindings, since after this patch it would not be a predicate anymore?

You are right. Done.

PR #632, opened 2024-05-07
Tests for ill-formed `LET`

jpellegrini:


egallesio:
Oops, I had also added some tests in my previous commit, but forgot to add the test file. Since, Your tests are better, I have used yours. Thanks for your help @jpellegrini.

PR #634, opened 2024-05-16
Add one more example

jpellegrini:
`dice.stk`  is a very simple example of scheme program that asks for    a number `N`, then runs two dice `N` times, and finally prints a    histogram of the distribution of outcomes.

Concepts illustrated:

* Loading and using SRFIs; specifically, this shows the usefulness of SRFi-133 AKA (scheme (vector))
* Using STklos extensions (1+ for example)
* A very simple programming style (librares, constatns, procedures, main entry point)
* A commenting style (write a main comment before each procedure, and show an example along with the comment

It's a very basic example of a Scheme program, but I guess it is interesting for someone beginning to use SRFIs and libraries. :)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,6 +22,10 @@ The files in this directory show some examples in STklos:
 - `secho.stk` is a simple echo program written is STklos. Use the
   `--help` option to see the available options.
 
+- `dice.stk` is a very simple example of scheme program that asks for
+   a number `N`, then runs two dice `N` times, and finally prints a
+   histogram of the distribution of outcomes.
+
 - `fork-test.stk` is a simple program using the Unix `fork(2)`
   primitive.
 
--- a/examples/dice.stk
+++ b/examples/dice.stk
@@ -1,0 +1,162 @@
+#!/usr/bin/env stklos
+;;;;
+;;;; dice.stk  -- A script to throw dice and show the resulting distribution
+;;;;
+;;;; Copyright © 2024 Jeronimo Pellegrini <j_p@aleph0.info>
+;;;;
+;;;;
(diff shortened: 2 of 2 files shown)


egallesio:
Thanks for this example. I have merged your PR. 

I see that you have tried to be very standard in your writing (it seems to be aimed at students, am I right?). I think it would be interesting to add a possibility to pass the number of experiments from the command line (and use read if it is not the case). 
We could replace the run function by (probably the 4 last lines should be in a separate procedure)

```scheme
(define (main argv)
  ;; This is the main procedure. If we have arguments on the command line, we
  ;; use the first one as the number of iterations. Otherwise, we ask for this
  ;; number. Once we have that number we run the experiment and plot the histogram
  (let ((R (if (> (length argv) 1)
               ;; We have at least an argument and argv contains all the words of
               ;; the command line. If we entered at  the prompt:
               ;;     $ dice 10_000
               ;; argv will be ("dice" "10_000" ...).
               ;; We use read-from-string her to convert the argument to a number
               (read-from-string (list-ref argv 1))
               ;; No argument => ask interactively to the user
               (begin
                 (display "How many times should I throw the two dices? ")
                 (read)))))

    ;; R is the number of iterations we have to run
    (newline)
    (display "Ok, here is a histogram of the distribution:")
    (newline)
    (plot-distribution (make-histogram (run-experiment R)))))
```

This is a bit more complicate, and perhaps its doesn't fulfill your initial goal with this example. 
What do you think?

jpellegrini:
Hi @egallesio !
No, I don't think it's too complicated. Beginners like when the program looks like a "real world program". I like your version!

egallesio:
So, I'll modify it accordingly.

Issue #260, opened 2021-09-03
Optional procedure signature for primitives and library procedures (wishlist)


jpellegrini:
Hi @egallesio !
Not an essential thing, but I thought it would be nice to have a configure option that would turn signature generation early so the user could see signatures of imported libraries/SRFIs and core procedures.

jpellegrini:
Some comments:

1. About core procedures:

In `stklos.h`, we have
```c
struct primitive_obj {
  stk_header header;
  char *name;
  SCM (*code)();
  SCM plist;
};

#define PRIMITIVE_NAME(p)       (((struct primitive_obj *) (p))->name)
#define PRIMITIVE_FUNC(p)       (((struct primitive_obj *) (p))->code)
#define PRIMITIVE_PLIST(p)      (((struct primitive_obj *) (p))->plist)
```

I suppose we *could* also support formals here. Then, the user would opt for either compiling the formals with the C code or not compiling them - that would be a compile-time option. And if the user decides not to include the formals, they should not take any extra space in the generated binary (that's easy to do :)

2. `bonus.stk`, other closures (non-primitives) that come with STklos core, libraries and SRFIs:

Also a compilation option, I suppose, that would turn the parameter `compiler:keep-formals` on or off before compiling those...

jpellegrini:
@egallesio I think you just fixed this! 😄

egallesio:
Yeah, you see it. 😉 
I hope that you find the way used convenient.

PR #635, opened 2024-05-17
Add `--without-keep-formals` option to configure

jpellegrini:
Hi @egallesio !
This PR partially addresses issue #260 . It adds optional formal parameter information for closures in core STklos and provided libraries -- but *not* for primitive procedures.

## Description

This adds a new option ot the configure script.

If the user passes `--without-keep-formals` to configure, then the compiler `keep-formals` parameter will be true when compiling closures in core STklos and all provided libraries.

However, this does not affect primitive procedures, because they are C functions, and having the C compiler include the code to add the formal parameters is, although possible... Tricky, at least.

How it works:

We have three autoconf variables:

```
    STR_KEEP_FORMALS="no"    # for output at the end of this script
    KEEP_FORMALS=0           # for tmpcomp (and in the future, C files)
    SCM_BOOT_KEEP_FORMALS="" # for the Makefile compiling the boot image
```

The tmpcomp file needs to be called in the Makefiles with the shell variable
```
KEEP_FORMALS set:
KEEP_FORMALS=@KEEP_FORMALS@ tmpcomp ...
```

For the boot image, the variable `SCM_BOOT_KEEP_FORMALS` is directly used.

And... Well, autotools is reeeeeally not something I ever got to understand well, so perhaps there are better ways to achieve what I did.

## Future work (primitive procedures)

Since primitives are defined as C functions, this becomes tricky.

The options I thought:

* Do not store the formal parameters of primitives in their property list, but rather keep them as a string in the structure. **Problem**: simple, but would be asymmetric (too different from the way it's done for closures).
* Include the C code to actually set the formals in the property list.  **Problem**: maybe a bit too complex?
* Do not keep formals for primitives. **Problem**: well, we won't have formals for primitives :)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -481,6 +481,29 @@ else
     CONTROL_FX_PARAMETERS=0
 fi
 
+# -----------------------------------------------------------------------
+# Should the compiler keep the formal parameters of procedures in core
+# and libraries?
+# -----------------------------------------------------------------------
+AC_ARG_WITH(keep-formals,
+       [  --without-keep-formals  don't keep formal parameters of closures when compiling
+                          procedures in core and libraries (this does not affect compilation of closures
+                          done by the user later, the effect is to have them for out-of-the-box STklos
+                          closures. Formal parameters of primitive procedures are never kept, regardless
+                          of this option)],
+       DONT_KEEP_FORMALS="$withval", DONT_KEEP_FORMALS="no")
+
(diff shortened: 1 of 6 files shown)


jpellegrini:
Forgot to mention: the idea is to have formal parameters included by default, but allow the user to not have them for embedded systems or whatever, where size could be an issue.

PR #636, opened 2024-05-18
Optimize `in-module` for `SCHEME`

jpellegrini:
`(in-module mod symb . default)` will now use `%%in-scheme` when `mod` is `SCHEME`.

```scheme
(macro-expand '(in-module SCHEME list))

=> (apply symbol-value* 'list (find-module 'SCHEME) '())

stklos> (disassemble-expr '(in-module SCHEME list))

000:  PREPARE-CALL
001:  GLOBAL-REF-PUSH      0
003:  CONSTANT-PUSH        1
005:  PREPARE-CALL
006:  CONSTANT-PUSH        2
008:  GREF-INVOKE          3 1
011:  PUSH
012:  NIL-PUSH
013:  GREF-INVOKE          4 4
016:
```

The new code:

```scheme
(macro-expand '(in-module SCHEME list))

=> (%%in-scheme 'list)

stklos> (disassemble-expr '(in-module SCHEME list))

000:  CONSTANT             0
002:  INSCHEME
003:
```

Following is an illustration of the speedup.

```scheme
(time
  (let ((a #f))
    (repeat 10_000_000
      (set! a (in-module SCHEME list))) a))
```

Old code: around 1200 ms
New code: around  300 ms

However, this optimization is only done when there is no default passed to `in-module`, because using `with-handler` around `%%in-scheme` resulted in a huge slowdown.

If there is a way to call %%in-scheme and trap its result without using `with-handler`, maybe it could be added later.

### DIFF
--- a/lib/module.stk
+++ b/lib/module.stk
@@ -634,7 +634,10 @@ doc>
 doc>
 |#
 (define-macro (in-module mod symb . default)
-  `(apply symbol-value* ',symb (find-module ',mod) ',default))
+  (if (and (eq? mod 'SCHEME)
+           (null? default))
+      `(%%in-scheme ',symb)
+      `(apply symbol-value* ',symb (find-module ',mod) ',default)))
 
 
 ;;=============================================================================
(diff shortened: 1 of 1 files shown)


egallesio:
Nice. Merged.

Issue #351, opened 2021-12-20
Online help for SRFIs?


jpellegrini:
Hi @egallesio !

I was wondering if it'd be OK to have online `help` for SRFI procedures.

I have added help for one SRFI-217 procedure, just to see if it'd work.
The help when I search the *symbol* is always available, but if I search the procedure, then of course we need the SRFI loaded first:

```
stklos> (help 'iset?)
Help for iset?:
Documentation:
  (iset? obj)
  Returns |#t| is |obj| is an iset and |#f| otherwise.
stklos> (help iset?)
**** Error:
%execute: variable `iset?' unbound
	(type ",help" for more information)
stklos> (require "patricia")
"patricia"
stklos> (import stklos/patricia)
stklos> (help iset?)
Help for iset?:
Documentation:
  (iset? obj)
  Returns |#t| is |obj| is an iset and |#f| otherwise.
```

I'm not sure if adding documentation to all SRFIs wouldn't be too much in terms of space. 

By the way, I have tried `grep`ping `DOCDB` for `iset` and it's not there. So where is it stored?

aartaka:
@jpellegrini @egallesio just not to duplicate work, y’all might like [scheme-index repository](https://github.com/schemeorg-community/scheme-index), which is probably the biggest and machine-processable-est set of documentation and metadata on SRFIs (and standard libs!) ever.

I have already successfully used it for e.g. [scheme-index.aartaka.me](https://scheme-index.aartaka.me) and [schemedoc.ed](https://codeberg.org/aartaka/schemedoc.ed). It’s really easy to process once you know the structure (described exhaustively in scheme-index README.)

jpellegrini:
@aartaka it looks like a very good idea!  I'm not working on this right now, but it certainly seems like fetching from scheme-index is the right thing to do.

PR #637, opened 2024-05-18
Document procedures in `(scheme list)` / SRFI 1

jpellegrini:
Hi @egallesio !

This is just an idea. We could bring documentation for SRFI procedures and macros so the user can have online help.
However, I suppose there should be compile options to generate (or not) documentation:
* for SRFIs
* for STklos libraries
* for standard Scheme libraries
So far I just included the documentation in SRFI 1. And I believe I had already done for SRFI 25.

This PR also removes the implementation of `last-pair`, which is already in STklos core. I also reordered some of the procedures when writing the documentation.

### DIFF
--- a/lib/scheme/list.stk
+++ b/lib/scheme/list.stk
@@ -302,6 +302,20 @@
 ;;; Constructors
 ;;;;;;;;;;;;;;;;
 
+#|
+<doc EXT xcons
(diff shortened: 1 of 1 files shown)


jpellegrini:
This is related to issue #351 .

PR #638, opened 2024-05-18
Fix `lset=` in SRFI 1 / `(scheme list)` (small issue, side-effects only)

jpellegrini:
The reference (sample) implementation had a bug that reversed the arguments to =. The implementation has been corrected to match the SRFI text, and we do so in STklos also.

See:
https://srfi.schemers.org/srfi-1/srfi-1.html#lset%3D-element-equality-order

The difference in behavior is only for side-effects. It can be seen in this example:

```
;; OLD version of lset=
scheme/list> (lset= (lambda (x y) (print x) (= x y)) '(1 2 ) '(1 2 ))
1
2
2
1
2      ;; <= HERE!
2

;; NEW version of lset=
scheme/list> (lset=-new (lambda (x y) (print x) (= x y)) '(1 2 ) '(1 2 ))
1
2
2
1
1      ;; <= HERE!
2
```

### DIFF
--- a/lib/scheme/list.stk
+++ b/lib/scheme/list.stk
@@ -1579,15 +1579,18 @@ FILTER an FILTER! are primitives in STklos
 
 (define (%lset2<= = lis1 lis2) (every (lambda (x) (member x lis2 =)) lis1))
 
-(define (lset<= = . lists)
-  (check-arg procedure? = lset<=)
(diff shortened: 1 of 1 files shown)


egallesio:
Thanks for your vigilance @jpellegini. Merged

PR #640, opened 2024-05-20
Add optional "name" argument to `error` calls

jpellegrini:
Not in all of them, but in procedures that are publically visible and where it make ssense to inform who is triggering the error.

Does this make sense, @egallesio ?

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -1591,7 +1591,7 @@ doc>
         (loop (cons sexp res) (reader p)))))
 
 (define (port->string p)
-  (unless (input-port? p) (error "bad port ~S" p))
+  (unless (input-port? p) (error 'port->string "bad port ~S" p))
   (with-output-to-string
     (lambda () (copy-port p (current-output-port)))))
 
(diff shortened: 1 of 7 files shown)


egallesio:
Hi @jpellegrini,

In fact, the name of the procedure which caused the error can be found in the error backtrace. For some reason, if didn’t work when `error` is called in a Scheme procedure (it worked before). Fixing the problem was on my TODO list since a long time. Thanks to put it back on the top of the list with this PR.   

Now we have: 
```scheme
stklos> (define (foo) (error "BAD"))
;; foo
stklos> (foo)
**** Error:
foo: BAD
stklos> 
```

Consequently, I close this PR.

jpellegrini:
That's great!

But it does lead to some unexpected messages, no?

```scheme
stklos> (exact-integer-sqrt -1)
**** Error:
>=: non negative integer expected. It was: -1
```

egallesio:
Weird. I'll look at this.

jpellegrini:
@egallesio the commit you have pushed some minutes ago seems to have fixed the problem! :)

```scheme
stklos> (exact-integer-sqrt -1)
**** Error:
exact-integer-sqrt: non negative integer expected. It was: -1
```

egallesio:
> the commit you have pushed some minutes ago seems to have fixed the problem! :)

:open_mouth: Impressed by your reactivity! 

Yes. It should be OK now. And when calling `error`, the function which called it should be automatically found.

PR #642, opened 2024-05-22
Enhancement to help initial banner

jpellegrini:
Maybe not interesting, since the banner gets larger, but it also becomes more informative. :)


egallesio:
Good. Merged.

PR #646, opened 2024-05-29
Example of module written in C and Scheme

jpellegrini:
@egallesio I'm not sure where this example would fit, so I included as a subdirectory of `examples`. Do you think it makes sense to have this?

This  is an example of a STklos module written part in C and part in Scheme.

What we illustrate with this example:

* Combining C and Scheme to build an STklos module;
* Selectively exporting symbols from the C file;
* Referencing, in the Scheme file, symbols defined in the C file;
* Writing help for procedures, not only in Scheme files, but also in C files.

* The example was built for Linux, but can be adapted for other systems.
* In order to write STklos modules in C, it is necessary to understand how typical STklos C code works. This can be found in the `hacking` file.


egallesio:
Hi @jpellegrini,

I have integrated your example (great idea BTW). I have modified the `Makefile` to permit to compile it with the STklos configuration options and without the need to have stklos previously installed. This `Makefile` is of course more complicated (generated by a `Makefile.in` …).  So I have let the one you wrote for this PR as `Makefile-sample` for keeping things as simple as possible. 


Thanks for this PR.

PR #648, opened 2024-05-29
Detect division by zero when inlining

jpellegrini:
These two situations are similar, but one signals a compiler error, and the other signals an error after copilation:

```scheme
stklos> (/)
**** Error:
compiler-error: /: needs at least one argument
stklos> (/ 2 0)
**** Error:
/: cannot make rational with null denominator
```

With this PR, both are treated similarly.


egallesio:
Hi @jpellegrini, 

I have applied this PR and changed a bit your code since we had some inconsistencies depending on the fact we had, or not, constant folding.

PR #649, opened 2024-05-30
Rewrite rule for equal?

jpellegrini:
Hi @egallesio !
Maybe this one is interesting? :)

- If none of the arguments is constant, then keep the expression as is;
- If one of the arguments is an eq?-comparable object, rewrite into (eq? a b)
- If both arguments are constant, then we compute (equal? a b)

This is the result:

```
stklos> (disassemble-expr '(equal? a b))

000:  GLOBAL-REF-PUSH      0
002:  GLOBAL-REF           1
004:  IN-EQUAL

stklos> (disassemble-expr '(equal? a #eof))

000:  GLOBAL-REF-PUSH      0
002:  CONSTANT             1
004:  IN-EQ

stklos> (disassemble-expr '(equal? 25 #eof))

000:  IM-FALSE
```

And maybe the same could be done for `eqv?`...

### DIFF
--- a/lib/comprewrite.stk
+++ b/lib/comprewrite.stk
@@ -159,3 +159,58 @@
                    expr))
              expr))
        expr)))
+
+(compiler:add-rewriter!            ;; 'EQUAL?' rewriter
+ 'equal?
+ ;; (equal? ATOM ATOM) ==> #t or #f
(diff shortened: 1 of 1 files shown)


jpellegrini:
> And maybe the same could be done for `eqv?`...

And `member`, and `memq`, maybe?

PR #651, opened 2024-06-02
Enhancements to `expt`

jpellegrini:
* Don't use Scheme arithmentic where not needed in exponentiation by squaring (the exponent is known to be a fixnum).
* Fast path for exponent=2 (much much faster).
* Change name `exact_exponent_expt` -> `fixnum_exponent_expt`
* Move the bignum check to `my_expt`, so the exponent is always a fixnum when calling `fixnum_exponent_expt`.
* Since the exponent is always a fixnum in `fixnum_exponent_expt`, compare with `MAKE_INT(0)` instead of using `zerop`.

Results:
* for some integer exponents, `expt` is faster
* specifically, for exponent = 2, it is more than twice as fast

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -3883,19 +3883,26 @@ DEFINE_PRIMITIVE("sqrt", sqrt, subr1, (SCM z))
 doc>
  */
 
-static inline SCM exact_exponent_expt(SCM x, SCM y)
+/* fixnum_exponent_expt enad my_expt are mutually recursive: */
+static SCM my_expt(SCM x, SCM y);
+
+static inline SCM fixnum_exponent_expt(SCM x, SCM y)
 {
+  /* NOTE: 'y' MUST be a fixnum (this function will *not*
+     verify, but relies on that.  */
   mpz_t res;
   SCM scm_res;
 
-  /* y is already known to be exact; so if it is zero,
+  /* y is already known to be a fixnum; so if it is zero,
      return exact one. */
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @jpellegrini,

Just realized that I forgot to merge thi PR before working on `expt`. 
I'll try to do the adaptation.

egallesio:
Merged by hand. It seems OK.

Furthermore, `exact_exponent_expt` take a long instead of a SCM now                        

The function signature was                                                  
```c
static inline SCM fixnum_exponent_expt(SCM x, SCM y)                    
```
It is now                                                                   
```c
static inline SCM fixnum_exponent_expt(SCM x, long y)                   
```

We have fewer conversions from Scheme fixnums to longs now.                  
Not tested if it is really more efficient (it should).

jpellegrini:
It seems generally faster!

This:
```scheme
(let ((a 0)
      (b 10)
      (e 2.0))
  (time
   (repeat 10_000_000
     (set! a (expt b e))))
  a)
```

Runs i nsomething like  half the time, compared to the STklos checkout from last week. :grinning:

jpellegrini:
> Runs i nsomething like half the time, compared to the STklos checkout from last week. 😀

Even with the float exponent!

egallesio:
That's great! 
It's crazy the number of places and things to optimize/correct on a single function.

PR #654, opened 2024-06-22
Simplify error checking

jpellegrini:
@egallesio this is mostly a matter of style. See if you agree...

In the compiler, instead of

```scheme
  (if some-condition
      ( X ...
          ...
          ... )
      (compiler-error Y))
```

we do

```scheme
  (unless some-condition (compiler-error Y))
  ( X ...
      ...
      ... )
```

Or `when`, when it makes sense

This makes error checking isolated in the place where it programatically happens, and also reduces source code size a little bit, making it clearer.

There are already uses of the macros `when` and `unless` in the compiler, so we're not introducing a dependency on them.

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -312,9 +312,8 @@
 doc>
 |#
 (define (compile-quote expr env tail?)
-  (if (= (length expr) 2)
-      (compile-constant (cadr expr) env tail?)
-      (compiler-error 'quote expr "bad usage in ~S" expr)))
+  (unless (= (length expr) 2) (compiler-error 'quote expr "bad usage in ~S" expr))
+  (compile-constant (cadr expr) env tail?))
 
 ; ======================================================================
 ;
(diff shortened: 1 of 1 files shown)


jpellegrini:
I know that there are other places in STklos source that use the `if` style for error checking; I only changed the compiler, but I could in the future do the same in other files, gradually -- if you believe this is a good change.

egallesio:
Hi @jpellegrini,

The form you propose is not equivalent to the original writing. In fact `compiler-error` calls error in interactive mode, bit **it doesn't** when we compile a file (it just prints an error message). 
For instance, in

```scheme
(define (compile-quote expr env tail?)
  (unless (= (length expr) 2) (compiler-error 'quote expr "bad usage in ~S" expr))
  (compile-constant (cadr expr) env tail?))
``` 

If you call `compile-quote` with `(quote)`, you'll access the `cadr` of a list of length 1, which will abort the compilation of the rest of the file.

jpellegrini:
Oh, I saw that the new form wasn't equivalent, but I thought it wouldn't be relevant, because it passed all tests. Maybe we need more tests then! :grin:

egallesio:
> Oh, I saw that the new form wasn't equivalent, but I thought it wouldn't be relevant, because it passed all tests. Maybe we need more tests then! 😁

The problem here is that the tests (in `make test`) are not compiled, but just loaded.

jpellegrini:
So since the benefits would be minimal, and this introduces problems, I believe it would be OK to close this one?

PR #657, opened 2024-07-05
Make equality tests more precise

jpellegrini:
When comparing ints and reals, don't let the imprecision coming from conversions interfere.

Before:
```
stklos> (= 4999999999999999727876154935214080.0 5000000000000000000000000000000000)
#t
```

Now:
```
stklos> (= 4999999999999999727876154935214080.0 5000000000000000000000000000000000)
#f
```

We detect that we're comparing an inexact number with an exact integer and see if that inexact actually represents the same integer.

This does not yet fix the problem for complexes, but it's already better...


jpellegrini:
For complexes, I suppose the right thing would be to run `do_compare` on both real and imaginary parts, separately, but not do `sub2` on the complex number itself. (What do you think @egallesio ?)

egallesio:
Hi @jpellegrini,

This seems OK. I'll merge this PR asap. Thanks.

jpellegrini:
The complex number equality fix is not that hard - I'm working on it

jpellegrini:
Okay @egallesio !

The complex test is now as good as the real test. :smile:  And we have tests included.

egallesio:
Thanks for this PR @jpellegrini,

I have fixed some problems in `general_diff` which uses `d1`and `d2` which are always set to 0 (I don't understand why it worked so well).

Furthermore, tests with a real and a rational were incorrect `(= 1/3 #i1/3)` returned `#t` instead of `#f` (it seems that only Gauche returns `#t`).

I have also done some rewriting for `do_compare` since, it tested things that have already been verified before calling it, or bad numbers, whereas it is always called with numbers...

Rewriting of do_compare permits to see some other bugs 

```
(min 1 +nan.0 2)  -> +nan.0
(- 0+0.0i 0.0)    -> 0.0+0.0i (and not -0.0+0.0i)
```

That's quite a lot of problems for a single function, hope I have not introduced new ones :smile:


Closing this PR since Github didn't close it.

PR #660, opened 2024-07-20
enhancements to `vm.adoc`

jpellegrini:
* The `tick` function mentioned in the stats-gathering documentation is not the one used in the VM (it was a previous version that was never commited to  STklos). This patch includes the correct version;
* The documentation for `%vm-collect-stats` is more clear now;
* Include documentation for `%vm-config`.


egallesio:
Thanks. It is merged.

PR #661, opened 2024-07-21
More information in `QUICK-INSTALL.md`

jpellegrini:
* Include a more recent version of example of configure report (with the message about asciidoctor)
* Include environment variables and their use in the configure script
* Include `--with-gmp-light` as a synonym for `--with-provided-bignum`


egallesio:
That is perfect. I suppose it is sufficient, and it could be easily extended if needed. 
Thanks @jpellegrini

PR #663, opened 2024-07-23
Report C compiler version in configure script

jpellegrini:
It may be useful to immediately see if we're using gcc version 14 instead of 13, for example.

So now there are two lines:
* one for the compiler binary
* one for the version string (1st line reported by the compiler when called with `--version`)


egallesio:
Good idea. Merged.

PR #665, opened 2024-07-24
Optimize arithmetic for inexact real numbers

jpellegrini:
When we do "a + b" and one of the arguments is not real, it may be converted to real by the 'convert' function (in number.c).

However,  `convert` will sometimes allocate a new cell for the real number. But sometimes it is not necessary.

Then, STk_add2 will again allocate a new cell, in

```c
o1 = double2real(REAL_VAL(o1) + REAL_VAL(o2));
```

We avoid the new allocation by checking first.

We do the same for all four operations (+, -, *, /).

Some benchmarks:

BEFORE:

```
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (+ 1.1 x )))) a)
;    Elapsed time: 9830.525 ms
;    Allocations: 1600000000 bytes in 100000000 allocation calls (GC: 1187)
1.6
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (* 1.1 x )))) a)
;    Elapsed time: 9446.878 ms
;    Allocations: 1600000000 bytes in 100000000 allocation calls (GC: 1184)
0.55
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (- 1.1 x )))) a)
;    Elapsed time: 9440.226 ms
;    Allocations: 1600000000 bytes in 100000000 allocation calls (GC: 1184)
0.6
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (/ 1.1 x )))) a)
;    Elapsed time: 9519.47 ms
;    Allocations: 1600000000 bytes in 100000000 allocation calls (GC: 1184)
2.2
```

AFTER:

```
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (+ 1.1 x )))) a)
;    Elapsed time: 5664.624 ms
;    Allocations: 800000000 bytes in 50000000 allocation calls (GC: 593)
1.6
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (* 1.1 x )))) a)
;    Elapsed time: 5792.377 ms
;    Allocations: 800000000 bytes in 50000000 allocation calls (GC: 592)
0.55
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (- 1.1 x )))) a)
;    Elapsed time: 5564.795 ms
;    Allocations: 800000000 bytes in 50000000 allocation calls (GC: 592)
0.6
stklos> (let ((x 1/2) (a 1.1)) (time (repeat 50_000_000 (set! a (/ 1.1 x )))) a)
;    Elapsed time: 5978.568 ms
;    Allocations: 800000000 bytes in 50000000 allocation calls (GC: 591)
2.2
```

So, basically we have something like half the amount of allocations and GC steps, and also something close to twice the speed, when we operate on a real with some other type (different from real, and also different from complex).

When we operate on two reals, there is no difference in timing.

Passes all tests.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2144,6 +2144,8 @@ doc>
  */
 SCM STk_add2(SCM o1, SCM o2)
 {
+  SCM old1 = o1;
+  SCM old2 = o2;
   switch (convert(&o1, &o2)) {
     case tc_bignum:
         {
@@ -2168,7 +2170,24 @@ SCM STk_add2(SCM o1, SCM o2)
         }
       case tc_real:
         {
-          o1 = double2real(REAL_VAL(o1) + REAL_VAL(o2));
+          if (o1 != old1) {
+            /* We already allocated a new cell for o1, so we just
+             * modify it here! */
+            REAL_VAL(o1) = (REAL_VAL(o1) + REAL_VAL(o2));
+            if (isnan(REAL_VAL(o1)) && signbit(REAL_VAL(o1))) /* convert -nan.0 to +nan.0 */
+              REAL_VAL(o1) = STk_NaN;
+            return o1;
+          } else if (o2 != old2) {
+            /* We already allocated a new cell for o2, so we just
+             * modify it here! */
+            REAL_VAL(o2) = (REAL_VAL(o1) + REAL_VAL(o2));
+            if (isnan(REAL_VAL(o2)) && signbit(REAL_VAL(o2))) /* convert -nan.0 to +nan.0 */
+              REAL_VAL(o2) = STk_NaN;
+            return o2;
+          } else
+            /* We did NOT allocate a new cell for o1 or o2, so we just
+             * DO allocate it for o1 here! */
+            o1 = double2real(REAL_VAL(o1) + REAL_VAL(o2));
           break;
         }
       case tc_complex:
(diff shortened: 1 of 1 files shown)


jpellegrini:
@egallesio does this make sense?

I wish it was possible to not allocate reals when we do `a+b`, both being real, but that is currently not possible in STklos as I see...  But when they are different types, it is indeed possible!

jpellegrini:
If this is OK, then maybe we can also do it for other types (complexes and rationals)?

egallesio:
Hi @jpellegrini,

I have done some work on this point since a couple of days and I have done it for all the types of numbers. The code is rather long and boring, and the function `convert` is completely gone. It is of course more efficient since we have less allocation, but a pretty side effect is that some special cases have disappeared (for instance, the multiplication of a real by a infinite), but there are still some problems to be resolved before I push my code.

egallesio:
In particular, I have some regressions on  #586 and #587 that I don't understand. 
I'm afraid that this will need some more testing on some  rough corners .

jpellegrini:
> Hi @jpellegrini,
> 
> I have done some work on this point since a couple of days and I have done it for all the types of numbers. The code is rather long and boring, and the function `convert` is completely gone. It is of course more efficient since we have less allocation, but a pretty side effect is that some special cases have disappeared (for instance, the multiplication of a real by a infinite), but there are still some problems to be resolved before I push my code.

That's great! :grinning:  I look forward to see this new code!

egallesio:
@jpellegrini ,

Finally, I have committed my code... 
Now we should not  box useless numbers for the 4 operations: `+`, `-`, `*`, `/`. As I said before, the code is long and boring, but it should be OK. It also fixes some inconsistencies we had before (I had to fix some tests for that).

I close the PR, by hand

PR #670, opened 2024-07-30
Report system library versions

jpellegrini:
Like this:

```
Control fx parameters:  yes
System libraries used:  libffi (3.4.6) libpcre2 (10.42) libgmp (6.3.0) libgc (8.2.6)
   Compiled libraries:
 Documentation update:  yes (since Asciidoctor is installed)
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -309,13 +309,14 @@ else
      if $PKGCONFIG --exists bdw-gc ;then
         GCINC=$($PKGCONFIG  bdw-gc --cflags)
         GCLIB=$($PKGCONFIG  bdw-gc --libs)
+        GCVER=$(${PKGCONFIG} --modversion bdw-gc)
      else
         GCINC=""
         GCLIB="-lgc"
      fi
      DLLIBS="$DLLIBS -lgc"
      AC_DEFINE(HAVE_GC, 1, [Use system GC])
-     SYST_LIBS="libgc $SYST_LIBS"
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @jpellegrini, 

I'm still alive :wink: 

Thanks for this PR. I have shortened a bit le line to be sure to fit in less than 80 chars. 
However, when we use the SRFI-176 `-V` option, I have kept the old result form (without library version), for compatibility reasons and because it is probably more clear. 
Hope it is OK .

jpellegrini:
Hi @egallesio !
> However, when we use the SRFI-176 `-V` option, I have kept the old result form (without library version), for compatibility reasons and because it is probably more clear.

I agree it's a good idea. But I think you didn't push this commit:
```
$ stklos -V
(version "2.00.466")
(command "stklos")
(scheme.id stklos)
...
(stklos.system-libs libffi (3.4.6) libpcre2 (10.42) libgmp (6.3.0) libgc (8.0.4))
...
```

egallesio:
Ouch. Weird, I verified that was the case before committing, but I have messed my git history (with a try with  `git worktree` which I definitively don't understand :disappointed:). 
I look at that. Thanks for signaling it.

egallesio:
This should be OK now.

jpellegrini:
> This should be OK now.

Eh... Well, the commit didn't show up for me yet. (?)

egallesio:
Less affirmative :smile: : "*This should be OK now.*"

PR #671, opened 2024-07-31
Add `%make-cpointer-null` and test passing `NULL` to C functions via FFI

jpellegrini:
@egallesio I'm not sure this is a good idea, but it seems that the FFI would be mor einteresting if we could pass `NULL` to some functions. For example, `strtol`, as in the tests here.

I have included the new function with the `%` prefix.

What do you think?


egallesio:
Hi @jpellegrini, 

In fact, the Scheme value `#void` is interpreted as a C `NULL` when a pointer is expected. Hence, this function is probably not necessary.

jpellegrini:
Ah, ok! I have updated the other PR, and I'm closing this. :smile:

PR #672, opened 2024-07-31
Explain how to pass `NULL` pointer via FFI

jpellegrini:


egallesio:
Thanks for this PR. It is useful to explicit the way to pass a NULL to a C function.

PR #675, opened 2024-08-14
Add `,date` to REPL

jpellegrini:
This commit adds a `,date` command to the REPL:

```
stklos> ,date
Wed Aug 14 13:07:26 2024
stklos> ,dt
Wed Aug 14 13:07:27 2024
```

It uses the internal `date` primitive.

### DIFF
--- a/lib/repl.stk
+++ b/lib/repl.stk
@@ -245,6 +245,11 @@ doc>
   (repl-add-command '(shell !)
                     "Run a shell command"
                     (lambda () (system (read-line))))
+
+  ;; ==== date
+  (repl-add-command '(date dt)
+                    "Print the current date and time"
+                    (lambda () (print (date))))
   ;; ==== time
   (repl-add-command '(time t)
                     "Print the time used to run the next expression"


egallesio:
That's funny, because I had something like that under the hood: the idea was to use a sole `!` for shell commands (as `ed(1)` or `vi(1)` :wink:. This is more general that what we have now, and we could delete `,ls` and `,shell` and `,pwd` . To have the date, we could just have `!date`  (or even `!LC_ALL=fr_FR.utf8 date` to have it in another locale). 
When we have `!emacs foo` we just have to verify that we haven't a `!emacs` variable in the environment and launch `emacs`. I have even tested to not have the `!`, and just `emacs` at the prompt suffices to launch `emacs`. The code is simpler, but the messages are more error-prone (if you want to see the value of the variable 'the-unbound-variabbble`, you'll have a message telling you that a command was not found :smile:). Furthermore, `date` will return you  `#[primitive date]` and not the current date. 
I have the code ready (need to alter the repl help message) and I push a version for that.

jpellegrini:
That sounds great! :smile:

egallesio:
With last commits, `,date` can be obtained with `!date` (but it uses an external process).
Perhaps, we can close this PR

jpellegrini:
Yes, agreed! :)

PR #676, opened 2024-08-23
Allow STklos to compile without the FFI

jpellegrini:
Hi @egallesio !

This PR changes `ffi.c`, `configure.ac` and `do-test.stk` to allow STklos to be compiled without the FFI.

I did this because the FFI lib may not be available in some environments, and also because it may be interesting to have STklos running without the 47k taken by the libffi shared object file...
In particular, I couldn't compile STklos with musl libc in order to run on my wireless router (it used to be possible, but now for some reason it complains about the FFI -- but I don't need it in that context anyway).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -440,7 +440,7 @@ echo "*****"
 # *** If it's set to "no", then the --with-provided-ffi won't even be
 #     verified.
 AC_ARG_WITH(ffi,
-       [  --without-ffi don't include FFI code ],
+       [  --without-ffi           don't include FFI code ],
        AVOID_FFI="yes", AVOID_FFI="no")
 
 if test "$AVOID_FFI" = "no"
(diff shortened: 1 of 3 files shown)


jpellegrini:
Perhaps there is a nicer way to organize the options in `configure.ac`. Maybe

```
  --with-ffi=option
```
where option could be `system`, `provided` or `none`?

But if we do that then what if the user does
```
--with-ffi=system --with-provided-ffi
```
?

lassik:
musl support is valuable, for e.g. Alpine Linux and Void Linux.

jpellegrini:
> Alpine Linux and Void Linux.

And OpenWRT. :smile:

egallesio:
Hi @jpellegrini, 

Interesting, I'll try to merge this PR asap.

egallesio:
Finally done. I'm sorry for the delay.

PR #677, opened 2024-08-24
Update output example for `./configure`

jpellegrini:


egallesio:
Merged. Thanks, @jpellegrini.

PR #680, opened 2024-08-29
Document `initialize`

jpellegrini:
Hi @egallesio !

I needed the CLOS `initialize-instance` method and couldn't find it, then I found that STklos implements it with the name "`initialize`"... And also found that it wasn't documented yet, so I have written some documentation. Is the explanation I gave OK?

### DIFF
--- a/doc/refman/object.adoc
+++ b/doc/refman/object.adoc
@@ -682,6 +682,7 @@ we could define the following method to compare points:
 
 {{insertdoc 'define-class}}
 {{insertdoc 'make-instance}}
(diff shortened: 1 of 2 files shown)


egallesio:
This is perfect. 
I have just placed the documentation with `make` and finally renamed `initialize` as `initialize-instance` (since its friend is called `allocate-instance`). Of course, `initialize` is still usable. 
 
I have also added the definition of `make` at the end, of your explanation to explicit the relation between `make`, `allocate-instance`, and `initialize-instance`.

Thanks a lot for this PR.

PR #681, opened 2024-09-01
Add an example of object oriented program

jpellegrini:
Hi @egallesio !
One more example. This one has a mock `ansi-color?` procedure, but it would work better after PR #679 :)

This is the usual "geometric shapes" example, but one can actually draw the shapes on a terminal-based system! With colors!

* This is an example of medium difficulty. It illustrates several object orientation concepts *as they are implemented in STklos*.
* The geometric shapes are not pretty. It's an example of a object oriented program, and the focus is not on doing ASCII prettyfication of the figures.
* I have added some exercises at the end of the file. :)
* Not the best possible OO design, but I was focusing more on trying to exemplify several different CLOS concepts. :)

Concepts illustrated:

1. Class definition, `#:accessor`, `#:init-keyword`, `#:init-form`
2. Class inheritance
3. Generic methods, including the use of `next-method` inside `initialize`
4. Using *only* the class of an object to get different behavior (the initializers for `<triangle>` and `<parallelogram>` are mostly the same, but the objects will be drawn differently *because they are instances of different classes*)
5. `display-object` and `object-equal?`
6. One case in which `syntax-rules` would get much more complex than low-level macros
7. Specific STklos procedures: `1+`, `1-`, `key-get`

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -29,6 +29,9 @@ The files in this directory show some examples in STklos:
 - `fork-test.stk` is a simple program using the Unix `fork(2)`
   primitive.
 
+- `ascii-draw.stk` is an example of object-oriented program, using
+  a large part of the STklos CLOS-like system.
+
 - `threads.stk` is a program with 3 threads. It shows how to use
   threads and mutexes.
 
--- a/examples/ascii-draw.stk
+++ b/examples/ascii-draw.stk
@@ -1,0 +1,616 @@
+;;;;
+;;;; ascii-draw.stk  -- An object-oriented program to draw geometric
+;;;;                    shapes using ASCII characters
+;;;;
+;;;; Copyright © 2024 Jeronimo Pellegrini <j_p@aleph0.info>
+;;;;
+;;;;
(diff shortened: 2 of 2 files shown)


jpellegrini:
> * Not the best possible

In particular, I don't like the `#:points list-of-points` idea... But anyway, it's already there.

jpellegrini:
To see how it works, do this:

```scheme
(load "ascii-draw.stk")

(define v (make <canvas> #:rows 20 #:columns 40))
(define p (make <point>  #:x 10 #:y 25))
(define i (make <circle> #:center p #:radius 8 #:color 'red))

(define q (make <point>  #:x 5  #:y 8))
(define r (make <point>  #:x 10  #:y 7))
(define s (make <point>  #:x 7  #:y 4))
(define a (make <parallelogram> #:points (list q r s) #:color 'blue))

(draw v i)
(display v)
(draw v a)
(display v)

(reset! v)
(display v)

(define g (make <polygon> #:points (list q s r p) #:color 'blue))
(draw v g)
(display v)

(set! (canvas-background v) white)
(display v)
```

PR #682, opened 2024-09-01
More OO documentation

jpellegrini:


egallesio:
Thanks @jpellegrini,

That's really great. Merged.

PR #687, opened 2024-09-15
Fix Android build instructions

### LINKED ISSUE #658: Small build problem on android
Hi @egallesio !

I got a new Android phone, and tried compiling STklos.

It initially seems that PCRE will work:

```
***** Regexp support
*****
checking for pcre2_config_8 in -lpcre2-8... yes
... Using system libpcre2 library
*****
```
And the summary is:
```
SUMMARY
*******
               System:  Linux-4.14.113-g1cb23d5d0f9e
              OS nick:  LINUX_4_14
              OS type:  unix
       Install prefix:  /data/data/com.termux/files/usr
           C compiler:  gcc
    Compilation flags:  -g -O2
               Loader:  ld
       Thread support:  pthreads
Control fx parameters:  yes
System libraries used:  libffi libpcre2 libgmp libgc
   Compiled libraries: 
 Documentation update:  no (install Asciidoctor to rebuild documentation)
```

However, the generated Makefile does not link with `-lpcre2-8`, as determined in the configure phase, but with `-lpcre-8` (without the `2`):
```
gcc  -g -O2 -rdynamic  -o stklos base64.o boolean.o boot.o box.o char.o cond.o cpointer.o dynload.o env.o error.o extend.o ffi.o fixnum.o fport.o gnu-getopt.o hash.o keyword.o lib.o list.o misc.o md5.o number.o object.o parameter.o path.o port.o print.o proc.o process.o promise.o read.o regexp.o signal.o sio.o socket.o sport.o stklos.o str.o struct.o symbol.o syntax.o system.o utf8.o uvector.o vector.o vm.o vport.o thread-common.o thread-pthreads.o mutex-common.o mutex-pthreads.o -lgmp -lpcre-8 -lpcre2-posix -lffi -lgc -lm -lpthread 
ld.lld: error: unable to find library -lpcre-8
gcc: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [Makefile:584: stklos] Error 1
make[2]: Leaving directory '/data/data/com.termux/files/home/STklos/src'
make[1]: *** [Makefile:489: all] Error 2
make[1]: Leaving directory '/data/data/com.termux/files/home/STklos/src'
make: *** [Makefile:438: all-recursive] Error 1
```

I'm not sure why this happened, or how to fix it...

jpellegrini:
* In recent Android versions, one needs to pass LDFLAGS to point ld to the right path
* ld itself is in a different package, binutils-is-llvm

Should fix #658

### DIFF
--- a/INSTALL-android.md
+++ b/INSTALL-android.md
@@ -141,7 +141,7 @@ you are in a directory specifically created for Termux in your extenal SD card).
 After ssh-ing into the device, install the build tools. Termux does not have gcc, only clang.
 
 ```
-apt install libgc libpcreposix make clang git autoconf automake
+apt install libgc libpcreposix make clang git autoconf automake binutils-is-llvm
 ```
 
 Of course, clone the repository (or you can download a release tarball and extract it):
@@ -153,7 +153,7 @@ git clone https://github.com/egallesio/STklos
 Configure: in order to install STklos for usage within Termux, we use the prefix as below.
 
 ```
-./configure --prefix=/data/data/com.termux/files/usr/
+./configure LDFLAGS=-L/system/lib64/ --prefix=/data/data/com.termux/files/usr/
 make
 make test
 ```
@@ -170,10 +170,10 @@ Now run STklos!
 
 ```
 stklos -i
-  \    STklos version 1.40  (Id: 58dfbe0)
-   \   Copyright (C) 1999-2020 Erick Gallesio <eg@unice.fr>
-  / \  Université Côte d'Azur
- /   \ [Linux-3.18.140-perf-g1b0d9102ccae-aarch64/none/readline/utf8]
+  \    STklos version 2.00.565 (unstable -- 6338e863)
+   \   Copyright (C) 1999-2024 Erick Gallesio <eg@stklos.net>
+  / \  [Linux-4.14.113-g1c8cbedf56fe-aarch64/pthreads/readline/utf8]
+ /   \ Type ',h' for help
 stklos>
 ```
 
@@ -257,7 +257,7 @@ git clone https://github.com/egallesio/STklos
 Configure, `make`, `make test` and `make install`:
 
 ```
-./configure --prefix=/data/data/com.termux/files/usr/
+./configure LDFLAGS=-L/system/lib64 --prefix=/data/data/com.termux/files/usr/
 make
 make test
 make install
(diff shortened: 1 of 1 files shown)


jpellegrini:
However, in Android loading libraries doesn't seem to work, because `%library-prefix` is not set properly.

In `extraconf.h`, this is what we have:

```c
#define PREFIXDIR               "/data/data/com.termux/files/usr"
#define SCMDIR                  "${datarootdir}"
#define DEFAULT_BOOT_FILE       "${datarootdir}/boot.img"
#define EXECDIR                 "${exec_prefix}/lib"
#define BUILD_OS                "Linux-4.14.113-g1c8cbedf56fe"
```

So we will have:

```
stklos> (load-path)
("." "${exec_prefix}/lib" "${datarootdir}")
```

Now if I do 

```
./configure LDFLAGS=-L/system/lib64/  --prefix /data/data/com.termux/files/usr/  
                   --datadir=/data/data/com.termux/files/usr/share/stklos 
                   --libdir=/data/data/com.termux/files/usr/lib/stklos
```

Then STklos will not look for `.so` files, even when loading readline:

```
~/STklos $ stklos
#[condition &error-message 510340201536]: cannot open object file /data/data/com.termux/files/usr/lib/stklos/readline-complete.so (dlopen failed: cannot locate symbol "rl_completer_word_break_characters" referenced by "/data/data/com.termux/files/usr/lib/stklos/readline-complete.so"...)
**** FATAL ERROR: no handler present!
ABORT
```

But if I use the `-n` flag,

```
~/STklos$ stklos -n
  \    STklos version 2.00.565 (unstable -- 6338e863)
   \   Copyright (C) 1999-2024 Erick Gallesio <eg@stklos.net>
  / \  [Linux-4.14.113-g1c8cbedf56fe-aarch64/pthreads/readline/utf8]
 /   \ Type ',h' for help
stklos> (import (scheme list))
stklos>
```

It loads libraries...

This only happens in Android...

jpellegrini:
On my Android phone:
```
$ strings /data/data/com.termux/files/usr/lib/stklos/2.10/readline-complete.so|grep rl_completer_word_break_characters
rl_completer_word_break_characters
rl_completer_word_break_characters
```

On my notebook:
```
strings /usr/local/lib/stklos/2.10/readline-complete.so |grep rl_completer_word_break_characters
rl_completer_word_break_characters
rl_completer_word_break_characters
rl_completer_word_break_characters
```

One more reference on the notebook...

jpellegrini:
@egallesio I think this is **definitely** fixed! Tested and works fine, with readline and completion on the REPL!

PR #690, opened 2024-09-20
Do `(cd DIR && rm ...) || true`

### LINKED ISSUE #689: Makefiles use `(cd some-dir ; do-something)`
I was working om the Android port, and noticed that `make uninstall` removed files from the *source* tree.

This is because the Makefiles do

```
   (cd some-dir ; rm -rf ...)
```

But if `some-dir` doesn't exist... It will remove the files in the *current* directory!

Perhaps most uses of `;` in the Makefiles should be changed?

But using `&&` is not good either, since this will make the uninstall target fail if the files are not in place.

jpellegrini:
Otherwise we may end up removing source files when uninstalling.

Should fix #689

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -70,15 +70,15 @@ SUPPORTED-SRFIS: lib/srfis.stk
 	mv -f  $@.tmp $@
 
 clean-am:
-	for i in doc tests utils extensions;do (cd $$i; $(MAKE) clean) ;done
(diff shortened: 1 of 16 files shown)


jpellegrini:
@egallesio I have use d`(cd ... && rm ...) || true` when removing files, but I also changed `(cd DIR; make)` to `(cd DIR && make)`, so the build will fail on missing subdirectoies.

egallesio:
Great. Thanks, @jpellegrini. 
Merged.

PR #692, opened 2024-09-22
Add tests for case-insensitive reads

jpellegrini:
:grin:


egallesio:
Ouch, you were really fast with this one @jpellegrini. :open_mouth:
Thanks.

PR #695, opened 2024-10-10
Add SRFI-234: Topological Sorting

jpellegrini:
Hi @egallesio !

IT's great that we already have STklos 2.10!  Now, for the next release, here's one more SRFI... :grin:


egallesio:
> Now, for the next release, here's one more SRFI... 😁

I have seen that this one has been finalized a couple of days ago. It was on my TODO list, but, as usual, you were very fast. 

Thanks for this PR

PR #698, opened 2024-10-14
Making cross-compilation easier

jpellegrini:
@egallesio : 

I have been cross-compiling STklos for wireless routers (and it works really well - STklos has a very short startup time, and that makes a huge difference on some devices!)

But some recent changes made cross-compilation a bit more difficult. Looking into it, I see it would actually be nice if a few changes were applied to the `Makefile`s and also to `tmpcomp`/`tmpgenlex`.

### 1 -  Use `$STKLOS_BINARY`, not `src/stklos` during compilation
    
This is a very small change, but is crucial for cross-compilation.
    
When the variable `STKLOS_BINARY` is set, it will be used (during cross-compilation, it will be the locally installed stklos); when it is not set, then `${prefix}/src/stklos` is used.
    
Also, in `tmpcomp`, the prefix was computed as `../`, relative to utils. But in `tmpgenlex` it was hardcoded. This PR also changes `tmpgenlex` to do the same as `tmpcomp`.

### 2 -  Don't use `$@` when calling make recursively
    
This seems to break cross-copilation.
    
The program `tmpcomp` uses the environment variable `STKLOS_BINARY`  to tell which STklos executable it can use. However, when we use this rule:
  
```
../streams/derived.ostk: ../streams/primitive.ostk:
        (cd ../streams && $(MAKE) $@)
```
    
that variable is not passed recursively to the child `make`, and  the process fails during cross-compilation.
    
If we do this:

```
../streams/derived.ostk:
        (cd ../streams && $(MAKE) derived.ostk)

../streams/primitive.ostk:
        (cd ../streams && $(MAKE) primitive.ostk)
```
    
It (magically?) works.

### 3 - A short text explaining how to cross-compile 

Basically, it says one should set the `STKLOS_BINARY` variable.

### DIFF
--- a/INSTALL-cross-compile.md
+++ b/INSTALL-cross-compile.md
@@ -1,0 +1,36 @@
+# Cross-compiling STklos
+
+## How to cross-compile
+
+Cross-compilation is just like compilation
+
+```
+./configure prefix=some-tmp-dir CC=some-C-compiler CFLAGS=your-c-flags ...
+make
+make install
+```
+
+The relevant environment variables are described in the output
+of `configure --help`.
+
+***However:***
+
+* You *need* a STklos binary in order to compile STklos. When not
+  cross-copiling,
+  - the binary is generated as `src/stklos`
+  - the build proceeds using `src/stklos` as the STklos binary that
+    can be used to compile `.stk` files
+  However, when cross-compiling `src/stklos` will be a binary for a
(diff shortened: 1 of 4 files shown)


egallesio:
Thanks @jpellegrini,

I have merged your PR.

egallesio:
Really weird, the issue with "$@" BTW

PR #700, opened 2024-10-14
Don't reference char of empty symbol's name

jpellegrini:
```
stklos> ||
**** Error:
string-ref: index '0' out of bound in string '""'
```

This is because, if the line begins with a symbol, the REPL will try to look into its name to check if it's "!", (a shell command).  But the empty symbol has no chars in its name...

With this PR, the behavior is:

```
stklos> ||
**** Error:
%execute: symbol '||' unbound in module 'stklos'
```

### DIFF
--- a/lib/repl.stk
+++ b/lib/repl.stk
@@ -481,6 +481,7 @@ doc>
 
              ((and (symbol? e)
                    (not (symbol-bound? e))
+                   (not (eq? e '||))
                    (char=? (string-ref (symbol->string e) 0) #\!))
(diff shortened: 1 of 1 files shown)


jpellegrini:
@egallesio - sorry, I found a better way of implementing it (obviously, don;t call `zero?` and `string-length`, just compare to `||`), so I force-pushed the new version...

egallesio:
Thanks for this PR @jpellegrini
Merged.

PR #701, opened 2024-10-15
More portable use of test in configure.ac

jpellegrini:
1. Don't test `$VAR` against `""`

Instead, do `"X$VAR" == "X"`

2. Don't use `==`, use `=` for comparing strings.

This *seems* to *partially* fix the current Android compilation problems...

Actually, the Android problems remain, but with this PR, passing data and lib dir doesn't seem to be necessary (the report in the final step of configure is correct).

But it still fails to load readline.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -134,18 +134,18 @@ esac
 #   pdfdir = ${docdir}/pdf
 #   htmldir = ${docdir}/html
 # Of course, if the user specify any option, we use the value he/she gave
-if test "$DOCDIR" == "" ; then
+if test "X$DOCDIR" = "X" ; then
     # user has not given a "--docdir" option. Use our own convention
     docdir="${prefix}/share/doc/${PACKAGE_TARNAME}/${VERSION}"
     DOCDIR=${docdir}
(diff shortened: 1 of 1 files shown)


jpellegrini:
> This _seems_ to _partially_ fix the current Android compilation problems...
> 
> Actually, the Android problems remain, but with this PR, passing data and lib dir doesn't seem to be necessary (the report in the final step of configure is correct).

Also, before this patch, `configure` reported some errors (`test: unexpected operator`) -- that seemed to be caused by the use of `==` (when cross-compiling, a strict POSIX shell was used)

egallesio:
Thanks for this PR. 

I think that the problem you see comes from the use of `==` only. This is an old C programmer reflex :rofl: 

 I think that the trick of using `"X$VAR" = "X"` for testing an empty value is no longer useful with current shells. Using `"$VAR" = ""` should be sufficient. All the shells I have tested return now 1 to the expression `set "" && echo $#`. Anyway, your solution is safer (in French, we say "belt and suspenders" :smile:)

PR #702, opened 2024-10-24
Document `declare-new-error`

jpellegrini:
Hi @egallesio !
Did I get this right? :)

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -465,7 +465,7 @@ doc>
          (expr (read-ci p)))
     (close-port p)
     expr))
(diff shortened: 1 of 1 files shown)


jpellegrini:
I didn't know exactly where to include it in the reference manual, so I did not do that...

egallesio:
Thanks a lot for this PR. 
I have finally renamed `declare-new-error` to `define-new-error` (to be similar to other forms). Furthermore, I have moved
the form in SRFI-35 since it can be used only if the SRFI is loaded.

PR #711, opened 2025-01-29
[R7RS] Implement ellipsis escaping, as per R7RS

### LINKED ISSUE #704: `syntax-rules` does not support patterns beginning with ellipsis, like `(... ...)`
Hi @egallesio - happy holidays!

In page 24 of R7RS, the specification of patterns for `syntax-rules` contains this:

> A template of the form `(<ellipsis> <template>)` is identical to `<template>`, except that ellipses within the template have no special meaning. That is, any ellipses contained within `<template>` are treated as ordinary identifiers. In particular, the template `(<ellipsis> <ellipsis>)` produces a single `<ellipsis>`. This allows syntactic abstractions to expand into code containing ellipses.

```scheme
(define-syntax be-like-begin
  (syntax-rules ()
    ((be-like-begin name)
      (define-syntax name
        (syntax-rules ()
          ((name expr (... ...))
            (begin expr (... ...))))))))

(be-like-begin sequence)
(sequence 1 2 3 4)      ;; => result should be 4
```

But in STklos this raises an error when using the macro:

```scheme
stklos> (be-like-begin sequence)
**** Error:
map: malformed list '#f'
```

I'll take a look into this, probably before the beginning of 2025 -- but I wanted to report the issue already.

jpellegrini:
In Section 4.3.2 of R7RS (page 23 of the PDF), it is said that a <template> can be one of:

1. `(<element> ...)`
2. `(<element> <element> ... . <template>)`
3. `(<ellipsis> <template>)`
4. `#(<element> ...)`

Number (3) is an escaping pattern. Inside ir, ellipses are treated like a symbol, and not like the syntax-rules ellipsis. And that template *must* have only two elements.

`(<ellipsis> <template>)`
should be transformed into
`<template>`

but protecting the ellipses (they have no special meaning there).

So

```
(... a)          => a
(... (a b ... c) => (a b ... c)
(... ...)        => ...
(... a b)        => error (too many elements)
```

but the ellipses on the right side are ignored by `syntax-rules`.

The implementation is:

1. In MBE, include a check for malformed escapes (as in the last case above, `(... a b)`)
2. In MBE: `(mbe:subst-ellipsis expr ell new-ell)` will go through `expr`, and substitute *only* the sublists that match the form of an escaping ellipsis, doing the appropriate transform. However, inside the subtemplate, the ellipses will be transformed into a symbol which is internal to MBE, produced by gensym (so no match will be possible). This is in `find-clause`
3. in MBE, "replace" can be used to get the original ellipses back. This is also in `find-clause`
4. In runtime-macros, `syntax-rules` calls the internal MBE procedure `check-wrong-ellipsis-escape`
5. In MBE, `let-syntax` also does that.

Some tests, including the example in R7RS, were added.

Fix #704

### DIFF
--- a/lib/mbe.stk
+++ b/lib/mbe.stk
@@ -168,6 +168,8 @@ doc>
 (define mbe:ellipsis-sub-envs           #f)
 (define mbe:contained-in?               #f)
 
+(define ellipsis-marker (gensym "ELLIPSIS-"))
+
 ;=============================================================================
 
 ;; reverse assq:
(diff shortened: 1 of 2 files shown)


jpellegrini:
Oh, this should fix #704

jpellegrini:
There is a glitch in this, I'm fixing.

jpellegrini:
Ok, I got it working. But one of the examples does not work with let-syntax. I think it's something in let-syntax; I'll take a look later. Its code seems different from that of syntax-rules

jpellegrini:
@egallesio I just fixed this so it works with the recent changes to `mbe.stk` and `runtime-macros.stk`. It implements the R7RS ellipsis patterns properly, and passes all tests...

egallesio:
I have tried to merge this PR in a hurry (I will have some surgery tomorrow, and will be probably far from a computer for a couple of weeks). :disappointed:

jpellegrini:
Hey @egallesio no need to hurry!
STklos can wait, your health is more important! Good luck tomorrow!

PR #713, opened 2025-01-30
Add `quote-expanded`

### LINKED ISSUE #712: Can't macro-expand syntax defined with `let-syntax`
```scheme
(let-syntax ((%f (syntax-rules ()
                    ((%f) (format (... "2"))))))
        (%f))

 => "2"  (correct!)
```

But

```scheme
(let-syntax ((%f (syntax-rules ()
                    ((%f) (format (... "2"))))))
        (macro-expand* '(%f)))

 => (%f)  -- hmm, we could do better, maybe?
```

(Although Gauche does the same)

jpellegrini:
Five lines added to the compiler, nothing else :grin:   (Not counting the tests, of course)

`(quote-expanded form)`

`Quote-expanded` will first expand then quote (without evaluating) `form`, and it can be used not only on the REPL or on top-level contexts, but also inside `let-syntax`.

```scheme
(let-syntax ((%f (syntax-rules ()
                   ((%f) (format  "2")))))
  (quote-expanded (%f)))                    => (format "2")

(let-syntax ((%f (syntax-rules ()
                   ((%f x) (display x)))))
  (quote-expanded (%f '(a b c))))           => (display '(a b c))
```

Fixes #712

### DIFF
--- a/lib/compiler.stk
+++ b/lib/compiler.stk
@@ -314,6 +314,31 @@ doc>
       (compile-constant (cadr expr) env tail?)
       (compiler-error 'quote expr "bad usage in ~S" expr)))
 
+#|
+<doc syntax quote-expanded
+ * (quote-expanded form)
+ *
+ * |Quote-expanded| will first expand then quote (without evaluating)
+ * |form|, and it can be used not only on the REPL or on top-level
+ * contexts, but also inside |let-syntax|.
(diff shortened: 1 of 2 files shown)


jpellegrini:
Just fixed this for the new special forms system.

PR #716, opened 2025-01-30
Add SRFI 239 - Destructiring Lists

jpellegrini:
A define-macro implementation for this one. (Less than 100 lines of code...)

Tests included!

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -141,5 +141,6 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-235: Combinators
     - SRFI-236: Evaluating expressions in an unspecified order
     - SRFI-238: Codesets
+    - SRFI-239: Destructuring lists
     - SRFI-244: Multiple-value definitions
     - SRFI-253: Data (Type-)Checking
--- a/lib/srfi/239.stk
+++ b/lib/srfi/239.stk
@@ -1,0 +1,114 @@
+;;;;
+;;;;
+;;;; 239.stk         -- Implementation of SRFI-239
+;;;;
+;;;; Copyright © 2025 Jeronimo Pellegrini <j_p@aleph0.info>
+;;;;
+;;;;
+;;;; This program is free software; you can redistribute it and/or modify
+;;;; it under the terms of the GNU General Public License as published by
+;;;; the Free Software Foundation; either version 3 of the License, or
+;;;; (at your option) any later version.
+;;;;
+;;;; This program is distributed in the hope that it will be useful,
+;;;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;;;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;;;; GNU General Public License for more details.
+;;;;
+;;;; You should have received a copy of the GNU General Public License
+;;;; along with this program; if not, write to the Free Software
+;;;; Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307,
+;;;; USA.
+;;;;
+;;;;           Author: Jeronimo Pellegrini [j_p@aleph0.info]
+;;;;    Creation date: 30-Jan-2025 10:12
+
+
+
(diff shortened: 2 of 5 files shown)


egallesio:
Hi @jpellegrini,

 I have tested this PR. It doesn't work for the SRFI example

```scheme
(define type-of
    (lambda (obj)
      (list-case obj
                 [(_ . _) 'pair]
                 [() 'null]
                 [_ 'atom])))
```
The problem here is that we have the symbol `_` which is bound two times in a `let`. A solution could be to use `let*` (not very pretty, but this is a quick fix), but I have seen that in the `var-clause` you test explicitly for the `_` symbol.
What do you think?

jpellegrini:
A solution could be to use `let*`

Not sure this will work. There is another problem:

```scheme
(list-case '(_ . 2) [ (a . b)  (cons a a) ] )   => (|_| . |_|)
(list-case '(_ . 2) [ (a . b)  (cons b b) ] )   => (2 . 2)
(list-case '(_ . 2) [ (a . b)  (cons b a) ] )   => (2 . |_|)
```

Hm, need to think a bit more about this.

jpellegrini:
Wait, I tested it backwards!
This is what I meant:
```scheme
(list-case '(1 . 2) [ (_ . b)  (cons _ b) ] )     => (1 . 2)
(list-case '(1 . 2) [ (_ . _)  (cons _ _) ] )      => (2 . 2)
```

```

**Comment by jpellegrini** [comment]
This is what Gauche does:

```scheme
(list-case '(1 . 2) [ (_ . b)  (cons _ b) ] )  => (#<syntax null#_> . 2)
(list-case '(1 . 2) [ (_ . _)  (cons _ _) ] )  => (#<syntax null#_> . #<syntax null#_>)
```

egallesio:
One way could be to generate an uninterned symbol when we encounter a `_` to ensure it will not be used. 
???

jpellegrini:
> One way could be to generate an uninterned symbol when we encounter a `_` to ensure it will not be used.

Yes.
The SRFI says `_` is the same auxiliary syntax used in syntax-rules. We don't have a different object for that in syntax-rules. And this is funny:

(define-syntax f (syntax-rules  () ((_) _)))
(f)

```
Gauche: #<syntax null#_>
Gambit: error
Chibi: error
MIT: error
Kawa <macro _>
Sagittarius: #<syntax _>

STklos: #[syntax f]
```

Not sure what to do. I think that conceptually what you said is perfect. So maybe just using a `let*` is ok?

jpellegrini:
> using a `let*` is ok

And generating the uninterned symbol, as you suggested

egallesio:
> (But this is not the SRFI intention)
Yes, the intent is  is not a general pattern matching function (there is  another incoming SRFI for that).

jpellegrini:
> let `(_ . _)` mean "two identical values in CAR and CDR?

That would require a decision on what equality operator to use in the expression:

```scheme
(list-case '( (1 . 2) . (1 . 2) ) [ (a . a)  "equal car and cdr" ] )   ;; => ?
```

So it's better to just forbid two identical vars, I suppose? And hence, we can keep the `let`, since we don't actually need to generate bindings for the `_` variables:

```scheme
(pp (macroexpand '(list-case '(2 . 1) [ (_ . _) (print "it' s a pair!") ] )))
 =>
(let ((list-case-expr:2 '(2 . 1)))
  (cond ((pair? list-case-expr:2)
         (let () (print "it' s a pair!")))           ;; <== aha!
        ((null? list-case-expr:2)
         (error 'list-case
                "null list not allowed: ~s"
                list-case-expr:2))
        (else
         (error 'list-case
                "non-list not allowed: ~s"
                list-case-expr:2))))
```

And:

```scheme
(list-case '( (1 . 2) . (1 . 2) ) [ (a . a)  "equal car and cdr" ] ) 

list-case: double variable in clause: a
```

What do you think?

jpellegrini:
> What do you think?

I just pushed a commit that does this.  Is it OK?

jpellegrini:
Sorry, I had to force-push to get it all fixed. It should be OK now.

PR #717, opened 2025-01-30
[trivial] Add missing character in `macro-expand*` documentation

jpellegrini:


egallesio:
Thanks for the tags in the titles :smile: 

Merged.

PR #720, opened 2025-02-02
Make `list-copy` work on lists with cycles

### LINKED ISSUE #719: `list-copy` crashes on lists with cycles
```
stklos> (list-copy '#0=(1 2 3 . #0#) )
Segmentation fault
```

This is because its implementation is non-tail recursive, and does not detect cycles.

jpellegrini:
The current `list-copy` crashes on lists with cycles:
    
```
stklos> (list-copy '#0=(1 2 3 . #0#) )
Segmentation fault
```

This is because its implementation is non-tail recursive, and does  not detect cycles.

This patch makes `list-copy` detect cycle, and correctly copy the list.
    
About the cycle detecting and copying algorithm:
    
1. Detect cycles using Floyd's algorithm.  The algorithm runs two pointers, "fast" and "slow" through the list. Both are placed at the `CAR`, then at each step, slow goes forward one link, and fast goes forward two links.
    
2. If there is a cycle:
    
   2.i) we position fast back at the CAR and now move the two  pointers *one* link at each time. The pointers will *necessarily meet* exactly *at the beginning of the cycle.
    
   2.ii) If there is a cycle, we break the list in two parts. For example, the cycle below starts at `C`:
```    
                        +--------------+
                        |              |
                        v              |
              A -> B -> C -> D -> E -> F
```    
              We break the links from B to C and from F to C:
```    
              A -> B    C -> D -> E -> F
```    
Then we just copy the two lists and put back the links.
    
3. If there is no cycle, use a simple copying subroutine to copy  the list.
    
Remarks:
    
1. See that when there are cycles, we change the links *in the  original list* also (but we fix it before returning).
    
2. We use `STk_last_pair`, which is not optimal. But we also expect  lists with cycles not not be frequent, and also to not be huge.  And `STk_last_pair` is linear anyway.
3. The code still works for improper lists (as did the previous  code).
4. The performance seems the same for cycle-free lists, since the  cycle test is very fast:
    
```scheme
(import (scheme list))
(define L (iota 100_000))
(time (repeat 100 (list-copy L)))
```
    
This stays around ~430 ms on the same machine, with the old and the new version of `list-copy`.
    
5. However, since the algorithm used to actually do the copy is still non-tail recursive, `list-copy` will still crash on very long lists:
    
```scheme
(define L (iota 10_000_000))
(list-copy L)   ;; <= segfault
```

But this already happened before, and I'd say it's OK since for deep copying one does need a stack anyway (either implicitly with recursion or explicitly).
    
Remaining:
    
There is one case that is still not ok: if the **CAR** of the list contains a reference to a previous node, `list-copy` will crash    (as it did before).  This is probably very, very unusual.

Should at least partially fix #719

### DIFF
--- a/src/list.c
+++ b/src/list.c
@@ -232,7 +232,7 @@ DEFINE_PRIMITIVE("%cxr", cxr, subr2, (SCM l, SCM name))
    * NOTE: using strings (instead of keywords) is less efficient because
    * the char * is at the end of the object. Using symbols is also fast
    * (even a bit faster, don't know why), but it is harder  to detect that
-   * that we can inline when we have (%cxr lst 'daa), because of the quote. 
+   * that we can inline when we have (%cxr lst 'daa), because of the quote.
    */
   if (KEYWORDP(name)) {
     SCM lst   = l;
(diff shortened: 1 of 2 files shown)


jpellegrini:
@egallesio if this is OK, I can factor out the list traversal code and make it also work for `map`, `for-each`, `every`, `any`, etc.

jpellegrini:
PR #736 implements this, in a better way, and does much more... This can be closed.

egallesio:
Closed since PR  #736 is merged

Issue #721, opened 2025-02-17
GCC 15


jpellegrini:
Hi @egallesio !

This is just a reminder that STklos needs a few adjustments in order to be compiled with the (still not officially released) GCC 15:

```
make[2]: Entering directory '/home/jeronimo/pkg/scheme/STklos/src'
gcc-15 -DHAVE_CONFIG_H -I.   -DPCRE2POSIX_SHARED    -I/usr/local/include/stklos    -Wall -O2 -MT base64.o -MD -MP -MF .deps/base64.Tpo -c -o base64.o base64.c
In file included from base64.c:26:
./stklos.h:287:28: error: initialization of ‘void * (*)(void)’ from incompatible pointer type ‘void * (*)(void *, void *)’ [-Wincompatible-pointer-types]
  287 |         _sname, CPP_CONCAT(STk_, _cname), STk_nil};             \
      |                            ^~~~
./stklos.h:89:33: note: in definition of macro ‘CPP_CONCAT’
   89 | #define CPP_CONCAT(x, y)        x##y
      |                                 ^
base64.c:133:1: note: in expansion of macro ‘DEFINE_PRIMITIVE’
  133 | DEFINE_PRIMITIVE("base64-encode", base64_encode, subr12, (SCM f, SCM g))
      | ^~~~~~~~~~~~~~~~
./stklos.h:287:28: note: (near initialization for ‘STk_o_base64_encode.code’)
  287 |         _sname, CPP_CONCAT(STk_, _cname), STk_nil};             \
      |                            ^~~~
./stklos.h:89:33: note: in definition of macro ‘CPP_CONCAT’
   89 | #define CPP_CONCAT(x, y)        x##y
      |                                 ^
```

jpellegrini:
It looks like the problem is this:

In `stklos.h`:
```c
struct primitive_obj {
  stk_header header;
  char *name;
  SCM (*code)();
  SCM plist;
};
```

So the third field is `void * (*)(void)`, but the preprocessor expands this:

```
DEFINE_PRIMITIVE("base64-encode", base64_encode, subr12, (SCM f, SCM g))
```

into:

```c
SCM STk_base64_encode (SCM f, SCM g);       /* forward declaration */

struct primitive_obj STk_o_base64_encode = { 
  {tc_subr12, 0}, 
  "base64-encode", 
  STk_base64_encode, 
  ((SCM) (((SCM) ((unsigned long) (0 << 2 | 3)))))};  /* structure */

[...]

PR #722, opened 2025-02-18
Make STklos compile with GCC-15

jpellegrini:
I'm not actually sure about the first change. @egallesio what do you think?

* Use `void *code` instead of `SCM (*code)()` in `struct primitive_obj`, so GCC won't check types (it's just a void pointer now).

* Use `SCM` for the arguments to pthreads functions.

Passes all tests.

### DIFF
--- a/src/mutex-pthreads.c
+++ b/src/mutex-pthreads.c
@@ -57,7 +57,7 @@ void STk_make_sys_mutex(SCM z)
  *
  * Returns information about the state of the |mutex|. The possible results
  * are:
- * 
+ *
  *  * a thread *T*: the mutex is in the locked/owned state and
  *    thread *T* is the owner of the mutex
  *  * the symbol *not-owned*: the mutex is in the locked/not-owned
@@ -66,7 +66,7 @@ void STk_make_sys_mutex(SCM z)
  *    state
  *  * the symbol *not-abandoned*: the mutex is in the
  *     unlocked/not-abandoned state
- * 
+ *
  * @lisp
  * (mutex-state (make-mutex))  =>  not-abandoned
  *
@@ -121,7 +121,7 @@ DEFINE_PRIMITIVE("mutex-state", mutex_state, subr1, (SCM mtx))
  *    is not supplied),
  *  ** if T is terminated the mutex becomes _unlocked/abandoned_,
  *  ** otherwise mutex becomes _locked/owned_ with T as the owner.
- * 
+ *
  * After changing the state of the mutex, an "abandoned mutex exception" is
  * raised if the mutex was unlocked/abandoned before the state change,
  * otherwise |mutex-lock!| returns |#t|.
@@ -146,7 +146,7 @@ DEFINE_PRIMITIVE("%mutex-lock!", mutex_lock, subr3, (SCM mtx, SCM tm, SCM thread
     ts.tv_nsec = (suseconds_t) ((tmd - ts.tv_sec) * 1000000);
   }
 
-  pthread_cleanup_push((void (*)()) mutex_finalizer, mtx);
+  pthread_cleanup_push((void (*)(SCM)) mutex_finalizer, mtx);
 
   if (pthread_mutex_lock(&MUTEX_MYMUTEX(mtx)) != 0)
     STk_error_deadlock();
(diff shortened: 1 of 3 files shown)


jpellegrini:
Another solution (I'm not sure it would be good) would be to use a union of function pointers...

egallesio:
I have changed your fix and used a cast in the `DEFINE_PRIMITIVE` macro. The reason is that it is stated that a pointer on a function may use a different size than a pointer on a data. So, my understanding is that the field cannot be a `void*`. 

It seems the reason of the problem is that GCC-15 uses c17 (or more exactly gnu17) instead of gnu99. I don't have access to this version of the compiler, but it works when I use the `-std-c=gnu17` (and even `gnu23`). 
Can you test this commits, please, with your version of GCC?
Thank you.

jpellegrini:
Hi @egallesio ! Glad to see you're back!

So... It still doesn't compile:

```
In file included from string-obfuscate.c:77:
/usr/local/include/stklos/stklos.h:287:28: error: initialization of ‘void * (*)(void)’ from incompatible pointer type ‘void * (*)(void *, void *)’ [-Wincompatible-pointer-types]
  287 |         _sname, CPP_CONCAT(STk_, _cname), STk_nil};             \
      |                            ^~~~
/usr/local/include/stklos/stklos.h:89:33: note: in definition of macro ‘CPP_CONCAT’
   89 | #define CPP_CONCAT(x, y)        x##y
      |                                 ^
string-obfuscate.c:98:1: note: in expansion of macro ‘DEFINE_PRIMITIVE’
   98 | DEFINE_PRIMITIVE("string-add", obf_add, subr2, (SCM str, SCM n))
      | ^~~~~~~~~~~~~~~~
/usr/local/include/stklos/stklos.h:287:28: note: (near initialization for ‘STk_o_obf_add.code’)
  287 |         _sname, CPP_CONCAT(STk_, _cname), STk_nil};             \
      |                            ^~~~
/usr/local/include/stklos/stklos.h:89:33: note: in definition of macro ‘CPP_CONCAT’
   89 | #define CPP_CONCAT(x, y)        x##y
      |                                 ^
string-obfuscate.c:98:1: note: in expansion of macro ‘DEFINE_PRIMITIVE’
   98 | DEFINE_PRIMITIVE("string-add", obf_add, subr2, (SCM str, SCM n))
      | ^~~~~~~~~~~~~~~~
/usr/local/include/stklos/stklos.h:284:18: note: ‘STk_obf_add’ declared here
  284 |   SCM CPP_CONCAT(STk_, _cname) _params;                         \
      |                  ^~~~
/usr/local/include/stklos/stklos.h:89:33: note: in definition of macro ‘CPP_CONCAT’
   89 | #define CPP_CONCAT(x, y)        x##y
      |                                 ^
string-obfuscate.c:98:1: note: in expansion of macro ‘DEFINE_PRIMITIVE’
   98 | DEFINE_PRIMITIVE("string-add", obf_add, subr2, (SCM str, SCM n))
      | ^~~~~~~~~~~~~~~~
```

jpellegrini:
> So... It still doesn't compile:

I mean, with GCC-15. With GCC 14 it compiles.

egallesio:
Hi @jpellegrini ,

Are you sure you have the last version of the source? I don't see the cast in front of the `CPP_CONCAT`: 

>  /usr/local/include/stklos/stklos.h:287:28: error: initialization of ‘void * (*)(void)’ from incompatible pointer type ‘void * (*)(void *, void *)’ [-Wincompatible-pointer-types]
>  287 |         _sname, CPP_CONCAT(STk_, _cname), STk_nil};             \
 
Otherwise, I have tried to compile the following program 

```c
#include <stdio.h>

typedef void (*ptrfunc)();

void f1() { printf("F1\n"); }

void f2(int a) { printf("F2 %d\n", a); }

void f3(int a, int b, int c) { printf("F3 %d\n", a+b+c); }


int main(int argc, char *argv[]) {
  ptrfunc p1 = (ptrfunc) f1;
  ptrfunc p2 = (ptrfunc) f2;
  ptrfunc p3 = (ptrfunc)f3;
  return 0;
}
```

If I compile (with GCC14) this program it works for all version C90 C17 and C23.

```sh
$ gcc -std=c90 foo.c && echo OK
OK
$ gcc -std=c17 foo.c && echo OK
OK
$ gcc -std=c23 foo.c && echo OK
OK
```

However, if I delete the casts, I have an error only with c23:

```sh
$ gcc -std=c23 foo.c && echo OK

foo.c:14:17: error: initialization of ‘ptrfunc’ {aka ‘void (*)(void)’} from incompatible pointer type ‘void (*)(int)’ [-Wincompatible-pointer-types]
   14 |   ptrfunc p2 =  f2;
      |                 ^~
foo.c:15:17: error: initialization of ‘ptrfunc’ {aka ‘void (*)(void)’} from incompatible pointer type ‘void (*)(int,  int,  int)’ [-Wincompatible-pointer-types]
   15 |   ptrfunc p3 =  f3;
      |                 ^~
```

Can you test, please with GCC15?

jpellegrini:
Oops. Sorry, it does indeed compile correctly!

egallesio:
Great. We can close the corresponding issue (Issue #721)?

jpellegrini:
> Great. We can close the corresponding issue (Issue #721)?

Done! :)

PR #724, opened 2025-03-06
Optimize `make-list` and `vector->list`

jpellegrini:
Instead of going through the vector and consing, we allocate all list cells in a single memory region, in a single call to the allocator. This seems to be twice as fast as the old code.

```scheme
(time
  (let ((a (make-vector 1000 'x)))
    (repeat 100000 (vector->list a))
    'ok))
```
Old code: 4152.755ms 100000001 allocation calls
New code: 1805.502ms 100001 allocation calls

But this only changes the R5RS version of `vector->lsit`. I've tried to make the R7RS version in C, but that is trickier than what I thought. I'll work on that later.

### DIFF
--- a/lib/r7rs.stk
+++ b/lib/r7rs.stk
@@ -34,7 +34,7 @@
         truncate-remainder floor-remainder
         square exact-integer-sqrt exact inexact
         boolean=?
-        make-list member assoc
+        member assoc
         symbol=?
         string=?     %string2=?
         string<?     %string2<?
@@ -538,23 +538,6 @@ doc>
                         (eq? (car lst) val)
                         (verify val (cdr lst)))))))
     (verify e1 rest)))
-
-;;;; ----------------------------------------------------------------------
-;;;; 6.4 Pairs and lists
-;;;; ----------------------------------------------------------------------
-
-#|
-<doc R7RS make-list
- * (make-list k)
- * (make-list k fill)
- *
- * Returns a newly allocated list of k elements. If a second
- * argument is given, then each element is initialized to fill .
- * Otherwise the initial contents of each element is unspecified.
-doc>
-|#
-(define (make-list k :optional (fill (void)))
-  (vector->list (make-vector k fill)))
 
 
 ;;;; ----------------------------------------------------------------------
--- a/src/list.c
+++ b/src/list.c
@@ -726,9 +726,138 @@ DEFINE_PRIMITIVE("assoc", assoc, subr23, (SCM obj, SCM alist, SCM cmp))
  * the |car| and |cdr| of |obj|, respectively.
 doc>
  */
-DEFINE_PRIMITIVE("list-copy", list_copy, subr1, (SCM l))
-{
-  return CONSP(l) ? STk_cons(STk_list_copy(CAR(l)), STk_list_copy(CDR(l))): l;
+DEFINE_PRIMITIVE("list-copy", list_copy, subr1, (SCM l)) {
+  /*
+    About the cycle detecting and copying algorithm:
+
+    1. Detect cycles using Floyd's algorithm.  The algorithm runs two
+       pointers, "fast" and "slow" through the list. Both are placed
+       at the CAR, then at each step, slow goes forward one link, and
+       fast goes forward two links.
+
+    2. If there is a cycle:
+
+       2.i) we position fast back at the CAR and now move the two
+            pointers *one* link at each time. The pointers will
+            *necessarily meet* exactly *at the beginning of the cycle.
+
+       2.ii) If there is a cycle, we mark the cell in which it starts
+             example, the cycle below starts at C:
+
+                        +--------------+
+                        |              |
+                        v              |
+              A -> B -> C -> D -> E -> F
+
+              We keep an extra pointer to C.
+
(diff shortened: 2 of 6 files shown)


jpellegrini:
Also, it's probably nice to change `vector->string` in the same way.

jpellegrini:
This has an impact on `make-list`:

```scheme
(time
  (let ((a #f))
    (repeat 100000
      (set! a (make-list 1000)))
    'ok))
```

Old code: 5339.514 ms  100100000 allocation calls
New code: 2841.315 ms 200000 allocation calls

Now make 10x less lists, with 10x more elements per list:

```scheme
(time
  (let ((a #f))
    (repeat 10000
      (set! a (make-list 10000)))
    'ok))
```

Old code: 4928.766 ms 100010000 allocation calls
New code: 2097.78 ms 20000 allocation calls

jpellegrini:
> Also, it's probably nice to change `vector->string` in the same way.

Obviously not, I don't know what I was thinking. Of course, strings are not built by "consing characters". `vector->string` *already* works allocating all space at once.

jpellegrini:
But `string->list` can be optimized (see the second commit). Large speedup also.

egallesio:
> Instead of going through the vector and consing, we allocate all list cells in a single memory region, in a single call to the allocator. This seems to be twice as fast as the old code.

The problem with this implementation is that the list forms a single block for the GC. 
In the following code,
```scheme
stklos> (define v (make-vector 1000 'x))
;; v
stklos> (define l (vector->list v))
;; l
stklos> (set! l (last-pair l))
stklos> (gc)
```
the 999 first pairs will be GC'ed. With your solution, this will not be the case, since `l` points in the region allocated with `vector->list`.

jpellegrini:
> the 999 first pairs will be GC'ed. With your solution, this will not be the case, since `l` points in the region allocated with `vector->list`.

Hmm, you're right.  If libgc has some way of allocating a large region at once, but conceptually marking it as a sequence of cells, it would work. I'll see if this is the case.

egallesio:
>If libgc has some way of allocating a large region at once, but conceptually marking it as a sequence of cells, it would work.

Yep, and it would be great. (but, I doubt it will be easy).

jpellegrini:
https://github.com/ivmai/bdwgc/issues/713
:)

egallesio:
It would be great if they implement it (but I suppose that it will not be simple  with alignment problems).
Great idea to ask directly in GH.

jpellegrini:
@egallesio ...

Ivan Maidanski answered. We can use `GC_generic_malloc_many` from libgc! :)

Of course, the right way to do it is to create a new STklos macro... I'll work on that this week!

FYI, here's the documentation:

```c
/* Return a list of 1 or more objects of the indicated size, linked     */
/* through the first word in the object.  This has the advantage that   */
/* it acquires the allocation lock only once, and may greatly reduce    */
/* time wasted contending for the allocation lock.  Typical usage would */
/* be in a thread that requires many items of the same size.  It would  */
/* keep its own free list in thread-local storage, and call             */
/* GC_malloc_many or friends to replenish it.  (We do not round up      */
/* object sizes, since a call indicates the intention to consume many   */
/* objects of exactly this size.)                                       */
/* We assume that the size is non-zero and a multiple of                */
/* GRANULE_BYTES, and that it already includes EXTRA_BYTES value.       */
/* We return the free-list by assigning it to *result, since it is      */
/* not safe to return, e.g. a linked list of pointer-free objects,      */
/* since the collector would not retain the entire list if it were      */
/* invoked just as we were returning.                                   */
/* Note that the client should usually clear the link field.            */
GC_API void GC_CALL GC_generic_malloc_many(size_t lb, int k, void **result)
```

egallesio:
> Ivan Maidanski answered. We can use `GC_generic_malloc_many` from libgc! :)
> 
> Of course, the right way to do it is to create a new STklos macro... I'll work on that this week!

That's really great.

jpellegrini:
@egallesio -- `string->list ` would allow for similar optimization, but I'd like to validate this with you first.

jpellegrini:
There is another change that is possible: 
1. make STklos use the C function `STk_hash_table_keys`, which already exists, instead of the closure in `bonus.stk`
2. make `STk_hash_table_keys` allocate the list with the new `STk_make_list` primitive.

Then the speedup is like 5x. But some STklos code seems to rely on the order of hash table keys returned, so I won't include this here yet.

jpellegrini:
Summary:
* The PR works fine now
* `string->list` could use the same change as `vector->list`
* `uXXvector->list` too (from the `(scheme vector @)` library)
* `hash-table-map` and `hash-table-keys`  can be enhanced, but some internal STklos code would need to be adjusted/fixed

jpellegrini:
I'll reorganize this into another PR later. I'll leave it here as a draft.  My idea is to refactor all of this.

jpellegrini:
PR #736 implements this, in a better way, and does much more... This can be closed.

egallesio:
Very impressive! :open_mouth: 
Looking forward to testing it. 
I'll close this PR and the  #736 at the same time.
Thanks a lot @jpellegrini.

jpellegrini:
@egallesio  see PR #736

PR #727, opened 2025-03-07
[trivial] Document the implementation of `DEFINE_PRIMITIVE`

jpellegrini:


jpellegrini:
Just a C comment, but it's such a central macro in STklos that I thought it would be nice to document it.

egallesio:
Merged  (and deleted the previous comment, because yes, there was already a comment on `DEFINE_PRIMITIVE` :smile: )

PR #729, opened 2025-03-10
[simple] Be more informative about the 'bad closing parenthesis'

jpellegrini:
* We now tell the user what closing delimiter was found -- `(` `[` or `{`.
* If we were reading a list then we also inform what closing delimiter was expected instead.

Examples:

```
stklos> '[a)

**** Warning:
bad closing parenthesis `)` (expected `]`) on line 1 of '#[input-virtual-port 7ff782c93840]'
(a)

stklos> 123 ]
123

**** Warning:
bad closing parenthesis `]` on line 3 of '#[input-virtual-port 7f5fa526b840]'
```

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -167,10 +167,16 @@ static void error_eof_in_string(SCM port)
   signal_error(port, "end of file while reading a string on ~S", port);
 }
 
-
-static void warning_parenthesis(SCM port)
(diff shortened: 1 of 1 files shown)


egallesio:
Merged. Thanks @jpellegrini

PR #734, opened 2025-03-15
Fold reduce

jpellegrini:
@egallesio --

This PR adds `fold` and `reduce` to the core. This is why:

- They are redefined (in `logical.stk` and SRFI-224 at least, maybe in other places);
- Some code crashes with the version of reduce that they have (`logical.stk` for example);
- It's not uncommon for Scheme implementations to have these, although not common either;
- Well, they're fundamental, and having them written in C gives us a big speedup.

See the commit messages for more.

For reference, a table of Scheme implementations and their support for `fold` and `reduce`:

| Implementation | fold | fold-right | reduce | reduce-right |
|----------------|------|------------|--------|--------------|
| Bigloo         | no   | no         | yes    | no           |
| Chibi          | no   | no         | no     | no           |
| Chez           | no   | no         | no     | no           |
| Chicken        | no   | no         | no     | no           |
| Cyclone        | no   | no         | no     | no           |
| Gambit         | yes  | yes        | no     | no           |
| Gauche         | yes  | yes        | yes    | yes          |
| Guile          | no   | no         | no     | no           |
| Kawa           | no   | no         | no     | no           |
| MIT            | yes  | yes        | yes    | yes          |
| Sagittarius    | yes  | yes        | yes    | no           |
| SCM            | no   | no         | no     | no           |
| STklos         | no   | no         | no     | no           |
| Ypsilon        | no   | no         | no     | no           |

If it's OK, I can implement `reduce-right` and `fold-right` later.


jpellegrini:
And it passes all tests. Since we changed SRFI-1 to use core `fold` and `reduce`, then they pass all tests of SRFI-1. :smile:

PR #736, opened 2025-03-16
Optimize `list.c`

jpellegrini:
Hi @egallesio !
These are several enhancements to `list.c` and related functions in `proc.c`.

Passes all tests. :)

And there are some really big performance gains, besides some enhancements in argument checking and error messages. (This brings the new allocator for lists, which does not repeatedly cons, but asks for a large set of cells each time)

# Optimize make-list and vector->list

Allocate more cells at once in `make-list`.

`libgc` only offers an interface to "get several objects", without telling us how many. It's usually a not very large number like 128 objects for example). We give the size, and receive a list.

This patch does this:

1. Implements `SCM STk_must_malloc_list(int n, SCM init)` that will allocate a list of CONS cells.
2. Changes `vector->list` so it will use the new function
3. Removes `make-list` and use the new function instead

There is also an internal function `STk_must_malloc_list_upton`, which allocates a list of size "at most n". This function will receive a size argument `n`, and will call the `libgc` function to allocate many objects.
1. If there is more than `n`, it will run through the list, cleaning the pointers so they are released.
2. If there is less than `n`, it will return the number of missing cells (not allocated) in the parameter `n`

`STk_must_malloc_list`  does:
1. calls `STk_must_malloc_list_upton`
3. If there are enough cells, stop
4. If not, call  `STk_must_malloc_list_upton` again, and append the two lists. Appending is NOT linear, it is O(1) here, because `STk_must_malloc_list_upton` returns a pointer to the last cell :)

New timings:
```scheme
(time
  (let ((a (make-vector 1000 'x)))
    (repeat 100000 (vector->list a))
    'ok))
```
Old code: 4152.755ms 100000001 allocation calls
New code: 2552.173ms  934202 allocation calls

```scheme
(time
  (let ((a #f))
    (repeat 100000
      (set! a (make-list 1000)))
    'ok))
```
Old code: 5339.514 ms 100100000 allocation calls
New code:  2249.643 ms 935418 allocation calls

Now make 10x less lists, with 10x more elements per list:
```scheme
(time
  (let ((a #f))
    (repeat 10000
      (set! a (make-list 10000)))
    'ok))
```
Old code: 4928.766 ms 100010000 allocation calls
New code: 2706.778 ms 873160 allocation calls

Also -- one more thing.

The new code properly counts allocation.

The size of `struct cons_object` on my machine is `24`.
```c

typedef struct {
  int16_t type, cell_info;
} stk_header;

struct cons_obj {
  stk_header header;
  void * car;
  void* cdr;
};
int main() {
	printf("size is %ld\n", sizeof(struct cons_obj));
}
```
The above program prints 24.

So -- with the new code,

```scheme
(time (make-list 10))
;    Elapsed time: 0.017 ms
;    Allocations: 240 bytes in 1 allocation call (GC: 0)

(time (make-list 10000))
;    Elapsed time: 0.131 ms
;    Allocations: 240000 bytes in 116 allocation calls (GC: 0)
```
**Exactly** the requested number of bytes (ok, we're cheating, because we allocated more and threw hem away, but they're gone immediately after allocation).

The old code, for some reason, was allocating more than necessary:
```scheme
(time (make-list 10))
;    Elapsed time: 0.007 ms
;    Allocations: 328 bytes in 11 allocation calls (GC: 0)

(time (make-list 10000))
;    Elapsed time: 0.955 ms
;    Allocations: 320008 bytes in 10001 allocation calls (GC: 0)
```
# Optimize MAP

Allocate the result list at once instead of consing all of the result list.

```scheme
(import (scheme list))

(let ((L (iota 30000000)))
  (time (begin (map - L)
               -1)))
```
Old code: 1728.254 ms
New code: 1131.081 ms

```scheme
(let ((L (iota 20000000))
      (M (reverse (map - (iota 18000000)))))
  (time (begin (map * L M)
               -1)))
```
Old code: 1273.321 ms
New code: 1021.081 ms

# Optimize and enhance list-copy, add list-deep-copy

1. Use Floyd's algorithm to detect cycles. It now copies:
   - proper lists
   - improper lists
   - cyclic lists
2. Allocate the list at once, not cell-by-cell
3. Do not use recursion. This avoids crashes with large lists
4. For all this to work, we implement a new C function,  `STk_list_type_and_length()`, which
   - tells wether the object is a proper list, improper list,  cyclic list or no list at all;
   - calculates the length of the list (if it is one)
5. Since we did all that, move circular-list? from SRFI-1  to the core
6. Add list-deep-copy. This is necessary in the module import  code. The previous version of `list-copy` would recurse    into the list, and the new won't. Reasons for making it    different, and create the separate `list-deep-copy`:
   - Every Scheme except STklos behaves the same way (they  won't recurse into the list); Gauche even has a  specialized `alist-copy`.
   - The `list-copy` primitiveis much faster than the old  one

See comments in the code.

```scheme
(let ((L (iota 20000)))
  (time (begin (repeat 3000 (list-copy L))
               -1)))
```
Old code: 2157.288 ms
New code: 1590.699 ms

```scheme
(let ((L (iota 20000000)))
  (begin (list-copy L)
          -1))
```
Old code crashed
New code works fine (The C function doesn't do recursive calls).

# Optimize and enhance reverse

`reverse` is a bit faster now, since it calls list-copy to allocate cells all at once.

And now STklos tells when a list is circular or improper:

stklos> (reverse '(1 . 2))
**** Error:
reverse: improper list '(1 . 2)'

stklos> (reverse (circular-list 1 2 3))
**** Error:
reverse: list '#0=(1 2 3 . #0#)' is circular

stklos> (reverse (append '(11 22) (circular-list 1 2 3)))
**** Error:
reverse: list '(11 22 . #0=(1 2 3 . #0#))' is circular

The two last examples would make STklos enter an infinite loop before this change.

#  Optimize and enhance `last-pair`

There was already a `last-pair` primitive (?) in STklos, but there was also a closure overriding it.
We enhanced the primitive a bit, and made it be used by default.

# Enhance `filter`

It now complains more clearly about circular and improper lists.

### DIFF
--- a/lib/module.stk
+++ b/lib/module.stk
@@ -23,6 +23,17 @@
 ;;;;    Creation date:  1-Jun-2000 12:26 (eg)
 ;;;;
 
+;;; NOTE:
+;;;
+;;;
+;;; This code makes use of `list-deep-copy` because import sets, which are
+;;; alists, are copied. If one uses `list-copy`, then the assoc pairs are
+;;; not newly allocated, and are shared with the old import set. This will
+;;; cause problems, since this code uses set-cdr! to change the name of
+;;; a binding.
+;;;
+;;; So: *** use list-deep-copy, not list-copy! ***
+
 ;;=============================================================================
 ;;
 ;;                              MODULE-SYMBOLS*
--- a/lib/module.stk
+++ b/lib/module.stk
@@ -431,7 +431,7 @@ doc>
     (let* ((mod       (eventually-load-library name))
            (imported  (build-import-list mod
                                          keylist
-                                         (list-copy (%module-exports mod)))))
+                                         (list-deep-copy (%module-exports mod)))))
       ;; Make aliases for all symbols in imported list
       (if only-syntax
           (for-each (lambda (pair)
@@ -476,10 +476,10 @@ doc>
                 (%grab-file-information (symbol->string (car x))))
               imp)
     ;; Do importation at compile time AND at execution time
-    (%do-imports (compiler-current-module) (list-copy imp) #t)
+    (%do-imports (compiler-current-module) (list-deep-copy imp) #t)
 
     ;; Run time expansion
-    `(%do-imports (current-module) (list-copy ',imp) #f)))
+    `(%do-imports (current-module) (list-deep-copy ',imp) #f)))
 
 (define (%%import lib modules)
   (let ((imp (%parse-imports modules)))
@@ -490,7 +490,7 @@ doc>
               imp)
     ;; Do importation at compile time AND at execution time
     ;(%do-imports  (compiler-current-module) (list-copy imp) #t)
-    (%do-imports lib (list-copy imp) #f)))
+    (%do-imports lib (list-deep-copy imp) #f)))
 
 
 ;;=============================================================================
--- a/lib/r7rs.stk
+++ b/lib/r7rs.stk
@@ -34,7 +34,7 @@
         truncate-remainder floor-remainder
         square exact-integer-sqrt exact inexact
         boolean=?
-        make-list member assoc
+        member assoc
         symbol=?
         string=?     %string2=?
         string<?     %string2<?
@@ -538,23 +538,6 @@ doc>
                         (eq? (car lst) val)
                         (verify val (cdr lst)))))))
     (verify e1 rest)))
-
-;;;; ----------------------------------------------------------------------
-;;;; 6.4 Pairs and lists
-;;;; ----------------------------------------------------------------------
-
-#|
-<doc R7RS make-list
- * (make-list k)
- * (make-list k fill)
- *
- * Returns a newly allocated list of k elements. If a second
- * argument is given, then each element is initialized to fill .
- * Otherwise the initial contents of each element is unspecified.
-doc>
-|#
-(define (make-list k :optional (fill (void)))
-  (vector->list (make-vector k fill)))
 
 
 ;;;; ----------------------------------------------------------------------
--- a/lib/scheme/list.stk
+++ b/lib/scheme/list.stk
@@ -467,14 +467,15 @@
           (not (null? x))))
     (not (null? x)))))
 
-(define (circular-list? x)
-  (let lp ((x x) (lag x))
-    (and (pair? x)
-     (let ((x (cdr x)))
-       (and (pair? x)
-        (let ((x   (cdr x))
-              (lag (cdr lag)))
-          (or (eq? x lag) (lp x lag))))))))
+;; In STklos core
+;; (define (circular-list? x)
+;;   (let lp ((x x) (lag x))
+;;     (and (pair? x)
+;;      (let ((x (cdr x)))
+;;        (and (pair? x)
+;;         (let ((x   (cdr x))
+;;               (lag (cdr lag)))
+;;           (or (eq? x lag) (lp x lag))))))))
 
 (define (not-pair? x) (not (pair? x)))  ; Inline me.
 
@@ -1769,6 +1770,9 @@ FILTER an FILTER! are primitives in STklos
 (redefine make-list)
 (redefine assoc)
 (redefine member)
+
+;; Not R7RS, but also moved to STklos core:
+(redefine circular-list?)
 
 )  ; END OF MODULE
 
--- a/lib/scheme/list.stk
+++ b/lib/scheme/list.stk
@@ -694,11 +694,12 @@
 
 (define (last lis) (car (last-pair lis)))
 
-(define (last-pair lis)
-  (check-arg pair? lis last-pair)
-  (let lp ((lis lis))
-    (let ((tail (cdr lis)))
-      (if (pair? tail) (lp tail) lis))))
+;; Already in STklos core
+;; (define (last-pair lis)
+;;   (check-arg pair? lis last-pair)
+;;   (let lp ((lis lis))
+;;     (let ((tail (cdr lis)))
+;;       (if (pair? tail) (lp tail) lis))))
 
 
 ;;; Unzippers -- 1 through 5
@@ -1772,6 +1773,7 @@ FILTER an FILTER! are primitives in STklos
 (redefine member)
 
 ;; Not R7RS, but also moved to STklos core:
+(redefine last-pair)
 (redefine circular-list?)
 
 )  ; END OF MODULE
--- a/lib/scheme/vector/tagvector-template.stk
+++ b/lib/scheme/vector/tagvector-template.stk
@@ -403,12 +403,14 @@
 
 
 (define (TAGvector->list v :optional (start 0) (end (TAGvector-length v)))
-  (let ((v (TAGvector-copy v start end)))
-    (let loop ((i (fx- (TAGvector-length v) 1))
-               (lst '()))
-      (if (fx< i 0)
-          lst
-          (loop (fx- i 1) (cons (TAGvector-ref v i) lst))))))
+  (let* ((len (fx- end start))
+         (L (make-list len))
+         (ptr L))
+    (repeat len
+      (set-car! ptr (TAGvector-ref v start))
+      (set! ptr (cdr ptr))
+      (set! start (fx+ 1 start)))
+    L))
 
 
 (define (reverse-list->TAGvector lst)
--- a/src/list.c
+++ b/src/list.c
@@ -1076,7 +1076,10 @@ doc>
  */
 DEFINE_PRIMITIVE("last-pair", last_pair, subr1, (SCM l))
 {
-  if (!CONSP(l)) error_wrong_type(l);
+  int len;
+  SCM x = STk_list_type_and_length(l, &len);
+  if (!x) error_bad_list(l);
+  if (CONSP(x)) error_circular_list(l);
 
   while (CONSP(CDR(l)))
     l = CDR(l);
--- a/src/list.c
+++ b/src/list.c
@@ -47,6 +47,11 @@ static void error_bad_list(SCM x)
   STk_error("bad list ~W", x);
 }
 
+static void error_improper_list(SCM x)
+{
+  STk_error("improper list ~W", x);
+}
+
 static void error_bad_proc(SCM x)
 {
   STk_error("bad procedure ~S", x);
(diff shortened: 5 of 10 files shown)


jpellegrini:
By the way... I know that R7RS says "it is an error" if the list passed to `list-copy` is circular. But it doesn't require us to signal an error, and we can do whatever we think is best... :smile:

jpellegrini:
This one is ready...

jpellegrini:
I'll probably do more enhancements to the way libgc is used... In the near future. But this is already some improvement, no? :)

egallesio:
**Long message below**

Still working on this PR.

In fact this version is fast for big list, but very VERY slow on small lists. and we often manipulate very small lists.

Some statistics about calls to `STk_must_malloc_list`. When launching `stklos` From start to prompt: 10190 calls to the function
 - 2 calls with `n`=0
 - 10086 calls with `n`=1
 - 57 calls with `n`=2
 - 17 calls with `n`=3
 - 52 calls with another value of `n` (longest list is 1208)

Some tests with `STk_must_malloc_list` with two versions
	- `cons` which use a trivial function which loops and conses `n` cells (a C version of `make-list`).
	- `many` which uses `GC_malloc_many`
```sh
***** 1 ******
cons 1
;    Elapsed time: 70.391 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)
many 1
;    Elapsed time: 2975.657 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 1432)

****** 10 ******
cons 10
;    Elapsed time: 491.574 ms
;    Allocations: 240000000 bytes in 10000000 allocation calls (GC: 125)
many 10
;    Elapsed time: 3047.067 ms
;    Allocations: 240000000 bytes in 1000000 allocation calls (GC: 1432)

****** 50 ******
cons 50
;    Elapsed time: 2282.107 ms
;    Allocations: 1200000000 bytes in 50000000 allocation calls (GC: 626)
many 50
;    Elapsed time: 3191.236 ms
;    Allocations: 1200000000 bytes in 1000000 allocation calls (GC: 1485)

****** 70 ******
cons 70
;    Elapsed time: 3018.662 ms
;    Allocations: 1680000000 bytes in 70000000 allocation calls (GC: 877)
many 70
;    Elapsed time: 3326.882 ms
;    Allocations: 1680000000 bytes in 1000000 allocation calls (GC: 1582)

****** 100 ******
cons 100
;    Elapsed time: 4296.549 ms
;    Allocations: 2400000000 bytes in 100000000 allocation calls (GC: 1253)
many 100
;    Elapsed time: 3437.963 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1618)

****** 300 ******
cons 300
;    Elapsed time: 12758.79 ms
;    Allocations: 7200000000 bytes in 300000000 allocation calls (GC: 3759)
many 300
;    Elapsed time: 9997.168 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 4728)
```

What we see: when `n` is small `(make-list n)` allocate too much useless cells and the GC is called more often. The situation seems better when `n > 100`. (At least on my machine)

So we change  `STk_must_malloc_list` to use
 - an exact consing algorithm for small lists and
 - the `GC_malloc_many` otherwise.
 We have:
 
```sh
****** 1 ******
cons 1
;    Elapsed time: 69.847 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)
many 1
;    Elapsed time: 2948.124 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 1434)
cons+many 1
;    Elapsed time: 71.363 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)

****** 10 ******
cons 10
;    Elapsed time: 531.227 ms
;    Allocations: 240000000 bytes in 10000000 allocation calls (GC: 126)
many 10
;    Elapsed time: 3175.606 ms
;    Allocations: 240000000 bytes in 1000000 allocation calls (GC: 1436)
cons+many 10
;    Elapsed time: 508.629 ms
;    Allocations: 240000000 bytes in 10000000 allocation calls (GC: 126)

****** 50 ******
cons 50
;    Elapsed time: 2277.293 ms
;    Allocations: 1200000000 bytes in 50000000 allocation calls (GC: 632)
many 50
;    Elapsed time: 3241.686 ms
;    Allocations: 1200000000 bytes in 1000000 allocation calls (GC: 1510)
cons+many 50
;    Elapsed time: 2188.101 ms
;    Allocations: 1200000000 bytes in 50000000 allocation calls (GC: 632)

****** 70 ******
cons 70
;    Elapsed time: 3038.674 ms
;    Allocations: 1680000000 bytes in 70000000 allocation calls (GC: 884)
many 70
;    Elapsed time: 3342.847 ms
;    Allocations: 1680000000 bytes in 1000000 allocation calls (GC: 1591)
cons+many 70
;    Elapsed time: 3109.275 ms
;    Allocations: 1680000000 bytes in 70000000 allocation calls (GC: 883)

****** 100 ******
cons 100
;    Elapsed time: 4371.237 ms
;    Allocations: 2400000000 bytes in 100000000 allocation calls (GC: 1264)
many 100
;    Elapsed time: 3578.991 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1631)
cons+many 100
;    Elapsed time: 3572.666 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1628)

****** 300 ******
cons 300
;    Elapsed time: 13769.275 ms
;    Allocations: 7200000000 bytes in 300000000 allocation calls (GC: 3802)
many 300
;    Elapsed time: 10643.509 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 4777)
cons+many 300
;    Elapsed time: 10267.207 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 4766)
```

We see that `cons+many` s always faster for all value of `n`, but it is the cause of more 
GC calls. We can do better. Keep the list of the (non used) allocated objects by `GC_malloc_many` before returning from `STk_must_malloc_list` and use it later for a later call to the function. We have

```sh
****** 1 ******
cons 1
;    Elapsed time: 76.289 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)
many 1
;    Elapsed time: 2971.893 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 1436)
cons+many 1
;    Elapsed time: 75.979 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)
pool 1
;    Elapsed time: 80.53 ms
;    Allocations: 24000000 bytes in 1000000 allocation calls (GC: 12)

****** 10 ******
cons 10
;    Elapsed time: 478.399 ms
;    Allocations: 240000000 bytes in 10000000 allocation calls (GC: 127)
many 10
;    Elapsed time: 2993.582 ms
;    Allocations: 240000000 bytes in 1000000 allocation calls (GC: 1438)
cons+many 10
;    Elapsed time: 483.016 ms
;    Allocations: 240000000 bytes in 10000000 allocation calls (GC: 127)
pool 10
;    Elapsed time: 333.861 ms
;    Allocations: 240000000 bytes in 1000000 allocation calls (GC: 127)

****** 50 ******
cons 50
;    Elapsed time: 2199.064 ms
;    Allocations: 1200000000 bytes in 50000000 allocation calls (GC: 635)
many 50
;    Elapsed time: 3233.748 ms
;    Allocations: 1200000000 bytes in 1000000 allocation calls (GC: 1523)
cons+many 50
;    Elapsed time: 2209.382 ms
;    Allocations: 1200000000 bytes in 50000000 allocation calls (GC: 635)
pool 50
;    Elapsed time: 1426.691 ms
;    Allocations: 1200000000 bytes in 1000000 allocation calls (GC: 636)

****** 70 ******
cons 70
;    Elapsed time: 3046.696 ms
;    Allocations: 1680000000 bytes in 70000000 allocation calls (GC: 888)
many 70
;    Elapsed time: 3403.031 ms
;    Allocations: 1680000000 bytes in 1000000 allocation calls (GC: 1602)
cons+many 70
;    Elapsed time: 3061.648 ms
;    Allocations: 1680000000 bytes in 70000000 allocation calls (GC: 890)
pool 70
;    Elapsed time: 1984.517 ms
;    Allocations: 1680000000 bytes in 1000000 allocation calls (GC: 890)

****** 100 ******
cons 100
;    Elapsed time: 4331.813 ms
;    Allocations: 2400000000 bytes in 100000000 allocation calls (GC: 1270)
many 100
;    Elapsed time: 3539.567 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1650)
cons+many 100
;    Elapsed time: 3529.448 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1647)
pool 100
;    Elapsed time: 2797.453 ms
;    Allocations: 2400000000 bytes in 1000000 allocation calls (GC: 1272)

****** 300 ******
cons 300
;    Elapsed time: 12901.767 ms
;    Allocations: 7200000000 bytes in 300000000 allocation calls (GC: 3816)
many 300
;    Elapsed time: 10267.843 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 4807)
cons+many 300
;    Elapsed time: 10294.343 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 4807)
pool 300
;    Elapsed time: 8288.592 ms
;    Allocations: 7200000000 bytes in 1000000 allocation calls (GC: 3824)
```

Here we have the same number of GC and we are always faster. 
The function just need to use a mutex to be thread-safe, but is is still small mount of code (~ 30lines).

jpellegrini:
> Keep the list of the (non used) allocated objects by GC_malloc_many before returning from STk_must_malloc_list and use it later for a later call to the function.

That was one of the ideas I was thinking of for the future... :smile:  Great to know you've implemented it!

jpellegrini:
> The function just need to use a mutex to be thread-safe, but is is still small mount of code (~ 30lines).

Regarding that, Ivan Maidanski says, about the libgc function we're using:

> please note that by default thread-local allocation is used (thus the lock is not acquired on every call of GC_malloc). Also, you could consider functionality provided by `gc_inline.h`.

(https://github.com/ivmai/bdwgc/issues/713)

egallesio:
Hi @jpellegrini,

I have finally finished the merge of this PR. It is  a great one. 
Thanks again.

jpellegrini:
Ran the test suite 10 times, and the average run time decreased something like 5%. :partying_face:

PR #737, opened 2025-03-16
[simple] Add `primitive?` (for completeness, since we have `procedure?` and `closure?`)

jpellegrini:
One of them could have been added to `bonus.stk`, written in Scheme, but... They we'd have the two twin functions in different places (and written in different languages)...


jpellegrini:
And this is funny:
```scheme
(primitive? primitive?) => #t
(primitive? closure?)    => #t

(closure? primitive?)    => #f
(closure? closure?)       => #f
```

PR #738, opened 2025-03-18
Document `autoload`

jpellegrini:
Hi @egallesio !
When we compile STklos, it complains that one last procedure has documentation that has not been inserted in the database... I'm not sure where it would go into the manual, but...  Here's some initial version of the doc. :)

### DIFF
--- a/lib/load.stk
+++ b/lib/load.stk
@@ -560,8 +560,59 @@ doc>
 <doc EXT autoload
  * (autoload file symbol ...)
  *
- * TODO
+ * This macro will arrange so that when |symbol| is referenced, |file| is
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @jpellegrini,

I have finally done some work on the front of documenting autoload. This is not yet what I wanted because you cannot autoload syntax :disappointed:. Anyway, the first parameter can now be a library specification (which loads AND imports). So, I have in my `stklosrc` file 

```scheme
(autoload (scheme list) iota)
```
and I now have access to your C function, instead of my old Scheme function, which was defined in this `stklosrc` file.

:sunglasses:

PR #740, opened 2025-03-19
[R7RS] Implement standard `case-lambda` from R7RS

jpellegrini:
I know that the one in `bonus.stk` works... But using the exact code from the standard would be nice. :)
And with the current `syntax-rules` implementation, it does work, except for the test for hygiene (but the previous version of `case-lambda` also failed this one).

I have implemented some more tests.


egallesio:
Hi @jpellegrini,


I have not completely merged this PR. The reason are that the syntax-case code
   1. uses syntax-case in the expansion and is less efficient, that the actual
      version
   2. produces a less readable expansion (imho)


For instance:

```scheme
(case-lambda
    (() 0)
    ((a) a)
    ((a b) (+ a b))
    ((a . rest) (+ a (apply foo1 rest))))
```

is expanded in

```scheme
(lambda G12
  (let ((G11 (length G12)))
    (cond ((= G11 0) 0)
          ((= G11 1) (apply (lambda (a) a) G12))
          ((= G11 2) (apply (lambda (a b) (+ a b)) G12))
          ((>= G11 1)
           (apply (lambda (a . rest) (+ a (apply foo1 rest)))
                  G12))
          (else
           (error "no matching clause in case-lambda with ~S for ~S"
                  '(() (a) (a b) (a . rest))
                  G12)))))
```

whereas with the R7RS definition, its expansion is

```scheme
(lambda args16
  (let ((len18 (length args16)))
    (let-syntax
      ((%%cl21
         (syntax-rules
           #:::
           ()
           ((%%cl21) (error "no matching clause"))
           ((%%cl21 ((p #:::) . body) . rest)
            (if (= len18 (length '(p #:::)))
              (apply (lambda (p #:::) . body) args16)
              (%%cl21 . rest)))
           ((%%cl21 ((p #::: . tail) . body) . rest)
            (if (>= len18 (length '(p #:::)))
              (apply (lambda (p #::: . tail) . body) args16)
              (%%cl21 . rest))))))
      (%%cl21
        (() 0)
        ((a) a)
        ((a b) (+ a b))
        ((a . rest) (+ a (apply foo2 rest)))))))
```

As said, it is also slower (mostly in the first example)

```scheme
(time (repeat 5_000_000 (foo)))
Elapsed time: 431.212 ms    (define-macro)
Elapsed time: 1583.687 ms   (define-syntax)

(time (repeat 5_000_000 (foo 1 2)))
Elapsed time: 1924.322 ms   (define-macro)
Elapsed time: 2198.109 ms   (define-syntax)

(time (repeat 5_000_000 (foo 1 2 3 4 5 6 7 8 9 10)))
Elapsed time: 31065.999 ms (define-macro)
Elapsed time: 33364.158 ms (define-syntax)
```

However, I stole your tests and integrated them in `test-r7rs.stk`

___

Since, I wrote the previous message, I have rewritten the 'case-lambda` macro
to be hygienic. It is now a bit less readable, but I could uncomment your final test.

PR #741, opened 2025-03-19
[trivial] Update `srfis.stk`

jpellegrini:
@egallesio we should probably have a script to update this from the SRFI site: https://github.com/scheme-requests-for-implementation/srfi-common/

This is the file wit SRFI data:

https://raw.githubusercontent.com/scheme-requests-for-implementation/srfi-common/master/admin/srfi-data.scm

Maybe:

1. Only store the number of implemented SRFIs in a list, `(define implemented-srfis '(0 1 2 4 ...))`
2. Have the compiler fetch the data from the site
3. Have a version of the SRFI data within STklos source, to be used when (2) doesn't work (due to lack of network access for example)

What do you think?

### DIFF
--- a/lib/srfis.stk
+++ b/lib/srfis.stk
@@ -279,21 +279,26 @@
     (238 "Codesets" () "srfi-238")
     ;; 239 Destructuring lists
     ;; 240 Reconciled records
-    ;; 241 Match -- Simple Pattern-Matching Syntax to Express Catamorphisms on Scheme Data (drafr)
-    ;; 242 The CFG Language (draft)
-    ;; 243 Unreadable objects (draft)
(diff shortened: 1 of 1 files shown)


egallesio:
Merged. Thanks, @jpellegrini.

Having a script would be cool, event if it is quite some work, for only adding comments in a file :laughing: .

> Only store the number of implemented SRFIs in a list, `(define implemented-srfis '(0 1 2 4 ...))`

We already have this information: it can be obtained with `(cdr ( assoc 'scheme.srfi (version-alist)))`

> Have the compiler fetch the data from the site

I don't understand why you want to use the compiler for that. Its is not sufficient to run the script from time to time?

Issue #745, opened 2025-03-19
`exit` should not signal errors


jpellegrini:
Hi @egallesio !

See this:
```
stklos> (exit "a string")
**** Error:
exit: bad return code '"a string"'
```
But `exit` should never fail. From R7RS:

`(exit obj)`: (...) Otherwise, exit should translate obj into an appropriate exit value for the operating system, if possible. _The exit procedure must not signal an exception or return to its continuation_.

`(emergency-exit obj)`:  Terminates the program without running any outstanding dynamic-wind after procedures and communicates an exit value to the operating system _in the same manner as exit_.

We can change the Scheme object to something like "the first byte of its representation in memory", or to "always 1", or whatever...

Or do you think it should stay as it is?

egallesio:
Yep! I close it.
Thanks @jpellegrini

PR #746, opened 2025-03-19
[simple] Make exit be as per R7RS

jpellegrini:
See issue #745 . 

`exit` should never signal errors, and it should actually exit (never get back to the REPL or to the program).

Bigloo, Chibi, Chez, Cyclone, Gauche, Guile never signal errors or return from `exit`. They all ignore the second argument, and will exit the program regardless of the type of the first argument.

Chicken will only signal an error for wrong type. For several arguments, so long as the first  is an integer, it does exit the program. But it does not accept booleans (?)

Loko, MIT Scheme and Sagittarius only signal an error for too many arguments. They will exit when the single argument is not a number.

Kawa and Gambit behave as STklos today (signal erorrs for too many arguments, and for wrong type).

Biwa and LIPS don't have an `exit` procedure.

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -1274,21 +1274,24 @@ DEFINE_PRIMITIVE("%pre-exit", pre_exit, subr1, (SCM retcode))
  * If  program has registered exit functions with |register-exit-function!|,
  * they are called (in an order which is the reverse of their call order).
  * @l
- * NOTE: The {{stklos}} |exit| primitive accepts also an
- * integer value as parameter (R7RS accepts only a boolean).
+ * This primitive accepts an integer value as parameter. If the value is not
+ * a small integer, then the integer 1 will be returned to the operating
+ * system. Other arguments are ignored.
 doc>
 */
-DEFINE_PRIMITIVE("exit", exit, subr01, (SCM retcode))
+DEFINE_PRIMITIVE("exit", exit, vsubr, (int argc, SCM *argv))
 {
   long ret = 0;
   SCM cond;
 
-  if (retcode) {
+  if (argc > 0) {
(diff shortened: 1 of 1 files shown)


egallesio:
Fine. 
Even if most of the implementations accept more than 1 parameter, it seems to deviate from R7RS which accepts 0 or 1 parameter only. Loko and MIT seem to be right on this point. Of course, we can say that is an extension (and the trick is done) .
What do you think?

jpellegrini:
> What do you think?

 I think they do that in order to keep the spirit of the procedure (never complain, just exit -- even if the argument is not the expected -- and they extend it to "even if more than the expected number").
I'm not sure about the number of arguments. Maybe just drop the restriction to "boolean or integer"? I really have no strong opinion on that.

egallesio:
Hi @jpellegrini.

I have finally understood why they admit more parameters (at least Gauche and Chez). They permit to have an exit handler, who could for instance display all the parameters before exiting. 
In STklos, we can register exit procedure (inspired from the Posix `atexit` function ) and this is quite redundant.

Anyway, it would be easy to have it in Scheme. Something like:

```scheme
(define sysexit exit)
(define exit-handler (make-parameter #f))

;; Register a very simplistic exit handler
(exit-handler  (lambda rest
                 (apply eprintf rest)
                 (newline (current-error-port))
                 (sysexit 42)))


(define (exit obj . rest)
  (if (null? rest)
      (sysexit obj)
      (if exit-handler
          (apply (exit-handler) obj rest)
          (sysexit #f))))    
```

Note: It is is easier to do it in Scheme, also because we use STk_exit in some places in the C code.

PR #755, opened 2025-03-25
C closures, or "parameterizable primitives"

jpellegrini:
@egallesio I'm not sure this is really useful, but my feeling is that it would be. Sometimes we need to write a procedure that returns a procedure, so it "cannot be written in C". But... It actually can! (So long as we make some small changes to the VM and to `stklos.h`. See if this makes sense...

### Description:

We can now build C primitives that behave like closures:

- Define a primitive that creates another primitive of type `param_vsubr`
- This new primitive, of type `param_vsubr`, will ALWAYS add its parameter as last argument to itself (the VM does this)

How it is used: 

In C, write

```c
DEFINE_PRIMITIVE("multiplier", multi, vsubr, (int argc, SCM *argv))
{
  /* argv[0] is the argument given by the user; argv[-1] is
     the argument stored in the primitive object. */
  return STk_mul2(argv[0], argv[-1]);
}

DEFINE_PRIMITIVE("make-multiplier", make_mult, subr1, (SCM par))
{
  SCM p;
  NEWCELL_WITH_LEN(p, param_vsubr, sizeof(struct pprimitive_obj));
  PRIMITIVE_NAME(p) = "multiplier";
  PRIMITIVE_FUNC(p) = STk_multi;
  PRIMITIVE_PARAM(p) = par;
  return p;
}
```

And from Scheme:

```scheme
(define times-five (make-multiplier 5))
(define times-four (make-multiplier 4))

(times-five 3) => 15
(times-four 3) => 12

stklos> times-four
stklos> ,d times-four
stklos> (%procedure-parameter times-four)
4
stklos> (%procedure-parameterizable? times-four)
#t
```


jpellegrini:
If this is OK I can also make a second PR adding a section to the HACKING document...

egallesio:
> If this is OK I can also make a second PR adding a section to the HACKING document...

Hi @jpellegrini , 

I have looked (rapidly) at your PR and I don't see the interest of such beasts in STklos, and it seems quite limited IMHO (1 parameter closed only). In a near future, I hope to be able to reify completely the environment and we would imagine then that we have search functions in the environment (closure subr would be then normal subr and the current VM doesn't need to be modified).  

But, as I said, I just looked at your code very rapidly and, perhaps, I'm wrong.

jpellegrini:
Ah, ok!  I agree that the changes you have planned are much better. I'm closing this PR then.

PR #756, opened 2025-03-27
[simple] Add SRFI 258

jpellegrini:
This one is now final.
When it was proposed, another one was proposed as alternative (260, "Generated Symbols"). Both are very easy to implement, so I'd say we can have them both (do you disagree?)


egallesio:
Merged this [simple] PR and two others. Thanks, @jpellegrini.
Sorry for my actual slowness.

PR #761, opened 2025-04-13
[simple] Use `STk_C_make_list` in `listify_top`

jpellegrini:
This makes processing of variable args faster.

```scheme
(import (scheme list))
(define (f . rest) 1)
(define-macro (call-f-with-500-args)
  (append '(f) (iota 500)))

(time (repeat 100000 (call-f-with-500-args)))
```

Old code:
2130.932 ms
50000000 allocation

New code:
1660.077 ms
100000 allocation

```scheme
(define-macro (call-f-with-2-args)
  (append '(f) (iota 2)))

(time (repeat 10000000 (call-f-with-2-args)))
```

Old code:
1256.636 ms
20000000 allocation

New code:
1159.518 ms
10000000 allocation

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -773,11 +773,17 @@ DEFINE_PRIMITIVE("%call-for-values", call_for_values, subr1, (SCM prod))
     case 0: return STk_nil;
     case 1: return LIST1(vm->val);
     default:  {
-                SCM  res = STk_nil;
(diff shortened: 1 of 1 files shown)


jpellegrini:
> For example, I hadn't time to work on it, but it seems that `%call-for-values` also conses.

I added a commit for this...

PR #762, opened 2025-04-14
More optimizations using `STk_C_make_list`

jpellegrini:
The most interesting is APPEND. It really becomes faster, and conses much less.

```scheme
(import (scheme list))

(let ((a (iota 10000))
      (b (map - (iota 10000))))
  (time (repeat 10000 (append a b))))
```

Old code:
4097.964 ms
100000000 allocation calls

New code:
2899.566 ms
10000 allocation calls

Plus, the code doesn't use goto anymore, AND we report different errors for cyclic lists, improper lists and no-lists. :)

### DIFF
--- a/src/list.c
+++ b/src/list.c
@@ -290,10 +290,11 @@ static SCM list_type_and_length(SCM l, int *len)
 
 SCM STk_argv2list(int argc, SCM *argv)
 {
-  SCM res = STk_nil;
-
-  while (argc--) {
-    res = STk_cons(argv[-argc], res);
+  SCM res = STk_C_make_list(argc,STk_false);
+  SCM ptr = res;
+  for(int i=0; i<argc; i++) {
+    CAR(ptr) = argv[i];
+    ptr = CDR(ptr);
   }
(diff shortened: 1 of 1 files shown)


jpellegrini:
A quick `grep` or `STk_cons` shows that there are relevant places where `STk_C_make_list` could also be used:
```
lib/stklos/itrie.c
lib/srfi/170.c
lib/srfi/238.c
lib/srfi/178.c
lib/scheme/sort.c
lib/scheme/vector/base.c
lib/scheme/list.c
```
(Not counting single uses of `STk_cons`, only those that seem to be called repeatedly in order to build a list.)

egallesio:
For some reason, the code of `STk_append2` in this PR fails when booting. The code seems correct, but. I have not (yet) understood why it fails (debugging is not simple since it happens very soon). 

Anyway, I have done some tests on the other patches, and as I said in #769, the gain of `STk_C_make_list` is less important now. I will look in further detail at all the (other) patches to see the one we can keep.

PR #764, opened 2025-04-22
Add SRFI 260 - Generated symbols

jpellegrini:
Hi @egallesio !

SRFI 260 is not yet final, but I guess it won't change much. I'll leave this as a draft, and make it an ordinary PR when the SRFI goes final...

(This is substantially different from SRFI-258, because the symbols generated are interned).


egallesio:
> SRFI 260 is not yet final, but I guess it won't change much. I'll leave this as a draft, and make it an ordinary PR when the SRFI goes final...

Good idea. Thanks, @jpellegrini

jpellegrini:
Hi @egallesio ! This one is final!

egallesio:
Great !
I have just added a test to verify that the user don't pass more than one parameter. 
Thanks.

PR #768, opened 2025-05-22
Optimize rational comparison

jpellegrini:
It's a change to `do_compare`, which becomes faster and conses less. This is what we do: instead of doing (x-y) and checking if the result is negative, positive or zero, when one of the arguments is rational, we cross-multiply and call `do_compare` recursively.

Comment from the source:

```
when comparing a rational to integers, or to another rational,
instead of computing the difference, we compare
a/b compared to c/d  <-->  ad compared to bc
a   compared to c/d  <-->  ad compared to c
and so on. This is faster and conses less.
But we do NOT do this for reals and complexes, because it's a
more complicated situation.
```

In particular, when there are only fixnums involved (in the denominators/numerators etc) we don't cons at all, and the speedup is really big.

Results:

```scheme
;; BIGNUM-RAT x BIGNUM-RAT
(let ((a (/ (expt 2 300)
            (expt 5 100)))
      (b (/ (expt 3 300)
            (expt 5 100))))
  (time
   (repeat 1_000_000
           (< a b))))
;; 3061.321 ms
;; 1104000000 bytes in 22000000 allocation calls
;; 965.31 ms
;; 616000000 bytes in 10000000 allocation calls

;; FIXNUM-RAT x FIXNUM-RAT
(let ((a (/ (expt 2 15)
            (expt 5 10)))
      (b (/ (expt 3 15)
            (expt 5 10))))
  (time
   (repeat 10_000_000
           (< a b))))
;; 4606.831 ms
;; 240000000 bytes in 10000000 allocation calls
;; 940.646 ms
;; 0 bytes in 0 allocation calls

;; FIXNUM-RAT x BIGNUM-RAT
(let ((a (/ (expt 2 15)
            (expt 5 10)))
      (b (/ (expt 3 300)
            (expt 5 100))))
  (time
   (repeat 1_000_000
           (< a b))))
;; 2711.321 ms
;; 808000000 bytes in 23000000 allocation calls
;; 1001.095 ms
;; 432000000 bytes in 11000000 allocation calls

;; FIXNUM x FIXNUM-RAT
(let ((a (expt 5 10))
      (b (/ (expt 3 15)
            (expt 5 10))))
      (time
       (repeat 10_000_000
               (< a b))))
;; 3998.837 ms
;; 240000000 bytes in 10000000 allocation calls
;; 735.415 ms
;; 0 bytes in 0 allocation calls

;; FIXNUM x BIGNUM-RAT
(let ((a (expt 5 10))
      (b (/ (expt 3 300)
            (expt 5 10))))
      (time
       (repeat 5_000_000
               (< a b))))
;; 2781.935 ms
;; 1000000000 bytes in 30000000 allocation calls
;; 1770.585 ms
;; 840000000 bytes in 20000000 allocation calls

;; BIGNUM x FIXNUM-RAT
(let ((a (expt 5 200))
      (b (/ (expt 3 15)
            (expt 5 10))))
      (time
       (repeat 1_000_000
               (< a b))))
;; 838.204 ms
;; 368000000 bytes in 10000000 allocation calls (GC: 94) ;; 644.563 ms
;; 336000000 bytes in 8000000 allocation calls (GC: 82)

;; BIGNUM x BIGNUM-RAT
(let ((a (expt 5 200))
      (b (/ (expt 3 300)
            (expt 5 10))))
      (time
       (repeat 1_000_000
               (< a b))))
;; 754.978 ms
;; 360000000 bytes in 9000000 allocation calls (GC: 94) 
;; 596.94 ms
;; 328000000 bytes in 7000000 allocation calls (GC: 82)
```

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -1045,6 +1045,15 @@ static inline long complex_diff(SCM r1, SCM i1, SCM r2, SCM i2)
 static long do_compare(SCM x, SCM y)
 {
   /* ASSERT: x and y are numbers and none is a NaN */
+
+  /* RATIONALS: when comparing a rational to integers, or to another rational,
+                instead of computing the difference, we compare
+                a/b compared to c/d  <-->  ad compared to bc
+                a   compared to c/d  <-->  ad compared to c
+                and so on. This is faster and conses less.
+                But we do NOT do this for reals and complexes, because it's a
+                more complicated situation.       */
+
  Top:
   switch (TYPEOF(x)) {
     case tc_real:
@@ -1056,12 +1065,21 @@ static long do_compare(SCM x, SCM y)
         default: break;
       }
       break;
-  case tc_integer:
+    case tc_integer:
       switch (TYPEOF(y)) {
         case tc_real:     return double_diff(INT_VAL(x), REAL_VAL(y));
         case tc_integer:  return (INT_VAL(x) - INT_VAL(y));
+        case tc_rational: return do_compare(STk_mul2(x, RATIONAL_DEN(y)),
+                                             RATIONAL_NUM(y));
         case tc_complex:  return complex_diff(x, MAKE_INT(0),
                                               COMPLEX_REAL(y),COMPLEX_IMAG(y));
+        default: break;
+      }
+      break;
+    case tc_bignum:
+      switch (TYPEOF(y)) {
+        case tc_rational: return do_compare(STk_mul2(x, RATIONAL_DEN(y)),
+                                            RATIONAL_NUM(y));
         default: break;
       }
       break;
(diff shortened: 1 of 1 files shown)


jpellegrini:
Although a solution to #769 would perhaps make this PR unnecessary?

egallesio:
That is very smart! I

> Although a solution to #769 would perhaps make this PR unnecessary?

Yes, but it will always be more efficient than allocating, no?

jpellegrini:
> Yes, but it will always be more efficient than allocating, no?

You're right... :)  So maybe this one can me merged?

jpellegrini:
Added some more optimizations:


1. When comparing pos/neg or neg/pos, immediately answer instead of
   computing a-b.
2. When comparing a bignum and a fixnum, the sign of the bignum alone
   determines the result, regardless of the values of both numbers.

```scheme
(let ((a 10)
      (b (- (expt 3 700))))
  (time (repeat 6_000_000 (> a b))))

Old code: 1732.233 ms, 1968000000 bytes in 24000000 allocations
New code:  198.913 ms,          0 bytes in        0 allocations

(let ((a (expt 5 200))
      (b (- (expt 3 700))))
  (time (repeat 6_000_000 (> a b))))

1519.886 ms, 1920000000 bytes in 18000000 allocations
 233.08  ms,          0 bytes in        0 allocations
```

jpellegrini:
The second commit also uncovered a bug in `(scheme bytevector)` (Issue #792 ) -- that's why two tests fail after this PR.

Issue #733, opened 2025-03-14
Modifying and getting contents of a pointer with ffi


Retropikzel:
I'm looking for a way to mutate and get the contents of a pointer using the ffi. Something similar for example to Sagittariuses pointer-ref-c-TYPE and pointer-set-c-TYPE!. https://ktakashi.github.io/sections/section63.html

From reading the manual and searching the codebase it seems that this is not currently possible? The manual says that support is minimal and it will evolve in future versions. May I ask what are the plans?

If you do not plan to add some accessors like these, would you be open for me to implement them and contribute them to STklos?

jpellegrini:
@egallesio -- did you see this issue?

Retropikzel:
I was thinking that since other pointer functions (like cpointer->string) start with cpointer these should propably too.

I'll just paste them all here.

```
cpointer-set-int8_t!
cpointer-set-uint8_t!
cpointer-set-int16_t!
cpointer-set-uint16_t!
cpointer-set-int32_t!
cpointer-set-uint32_t!
cpointer-set-int64_t!
cpointer-set-uint64_t!
cpointer-set-char!
cpointer-set-short!
cpointer-set-unsigned-short!
cpointer-set-int!
cpointer-set-unsigned-int!
cpointer-set-long!
cpointer-set-unsigned-long!
cpointer-set-float!
cpointer-set-double!
cpointer-set-pointer!

cpointer-ref-int8_t
cpointer-ref-uint8_t
cpointer-ref-int16_t
cpointer-ref-uint16_t
cpointer-ref-int32_t
cpointer-ref-uint32_t
cpointer-ref-int64_t
cpointer-ref-uint64_t
cpointer-ref-char
cpointer-ref-short
cpointer-ref-unsigned-short
cpointer-ref-int
cpointer-ref-unsigned-int
cpointer-ref-long
cpointer-ref-unsigned-long
cpointer-ref-float
cpointer-ref-double
cpointer-ref-pointer
```

I think the next question is should the types be named exactly like in c with _t suffix or should it be left out? To me it does not matter, some implementations have them some dont.

I gues also the cpointer-set/ref-pointer could be cpointer-set/ref-cpointer?

Before I make a pull request I need to add some tests. But I'm already using this on my library [https://git.sr.ht/~retropikzel/foreign-c](https://git.sr.ht/~retropikzel/foreign-c) and it's working well.

egallesio:
[...]

PR #771, opened 2025-06-01
Modifying and getting contents of a pointer with ffi #733

Retropikzel:
https://github.com/egallesio/STklos/issues/733

I had to add more FFI types, so I added them after the current FFI types to not break anything. At some point they should propably be organized better.

Since :char is unsigned I added :schar type. It's not in line with other types where signed is without prefix but it's backwards compatible.

This version only contains cpointer-set! and cpointer-ref. If cpointer-set-TYPE! and cpointer-ref-TYPE or something like that is desired they can be built on top of this.

Feedback definitely welcome, specially about the C code as I'm not much of C coder. And no hurry.

### DIFF
--- a/tests/test-ffi.stk
+++ b/tests/test-ffi.stk
@@ -183,10 +183,10 @@
 (cpointer-set! p :boolean #f)
 (test "cpointer-set/cpointer-ref ̣boolean2" #f (cpointer-ref p :boolean))
 
-(cpointer-set! p :float 42.42)
-(test "cpointer-set/cpointer-ref ̣1" 42.42 (cpointer-ref p :float))
+(cpointer-set! p :float 3.5)
+(test "cpointer-set/cpointer-ref ̣1" 3.5 (cpointer-ref p :float))
 
-(cpointer-set! p :double 42.42)
-(test "cpointer-set/cpointer-ref ̣1" 42.42 (cpointer-ref p :double))
+(cpointer-set! p :double 3.5)
+(test "cpointer-set/cpointer-ref ̣1" 3.5 (cpointer-ref p :double))
 
 (test-section-end)


jpellegrini:
> The tests fails, it sets the value 42.42 but the return value is something like 42.419NNNNNN.... Any ideas what could be the problem?

Hmmm, 42.42 is not representable as a double float:

```
stklos> (real-precision 50)
stklos> 42.42
42.4200000000000017053025658242404460906982421875
```

Maybe this is related to the problem? Although the number you mentioned is 42.419NNNN, which is different from 42.4200000...17....

egallesio:
> I forgot to mention that the float part is little bit unclear to me. The tests fails, it sets the value 42.42 but the return value is something like 42.419NNNNNN.... Any ideas what could be the problem?

This is the conversion from double to float which loose precision. 
```c
#include <stdio.h>

int main(int argc, char *argv[]) {
  double d = 42.42;
  float f = d; 
  printf("d=%lf f=%lf\n", d, f);
  return 0;
}
```
will print:

```
d=42.420000 f=42.419998
```

You'll have to give the comparison function for this specific test, something like:

```scheme
stklos> (test "demo" 42.42 42.419)     ;; without function 
  testing demo expects 42.42 ==> ERROR: got 42.419.
stklos> (test "demo" 42.42 42.419 (lambda (x y) (< (abs (- x y)) 0.1)))  ;; testing numbers are nearly equal
  testing demo expects 42.42 ==> OK.
```

Retropikzel:
> BTW, it seems you forgot to provide your test file in the PR 😃

Sorry about that. :man_facepalming:  I committed the tests now.

egallesio:
> I switched the double and float tests from using 42.42 to using 3.5. Tests now pass. If that test is okay to you then it's okay to me.

Thanks @Retropikzel
I have integrated your test (and I let the test with 42.42 with the code given above).

Note: I have modified a lot `ffi.stk` and `cpointer.stk`:
   - code doesn't use any more small ints, but an enum.
   - types can be declared in any order
   - `cpointer-ref` and `cpointer-set!` are now in C
   - all the types are now implemented (int8, int16, ...); even `long long`, but it has to be tested on other architectures
   -  added a primitive `c-size-of`
   - Offsets in `cpointer-xxx` are now multiplied by the size of the pointed object. So `(cpointer-ref a :int 2)` permit to access the third element of C array `a`. 
  - `:char`, `:schar` and `:uchar` are implemented, with some particularities. When the value comes from C it can be an integer or a character. When the value is read from C, the value is returned as
      - an integer if it is a `:schar` (some code use `signed chars` really as small integers and negative values cannot be converted in characters
      - a character if it is a `:uchar` (since there are no negative value)
      - a character if we have a `:char`. In C, the `char` type may be signed or not depending on the compilation chain (in `gcc` it can be changed with the `-fsigned-char` or `-funsigned-char` option). On Linux x86, characters are signed by default, and a negative value will yield an error.    
 
I'm not happy with the situation with characters  which is too complex. Perhaps, we should always return an integer and the user can convert it in a character if possible/necessary. (While writing this text, I think that this is the more reasonable option. What do you think?)

Retropikzel:
Thank you very much! c-size-of was going to be next thing I would have implemented.You saved me the effort.

I think always returning integer with :char is fine. FFI's are implementation specific anyway so it's not like user can expect sometimes char (on another implementation) and sometimes integer.

Closing this pr is okay. Thank you!

PR #772, opened 2025-06-03
[trivial] Remove bars from lambda in runtime macro definition

jpellegrini:
Runs perfectly without it...

`|lamdba|` $\longrightarrow$ `lambda`


egallesio:
The reason was that the word `lambda` is represented by a `λ` symbol in my emacs (but not `|lambda|`), but this is probably too personal :smile: 

Merged

PR #781, opened 2025-06-11
[simple] Add `readline-set-option` (and for example, allow the user to set parenthesis blinking when closing them)

jpellegrini:
For example,

```scheme
(readline-set-option "blink-matching-paren" "on")
```

will make STklos blink the opening parenthesis when we close it.

And we can dynamically change the options. If we do 

```scheme
(readline-set-option "blink-matching-paren" "off")
```

if stops blinking (no restart necessary!)

### DIFF
--- a/lib/readline-complete.c
+++ b/lib/readline-complete.c
@@ -130,6 +130,11 @@ DEFINE_PRIMITIVE("%init-readline-completion-function",readline_init_completion,s
 }
 
 DEFINE_PRIMITIVE("readline-set-option",readline_set_option,subr2, (SCM option, SCM value)) {
+  if (!STRINGP(option)) STk_error("bad string ~s", option);
+  if (!STRINGP(value)) STk_error("bad string ~s", value);
+  if (STRING_SIZE(option) + STRING_SIZE(value) > 195)
+    STk_error("option and value strings too long (max 195 bytes)");
+
   char s[201];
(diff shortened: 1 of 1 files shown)


jpellegrini:
The possible options are:

https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html

jpellegrini:
I didn't include documentation this time, because of some personal problems.

jpellegrini:
Also, the procedure lives in the module `READLINE`, so not available when it is not loaded.

jpellegrini:
Also... These options cannot be set in `stklosrc`, because it is read before the module `REPL-READLINE` is loaded (it would be really nice if it was possible).

egallesio:
I have added a hook called `readline-startup-hook` to permit the initialization of `readline` options. Normally it could be done with a C function which is called at startup, but for some reason, it is called before each prompt. So, this is all in Scheme. 

I was sure that this PR  would not be simple and it was not. The problem is that we can dynamically link to libedit and it is a always ... weird. 

   The primitive readline-set-option is non-effective with libedit but it can be                                          
    called. However, it seems to not work when called from the                                                             
    readline-startup-hook (primitive is not defined in this case). But it works ,                                          
    if run in the debugger :-<                                                                                             
 
I did not understand where the problem came from, and now the hook is called only when we are linking the true, the real,  GNU readline.

PR #783, opened 2025-06-18
Implement vector patterns in syntax-rules

### LINKED ISSUE #752: vector patterns are not working in `syntax-rules`
It seems that vector patterns are not working in `syntax-rules`:
```
stklos> (define-syntax f (syntax-rules () ((_ #(a b)) b)))

stklos> (f #(1 2))
**** Error:
f: no matching clause for (f #(1 2))
```

I'll take a look into that.

jpellegrini:
Hi @egallesio !

Finally I got some time to work on this issue. This PR makes vector patterns work:

```scheme
(let-syntax ((f (syntax-rules ()
               ((f (a #(b c) d))
                (list d c b a)))))
  (f (1 #(2 3) 4)))

=> (4 3 2 1)
```

Some tests included.

Would fix #752

### DIFF
--- a/lib/mbe.stk
+++ b/lib/mbe.stk
@@ -185,7 +185,7 @@ doc>
 ;; hyg:tag uses gensym to rename variables (actually, tag the ones which will be renamed).
 ;;
 ;; e is the form
-;; kk ar ethe keywords
+;; kk are the keywords
 ;; al is an assoc list
 ;;
 ;; returns a list with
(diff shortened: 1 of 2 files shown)


jpellegrini:
As far as I can see the only missing part of `syntax-rules` after the two pending PRs[0] would be the "remembering their environment" part, which depends on us being able to compile/eval the macro body referencing their original environment (`eval` doesn't currently do that).  After that I suppose STklos' syntax-rules would be fully R7RS compliant (I could be wrong though, but if anything is missing it is a minor thing)

[0] One is this (vector patterns) and the other is #711 -- ellipsis escaping.

jpellegrini:
> so that the PR will be correctly merged and attribute you the credit.

Ah, that's not really important. But it's weird that GitHub still lists it as "Changes can be cleanly merged".. Because they can't!

egallesio:
This time, it appears as merged (and closed). 
Thanks for signaling the issue with this PR.

PR #786, opened 2025-06-19
[R7RS] Fix `syntax-rules` AND make let-syntax accept 2 macro styles

jpellegrini:
1. Currently, let-syntax does not verify if the bindings actually conform to syntax-rules spec:

```scheme
(let-syntax ((f (syntax-thing () ((f) -1)))) (f)) => -1
```

"Syntax-thing"... :)

We implement several checks, and it's now much stricter.

2. Why not let the user enter either a syntax-rule style macro or a define-macro one?

```scheme
(let-syntax ((f (lambda (x) (* x 3))))
  (f 5))
                  => 15

(let-syntax ((f (lambda args 10))
             (g (syntax-rules () ((_) 20))))
  (list (f) (g)))
                  => (10 20)
```


jpellegrini:
I tagged it "R7RS" because it fixes the too-permissive `syntax-rules`

egallesio:
My intent was to allow the two macro styles too !!!
However, the final code is simpler than the code you submitted.  Since `syntax-rules` returns now a transformer function, there is nearly no special code in let-syntax (just evaluate the expander). Note that the expander can even be any `match` function which returns a lambda (even a global function :smile:)

I have added all your sanity checks to the `mbe` code and the tests.  Thanks a lot.

PR #795, opened 2025-07-11
Add `bytevector-sint-set!`, which was missing from `(scheme bytevector)`.

jpellegrini:
And provide tests...

The method for setting negative bignums in a K-byte region is to wrap around the range by adding the range itself to the number (this is necessary because `mpz_export` only knows magnitudes, not signals).


egallesio:
I have not investigated, but tests didn't pass on this one.

jpellegrini:
Yes, I see what the bug is... I must have gotten confused somehow. I thought I did run the tests in advance. I'll check

PR #799, opened 2025-07-21
Add SRFI 225 - Dictionaries

jpellegrini:
@egallesio , this is SRFI-225.

From the SRFI spec:
_The following DTOs are exported from this SRFI: srfi-69-dto, hash-table-dto, srfi-126-dto, mapping-dto, and hash-mapping-dto, provided that the corresponding libraries are available. In addition, eqv-alist-dto and equal-alist-dto are unconditionally exported._

So, this one usees SRFI-125 only, not 69... I hope this is OK.

### DIFF
--- a/lib/srfi/225.stk
+++ b/lib/srfi/225.stk
@@ -172,4 +172,7 @@
      equal-alist-dto)
 
     (import (srfi 225 srfi-125-impl))
-    (export hash-table-dto))
+    (export hash-table-dto)
+
+    (import (srfi 225 srfi-69-impl))
+    (export srfi-69-dto))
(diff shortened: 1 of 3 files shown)


jpellegrini:
> ~So, this one usees SRFI-125 only, not 69... I hope this is OK.~

Just added support for SRFI-69...

jpellegrini:
@egallesio I see that the SRFI document does not *require* the implementation as a library and sublibraries.
So it would be OK to include all code in a single file. These are in `lib/srfi/225/`:

* `core.stk`
* `core-impl.stk`
* `alist-impl.stk`
* `default-impl.stk`
* `srfi-69-impl.stk`
* `srfi-125-impl.stk`

Do I include them  literally in a single file, or can I put them in `/lib/srfi`, with appropriate names, and `include` them from `225.stk`?

Some adaptation may be needed I think (in order to avoid namespace conflict), but nothing complicated.

jpellegrini:
See #867

PR #802, opened 2025-07-27
[simple] Don't redefine some string procedures in SRFI-13

jpellegrini:
They're already in the core, as primitives, and their signatures are identical.

Passes all tests. :)


egallesio:
Done. I have just re-imported those functions from the SCHEME module, in case the user has redefined them.
Thanks.

PR #804, opened 2025-07-31
Rewrite any and every in C

jpellegrini:
Reasons:

- they're very important, used in several places of bonus.stk and also in SRFIs
- it becomes incredibly faster and conses much less
- the C code is almost the same in size

Timings:

1. Fast path (single argument)
```scheme
(let ((X (iota 5000)))
  (time
    (repeat 10000 (any null? X))))

Old code:
2529.879 ms
Allocations: 1280000 bytes in 30000 allocation calls

New code:
1030.316 ms
Allocations: 0 bytes in 0 allocation calls
```

2. Normal path (two argument in this case)

```scheme
(let ((X (iota 5000))
      (Y  (iota 5000)))
  (time
    (repeat 10000 (every = X Y))))

Old code:
17490.161 ms
Allocations: 4801760000 bytes in 100050000 allocation calls

New code:
2991.282 ms
Allocations: 800960000 bytes in 50020000 allocation calls
```

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -28,7 +28,6 @@
         ... => _ else
         improper-list?
         remove remove! delete delete!
-        every any
         call-with-input-string call-with-output-string
         read-case-sensitive
         open-input-virtual open-output-virtual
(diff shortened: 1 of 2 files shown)


jpellegrini:
I also changed `boolean=?` in `r7rs.stk`.



The new version is:

```scheme
(define (boolean=? . args)
  (or (every (lambda (x) (eq? x #f)) args)
      (every (lambda (x) (eq? x #t)) args)))
```

With the new `every` this is also faster and conses less.

```scheme
(let ((a (map not (iota 1000))))
  (time
    (repeat 5000 (apply boolean=? a))))

Old code:
751.712 ms
120480000 bytes in 20000 allocation calls

new code, *without* the new every:
492.645 ms
121040000 bytes in 30000 allocation calls

new code, *with* the new every:
327.0 ms
120400000 bytes in 15000 allocation calls
```

jpellegrini:
Now I realized:

* `symbol=?` can be optimized in the same way as `boolean=?`
* Both procedures would take at least two elements, according to R7RS.

So I included another patch in this PR addressing those issues.

For `symbol=?`:

```scheme
(let ((a (make-list 20000 'x)))
  (time
    (repeat 1000 (apply symbol=? a))))

Old code:
1786.216 ms
480096000 bytes in 4000 allocation calls

New code:
1276.569 ms
480080000 bytes in 3000 allocation calls
```

jpellegrini:
@egallesio -- I see in the errata of R7RS that " In fact it is an error to apply `boolean=?` to non-booleans". So we may leave it as it is (because R7RS lets us decide what to do with errors) or signal an error. Which would be preferable?

jpellegrini:
> @egallesio -- I see in the errata of R7RS that " In fact it is an error to apply `boolean=?` to non-booleans". So we may leave it as it is (because R7RS lets us decide what to do with errors) or signal an error. Which would be preferable?

I have added a last patch to the PR, fixing that -- so if you prefer, you may ignore this last patch...

egallesio:
> I have added a last patch to the PR, fixing that -- so if you prefer, you may ignore this last patch...

I have kept the test. I rewrote the function to be more efficient.

PR #812, opened 2025-08-08
[trivial] "R7RS" always in italics and with superscript 7 in `README.md`

jpellegrini:


jpellegrini:
Also fix rendering of STklos in README.md

PR #816, opened 2025-08-15
Document `icall` objects

jpellegrini:
And STklos' extendable special form set!

(I understand how it works -- it's really cool!)

### DIFF
--- a/lib/stklos/describe.stk
+++ b/lib/stklos/describe.stk
@@ -213,6 +213,13 @@ doc>
             (format port "an unnamed mutex with state ~a and ~a specific field"
                     state specific))))
 
+     ((%icall? x)
+      (format port "an icall (internal compiler call) object, ")
+      (format port "whose name is `~a' and compiling function set to ~a"
+              (%icall-name x)
+              (%icall-function x)))
+
+
      (else
       (let ((user-type-name (%user-type-name x)))
         (if user-type-name


doc/hacking/hacking.adoc
```diff
@@ -1045,6 +1045,145 @@ The code is generated as a list, in the `*code-instr*` global variable
...
+
+[source,scheme]
+----
+stklos> (define-macro (double . args)            `(#%double ,@args))
+----
+
+We may have an `icall` object with a different name from its compiling function. The
+procedure `%icall-function-set!` allows one to define the compiling procedure for
+that special form.
+
```
egallesio, line 1174: That is not the idea. The function slot in the icall object is `#f` and is changed at the first call. It is a kind of cache to avoid searching the associated function   at each call. This should not be modified by the user. I have added a comment at the beginning of `icall.c` and in the compiler.

doc/hacking/hacking.adoc
```diff
@@ -1045,6 +1045,145 @@ The code is generated as a list, in the `*code-instr*` global variable
...
+defined inside the `STKLOS-COMPILER` module) is called.
+
+[source, scheme]
+----
+(if bool something)
+  ~~> (#%if bool something)
+    ~~> (compile-%%if '((#%if bool something) #f #f)     ;; if env and tail are both #f
+----
+
+We now implement a special form `twice`, that compiles the code inside it twice.
```
egallesio, line 1112: I think that there is misunderstanding here. Internal icalls are useful for structure which need to produce code (e.g. new instructions), or need to access to the internals of the compiler. `Twice` could be easily implemented by a macro:

```scheme
( define-macro (twice . args) `(begin ,@args ,@args))
```
I think, that the example with `twice` is confusing and should not be here.

egallesio:
BTW, I have better performance when running `make tests` (as expected) on my desktop (A 12 years old, intel i7). Make tests uses now 5.5s  (instead of 6.5s). 
On my laptop, which is an i5 (only 10 years old), I don't see a big difference. I don't understand why). 
Have you seen a difference on your machine?

PR #819, opened 2025-09-04
[simple] Fix crash on integer division with infinities

jpellegrini:
This patch fixes the following problem:

```
stklos> (quotient -inf.0 2)
Floating point exception

stklos> ,i scheme/flonum
stklos> (flodd? +inf.0)
Floating point exception
```

and includes tests.

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -811,7 +811,10 @@ static inline SCM real2integer(SCM r)
 {
   double v = REAL_VAL(r);
 
-  if (floor(v) != v) {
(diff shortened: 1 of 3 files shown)


egallesio:
Oops, will fix it asap.

egallesio:
Finally, not so simple (tests were no more running). I had to add a handler in the constant folding code.
I'm not too happy with the current code, since we have two with-handler and I tried to factorize with a unique handler without success. It is on my TODO list.

Thanks for this PR.

PR #827, opened 2025-10-02
[simple] Fix `fxsqrt` and optimize `exact-integer-sqrt`

### LINKED ISSUE #826: `fxsqrt` returns wrong value for edge cases
```
stklos> (fxsqrt 1152921504606846753)
1073741824
-223
```

It should have been

```
1073741823
2147483424
```

I'm working on it...

jpellegrini:
```scheme
stklos> (fxsqrt 1152921504606846753)
1073741824
-223
```
Oops...

This was happening because we were not being cautious with the result from libc `sqrt`. It's fixed.
    
While we're at it, why not make `exact-integer-sqrt` use `fxsqrt` when  the argument is a fixnum?

Now `exact-integer-sqrt` is also faster *and* allocates less then it did before, for fixnums:

```scheme
(let
  ((k (- (expt 2 60) 223)))
    (time
      (repeat 5_000_000
        (exact-integer-sqrt k))))
Old code: 2148.563 ms, Allocations: 720000000 bytes
New code: 1387.103 ms, Allocations: 560000000 bytes
```
Timings for bignums ~stayed the same~ (got much better).

Included some new tests.

Fixes #826

### DIFF
--- a/lib/r7rs.stk
+++ b/lib/r7rs.stk
@@ -32,7 +32,8 @@
         exact-integer?
         floor-quotient floor/ truncate/ truncate-quotient
         truncate-remainder floor-remainder
-        square exact-integer-sqrt exact inexact
+        square
+        exact inexact
         boolean=?
         member assoc
         symbol=?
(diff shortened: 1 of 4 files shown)


jpellegrini:
I'm force-pushing a new version of this PR. We already have a method for integer square root with remainder in GMP, and another for fixnums (in `fixnum.c`). There is no reason to re-implement Newton's method in `exact-integer-sqrt`.

This time,

```scheme
(let
  ((k (- (expt 2 60) 223)))
    (time
      (repeat 5_000_000
        (exact-integer-sqrt k))))
Old code: 2148.563 ms, Allocations: 720000000 bytes
Old version of PR:: 1387.103 ms, Allocations: 560000000 bytes
THIS version: 161.115 ms, Allocations: 0
```

It's also much faster than the previous code for fixnums (of course, it's now a primitive, and also we delegate to GMP, wich is heavily optimized).

jpellegrini:
For bignums:

```scheme
(time 
  (repeat 1_000_000
    (exact-integer-sqrt
      100044444444444444444444444444444444444444444444444444440001)))
Old code: 9266.141 ms, Allocations 3512000000
New code:  368.026 ms, Allocations 128000000
```

PR #838, opened 2025-10-15
A set of primitives for debugging environments

jpellegrini:
I'm not sure this is OK, but it was reeeeeeally fun to write.

```scheme
(define e 0)
(let ((a 1)
      (b 2))
  (let-syntax ((m (syntax-rules () ((m) 10)))
               (s (syntax-rules () ((s) 20))))
    (let ((c 3))
      (set! e (#%current-environment)))))

(define f (%environment-dynamic-environment e))
(define s (%environment-static-environment e))

stklos> ,d f
stklos> (%frame-length f)
1
stklos> (%frame-ref f 0)
3

And, the really nice part:

stklos> (%debug-environment e)

 ----------------
/ FRAME with 1 variables:
| c --> 3
\
 ----------------
/ LEXICAL LEVEL with 2 syntaxes:
| s
| m
\
 ----------------
/ FRAME with 2 variables:
| b --> 2
| a --> 1
\
 ----------------
```

I think I could have written `%debug-environment` in Scheme, but it ws too late when I realized that. :smile:

### DIFF
--- a/src/env.c
+++ b/src/env.c
@@ -1026,7 +1026,7 @@ DEFINE_PRIMITIVE("%debug-environment", debug_env, subr12, (SCM e, SCM p)) {
   SCM dyn  = ENV_DYN_ENV(e);
   SCM stat = ENV_STAT_ENV(e);
 
-  STk_puts("\\\n ----------------\n", port);
+  STk_puts(" ----------------\n", port);
 
   while(CONSP(stat)) { /* For each level */
     SCM vars;


jpellegrini:
It doesn't print the frame owner...  Maybe that would be nice?

jpellegrini:
Also, this:
```c
  while(CONSP(stat)) { /* For each level */
```
means it only shows things up to the topmost lexical level... So won't work for nested function calls, like this

```
stklos> (define (g x) (let ((a 2)) (set! e (#%current-environment)) (* a x)))
;; g
stklos> (define (f x) (let ((b 10)) (- (g x) b)))
;; f
stklos> (f 50)
90
stklos> e
#[environment 7fecc99e7990]
stklos> (%debug-environment e)
 ----------------
/ FRAME with 1 variables:
| a --> 2
\
 ----------------
/ FRAME with 1 variables:
| x --> 50
\
 ----------------
```

But that would be much harder anyway (we captured `g`'s static info. Nobody did that for `f`, and we can't do that at runtime... **And** we don't know at compile time who will be calling `g`!)

egallesio:
Hi @jpellegrini,

That is really interesting. However, note that I intend to change the way environment are implemented. So, do not invest too much in the current representation of environments. 

I have already started some work on it, but I will perhaps postpone a bit, since I would like to release a stable version (the previous one is more than a year ago!) . 

I should also change the way new version are built, because it really takes too much time.

jpellegrini:
Hmm, ok, so I'll wait for the new environment  implementation.
By the way, maybe it's also good to wait for the new environments before trying new implementations of macro hygiene and macro lexical scope? I'll wait. :)

jpellegrini:
> I have already started some work on it, but I will perhaps postpone a bit, since I would like to release a stable version (the previous one is more than a year ago!) .

Any idea of what will be included from the current open pull requests?

PR #844, opened 2025-10-22
[simple] Optimize `repeat`

jpellegrini:
Before:
```scheme
(pp (macro-expand '(repeat 3 do-something)))
=> (begin
     (void)
     (let ((G4 3))
       (tagbody
         #:top
         (when (fx> G4 0)
           (set! G4 (fx- G4 1))
           do-something
           (-> #:top)))))
```

Now:
```scheme
(pp (macro-expand '(repeat 3 do-something)))
=> (begin
     do-something
     do-something
     do-something
     (void))
```

The speedup is quite interesting:

```
(time (repeat 100_000_000 (repeat 3 -1)))
Old code: 8415.120 ms
New code:  600.858 ms
```

:smile:

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -1133,19 +1133,25 @@ doc>
     (if (fixnum? count)
         ;; We can compute the number of loops at compile time and produce
         ;; simpler code. We can also use fx functions.
-        (if (and (= it 1) (positive? count))
-            ;; No loop unrolling
-            `(let ((,c ,count)) ,(%repeat c body #t))
-            ;; Unroll the loop.
-            (let ((valq (quotient count it))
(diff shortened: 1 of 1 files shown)


jpellegrini:
It does unrolling when the number requested is less than the number of iterations to unroll. Before, it was actually ignoring the idea of unrolling for tiny numbers of repetitions...

jpellegrini:
> The speedup is quite interesting:

Ok, that counts as cheating because the `begin` with only constants inside will be collapsed into a single `(void)`.

But this is interesting:

```scheme
(time (repeat 100_000_000 (repeat 3 (f))))
Old code: 13365.050 ms
New code:  5687.647 ms
```

PR #862, opened 2025-11-11
Local macros remember their static environment

jpellegrini:


jpellegrini:
I force-pushed to fix some mistakes, but the problem remains... It doesn't bootstrap, I can't see exactly why

PR #873, opened 2025-12-27
[test performance issue] Add SRFI 194 - Random Data Generators

jpellegrini:
@egallesio I have prepared SRFI 194, but there are some issues:

1. in `tests/srfis` there are two files which are included from the main test file. Is that OK?
2. This SRFI makes heavy use of closure-generating procedures. Hence, they cannot be written in C, so it's numeric code that is fully written in Scheme, *and* producing closures... So it's somewhat inefficient. I think I have before proposed something like "C closures", and the motivation was this.
3. Tests are very slow. The version here has two tests (the binomial tests, at the end of the test file) commented out, and even then, it takes more than 10x the run time of current tests.

I'm not sure how to proceed.  I have tried to optimize the SRFI and the tests, but didn't come up with very good enhancements.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -118,6 +118,7 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-190: Coroutines Generators
     - SRFI-192: Port Positioning
     - SRFI-193: Command line
+    - SRFI-194: Random data generators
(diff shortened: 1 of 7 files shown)


jpellegrini:
Also, one of the procedures initializes a table, and is not thread-safe (as per the comment in the implementation). But that can be easily solved; I was more concerned about the performance issues.

PR #876, opened 2026-01-20
[trivial] Fix 0/0 (which should not be zero)

jpellegrini:
```
stklos> (/ 0 0)
0
```

Oops...

This one-liner fixes it. :smile:


egallesio:
Well seen. @jpellegrini!!
Merged. Thanks a lot.

PR #882, opened 2026-03-25
Don't use K&R C

jpellegrini:
I think I found some pre-ANSI-C artifact in STklos... :grin: 

In `gnu-getopt.c`, one function was still using old style function definition, with argument types inside the function!

In Android/termux, this fails because of stricter behavior of clang. This will likely also break in other systems too in the near future.

### DIFF
--- a/src/gnu-getopt.c
+++ b/src/gnu-getopt.c
@@ -272,9 +272,7 @@ static char *posixly_correct;
 // STklos: #endif
 
 static char *
-my_index (str, chr)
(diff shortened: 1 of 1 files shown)


egallesio:
> Also maybe updating it with newer `getopt` would make sense?

I have done that in the new commits.
Now, all the modifications are at the top of the file and the end of the file is a completely unmodified version of `getopt`.  This should simplify possible future evolutions, if needed.

PR #884, opened 2026-04-24
R7RS Tests

Retropikzel:
Summer vacations are approaching, I was worried you might not have enough to do so I added some tests. :D 

jk, I wanted to learn making github workflows and thought I'd help add some more testing at the same time. And making use of that free compute github offers. The tests that error out the whole thing are commented out with sed, so are tests that use (chibi test) features and do not work with (srfi 64).

Please note that this is not some request to fix all of these things. Hopefully it is helpfull if for nothing else than catching regressions.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -3,19 +3,89 @@ name: C/C++ CI
 on: [push]
 
 jobs:
-  build:
-
+  build-ubuntu:
+    name: build-ubuntu
     runs-on: ubuntu-latest
-
     steps:
-    - uses: actions/checkout@v1
-    - name: timestamp correction
-      run:  ./utils/time-correct
-    - name: configure
-      run: ./configure
-    - name: make
-      run: make -j1
-    - name: make check
-      run: make check
-#    - name: make distcheck
-#      run: make distcheck
+      - uses: actions/checkout@v6
+      - name: timestamp correction
+        run:  ./utils/time-correct
+      - name: configure
+        run: ./configure
+      - name: make
+        run: make -j1
+      - name: make check
+        run: make check
+  build-alpine:
+    name: build-alpine
+    runs-on: ubuntu-latest
+    container:
+      image: alpine
+    steps:
+      - name: Install packages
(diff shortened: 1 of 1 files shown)


egallesio:
Hi @Retropikzel, 
It is really a good idea to add the (numerous) tests from Chibi to catch regressions. Note that there are already a bunch of Chibi tests which are already present in the `tests/test-r7rs.stk` file which is executed by the GitHub CI/CD at each commit. So there is some duplication here. 

I must admit that I don't see the advantage to use GitHub workflow, rather than integrate the test file in STklos distribution (if Alex agrees of course). Having it in the distribution permits to test it locally before pushing errors in a public repository. Moreover, I don't understand how the eventual errors will be visible if some test failed (I suppose
that it will be in the `log` file you specified in the workflow). Is it the case? Will the check sign at top of the Code page will be invalided if an error occurred?

Retropikzel:
Any tests that caused an error and the workflow to fail are commented out in the workflow file using sed. So the check sign will always be validated. The results are in the log artifact. Here is my forks run https://github.com/Retropikzel/STklos/actions/runs/24885105957

```
# of expected passes      1141
# of unexpected failures  53
```

I do agree that the overlap in tests is not a good thing and being able to run them locally is much better from usability standpoint.

Perhaps I should add the failing tests, or some of them, into the test file and make a pull request about that?

Edit: Or maybe I could file an issue listing them? They do error so there would be quite a lot to fix in one go.

egallesio:
Thanks for the explanation of how to grab the result. That's really interesting.

I have started to see how to integrate the Chibi file in the standard STklos tests and I have already seen some simple bugs that could be easily corrected. But as you said, there is a lot of work to pass all the tests :wink: 

Thanks, @Retropikzel, for this PR.

Retropikzel:
@egallesio Thank you!

egallesio:
I have fixed several minor problems thanks to your PR (they are pushed now). I have not yet integrated the chibi test file in the standard distribution (it will be, when the number of problems will be lowered)

Issue #885, opened 2026-04-28
Make "`rm`" not absolute reference to `/bin/rm` ?


jpellegrini:
The Guix package for STklos has to change it, for example:

```scheme
(define-public stklos
  (package
    (name "stklos")
    ...
              (add-after 'configure 'patch-rm-references
            (lambda _
              (let ((rm (which "rm")))
                (substitute* (find-files "." "^Makefile$")
                  (("/bin/rm") rm))))))))
```
This is from: https://github.com/lfam/guix/blob/master/gnu/packages/scheme.scm

Some Makefiles call it as `/bin/rm`, which... Is not guaranteed to work! (At least it doesn't work on Guix):

```
$ which rm
/home/jeronimo/.guix-profile/bin/rm
```

Of all 3rd party libraries, ony GMP seems to have the same problem. The occurrences:
```
grep 'bin/rm' `find|grep Makefile.am`
./src/Makefile.am:	/bin/rm -f *.o $(DOCDB) git-info.h boot.ok stklos
./src/Makefile.am:	/bin/rm -f boot.img Makefile extraconf.h stklos stamp-h1 stklosconf.h
./gmp/Makefile.am:	/bin/rm -f *.o *~ libgmp.a .libs/libgmp.a
./gmp/Makefile.am:	/bin/rm -f config.h stamp-h1
./gmp/Makefile.am:	/bin/rm -f config.status config.cache config.log configure.lineno \
./gmp/Makefile.am:	/bin/rm -rf ./.deps ./libs ./autom4te.cache
./gmp/Makefile.am:	/bin/rm -f Makefile
./doc/Makefile.am:	/bin/rm -f $(ALLDB)  stklos*.1 *~
./doc/Makefile.am:	for d in $(SUBDOCS); do (/bin/rm -f $$d/Makefile); done
./doc/Makefile.am:	/bin/rm -f Makefile
./extensions/curl/demos/Makefile.am:	/bin/rm -f $(EXE) *~ 
./extensions/curl/demos/Makefile.am:	/bin/rm -f Makefile
./extensions/curl/lib/stklos/Makefile.am:	/bin/rm -f $*.o
./extensions/fuse/demos/Makefile.am:	/bin/rm -f $(EXE) *~ 
./extensions/fuse/demos/Makefile.am:	/bin/rm -f stklos-fuse.so
./extensions/fuse/demos/Makefile.am:	/bin/rm -f Makefile
[...]

PR #886, opened 2026-04-28
Do not use absolute path for `rm`

jpellegrini:
This is a possible fix to #885 although this PR does not change `gmp/Makefile.{in,am}`...

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -73,11 +73,11 @@ $(DOCDB): $(OTHERDB)
 # Documentation
 #
 clean:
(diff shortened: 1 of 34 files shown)


egallesio:
I have merged your PR, but I'm still thinking it is dangerous. Using `$(RM)` instead of `rm -f`  (which can can be overloaded if needed) permits perhaps more customization. 

Anyway, during the conversion of a Makefile.am to a final Makefile, it seems that GNU tools introduce some `rm -f`. So perhaps, we can let everything in the current state.

PR #371, opened 2022-05-08
Optimize delete-duplicates in SRFI-1

jpellegrini:
Hi @egallesio !

While working on SRFI-116 I noticed that it had a quadratic algorithm for
`delete-duplicates`, and it also included a note about possible faster
approaches. In SRFI-116 I used a hash-table based element-marking
algorithm which runs in linear time (no tquadratic as the reference 
implementation).
Since SRFI-116 is based on SRFI-1, I thought I'd take a look  into SRFI 1 
also. So:

The reference implementaiton of SRFI-1 has the following caveat:

```
;; Beware -- these are N^2 algorithms. To efficiently remove duplicates
;; in long lists, sort the list to bring duplicates together, then use a
;; linear-time algorithm to kill the dups. Or use an algorithm based on
;; element-marking. The former gives you O(n lg n), the latter is linear.
```

I have implemented a simple element-marking algorithm for 
`delete-duplicates` using hash tables (because STklos has 
very good hash tables), and it is consistently faster for lists of size 
larger than 10.  It does require more space, since besides the list 
being build we also keep a hash table as large as the set of elements, 
but the difference in running time is quite large.

What do you think?

If you believe the extra space is too much, we can do the O(n log(n)) approach,
using quicksort (or heapsort, so no extra space is used).

## Some tests

I didn' timplement real benchmarks, but used the following code to have an 
idea of how faster the hash table approach is. And it really is faster!

```scheme
(import (srfi 1))
(import (srfi 27))

(define (delete-duplicates/hash lis :optional (eq equal?))
  (unless (list? lis) (error "bad list ~S" lis))

  (if (null? lis)
      '()
      ;; delete-duplicates with expected linear time.
      (let ((h (make-hash-table eq))
            (mark #(1))) ;; unique reference, guaranteed to
        ;; not be present in lis
        (let Loop ( (lis lis) )
          (if (pair? lis)
              (if (eq? (hash-table-ref/default h (car lis) mark) mark)
                  ;; not found, include it
                  (begin (hash-table-set! h (car lis) #t)
                         (cons (car lis) (Loop (cdr lis))))
                  ;; found (hash table returned mark), ignore it
                  (Loop (cdr lis)))
              lis)))))


;; Creates a random list, with random integers elt, such that
;; min-elt <= elt < max-elt
(define (make-random-list size min-elt max-elt)
  (let ((v (make-vector size)))
    (dotimes (i size)
      (vector-set! v i (+ min-elt
[...]

PR #890, opened 2026-05-19
Fix hash of circular lists

### LINKED ISSUE #889: `hash-table-hash` fails on circular structures
I just saw a report of this happening with `default-hash` on Gauche, and the same thing seems to happen with `hash-table-hash` in STklos:

```
stklos> ,i srfi/1
stklos> (define L (circular-list 10 20 30 40))
;; L
(hash-table-hash L)    <= infinite loop
```

I'm sure we can get this to work. I don't have enough time not, but I'm willing to fix this in the next weeks, if you think it's OK, @egallesio

jpellegrini:
This may fix #889.

We will only know if a list is circular after calling `STk_list_type_and_length`, so we do it when we get a cons to hash. If it is circular, the hash is length + the hash of the list if it were not circular. Other cases are unaffected.

Does it look OK?

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -120,10 +120,21 @@ static unsigned long sxhash(SCM obj)
   if (!BOXED_OBJP(obj)) return (AS_LONG(obj) >> 2);
 
   switch (BOXED_TYPE(obj)) {
-    case tc_cons:       h = sxhash(CAR(obj));
-                        for(tmp=CDR(obj); CONSP(tmp); tmp=CDR(tmp))
-                          h = HASH_WORD(h, sxhash(CAR(tmp)));
-                        h = HASH_WORD(h, sxhash(tmp));
+    case tc_cons:       int len;
+                        tmp = STk_list_type_and_length(obj, &len);
+                        /* If tmp is a cons, the list is circular! */
+                        if (CONSP(tmp)) {
+                          /* Sum len to h just to make the hash very different from
+                             the hash of the similar non-circular list: */
+                          h = sxhash(CAR(obj)) + (unsigned long) len;
+                          for(tmp=CDR(obj), i=1; i<len; tmp=CDR(tmp), i++)
+                            h = HASH_WORD(h, sxhash(CAR(tmp)));
+                        } else { /* non-circular list: */
+                          h = sxhash(CAR(obj));
+                          for(tmp=CDR(obj); CONSP(tmp); tmp=CDR(tmp))
+                            h = HASH_WORD(h, sxhash(CAR(tmp)));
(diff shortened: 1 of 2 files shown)


jpellegrini:
> the hash is length + the hash of the list if it were not circular

Actually,

`hash(CAR) + length`, then proceed to compute the hash as if it were a non-circular list.

egallesio:
Hi @jpellegrini,

Finally, it is more complex than I initially thought. It doesn't work on list with a circularity (not what is called a circular list). 

```scheme
stklos> (circular-list? '#0=(1 2 . #0#))
#t
stklos> (hash-table-hash '#0=(1 2 . #0#))
189
``` 
Here, it is OK with your code.

BUT
```scheme 
stklos> (circular-list? '#0=(1 2  #0#))
#f
stklos> (hash-table-hash '#0=(1 2  #0#))
segmentation fault
```

The general solution to this problem seems to be as complex as writing a list with shared structures. I don't like that: if the computation cost of the hash value is high, we will lose the interest of having a O(1) access to the table.   Furthermore, I'm not sure that using such lists as keys will be too frequent. 

I must admit that I don't know what to do here. Perhaps, keep the code of this PR, but signal in the documentation that this is not fully "rock solid".  

What do you think?

jpellegrini:
Hmm, maybe using just part of the list? This would give the same hash to lists that share the first N elements, but could work...

jpellegrini:
If a `max_depth` were used instead of trying to compute length etc, it would work also to restrict computing on very large lists, and would strictly keep the time $O(1)$... "Whatever the structure, at most N items are used to compute the hash".

egallesio:
Hi @jpellegrini, 

It's funny: I had the same idea! I have a prototype which seems to work.  I have to go just now, but I should have a commit later today (and it works on vectors with shared values too).

egallesio:
Hello @jpellegrini,

This is fixed now, for circular lists and lists with shared structures. 

However, when I fixed the problem with circular structures, it pops an issue with the implementation of  `delete-duplicates` in  PR #371. In fact the implementation was incorrect. We had a test 

```scheme
(delete-duplicates '((a . 3) (b . 7) (a . 9) (c . 1))
     (lambda (x y) (eq? (car x) (car y)))))
```

This test worked before because `(a . 3)`  and `(a . 9)` where in the same bucket and the predicate permits to not insert `(a . 9)` in the result.  Now, the hash value given by this two conses dispatches in two different buckets and it is no more OK. So, I have restored the original code (but I think that your code should be OK if the predicate is `equal?`, so perhaps we could choose your implementation if the predicate is `eq?` and the general case otherwise. A bit complicated, but perhaps interesting (I have not tested).

PR #891, opened 2026-05-20
Support return type checking in SRFI 253

aartaka:
This extension is described in [SRFI 273](https://srfi.schemers.org/srfi-273/srfi-273.html#spec--return-value), along with others (to be contributed separately.)

### DIFF
--- a/lib/srfi/253.stk
+++ b/lib/srfi/253.stk
@@ -129,7 +129,13 @@
 
 ;; ----------------------------------------------------------------------
 (define-syntax %lambda-checked
-  (syntax-rules ()
+  (syntax-rules (=>)
+    ((_ name (=> (returns ...) body ...) args (checks ...))
+     (lambda args
+       checks ...
+       (values-checked
+        (returns ...)
+        (begin body ...))))
     ((_ name (body ...) args (checks ...))
      (lambda args
        checks ...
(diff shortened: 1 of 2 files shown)


egallesio:
Hi @aartaka, 

Thanks for your contribution. I'll be happy to add it to STklos (and pleased to have a new contributor too :wink: ). 
However, have you tested it? I tried to add some tests in `tests/srfi/253.stk` and failed miserably. I think that there is an issue with `=>` which is seen as the `=>` in `cond` syntax. This probably due to the poor implementation we have of `define-syntax` in STklos, or I have not understood how to use type checking on return values. 
It would be cool if you ensure it is OK with a couple of tests.
Thanks

aartaka:
@egallesio I’ve added new tests for `lambda-checked` and `case-lambda-checked`, and they seem to be really confusing, so it must be something with the compiler. Can you check?

aartaka:
Sorry, I made typos in two of the tests. Now they should be representable. And there’re still syntax errors :frowning_face: 

```
	 Where: in compiler-error
	Reason: '#[closure 7fe2a53e1880]' requires at least 3 arguments (2 provided). Call was (test "lambda-check.16" ((lambda-checked ((b integer?)) => (integer?) 2) 1))

  - error
  - compiler-error
  - %macro-expand
  - <<let/call>>
  - compile
  - #[closure 7fe259c91b40]
  - <<let/call>>
  - #[closure 7fe259c91ac0]
  - %call-for-values
  - <<let/call>>
```

egallesio:
Hi @aartaka, 

Finally, the problem was that some calls to the `test` macro didn't provide the expected value. One corrected, everything was  OK. Your PR is now merged.
So everything is fine (and I didn't have to modify the implementation of `define-syntax` :smiley: , as I initially thought)

Thanks a lot for this contribution and for your help.

aartaka:
@egallesio thank you! I feel silly, this was a really simple mistake. But I’m glad it’s resolved now! Cheers!

PR #893, opened 2026-05-28
Shorten the assume in SRFI 253 check-arg—was redundant

aartaka:
Just a minor improvement—no more duplication of `assert` expression in SRFI 253.


egallesio:
Thanks for this PR @aartaka. It is merged.

PR #899, opened 2026-06-10
Fix typo in manual

### LINKED ISSUE #897: Typo in reference manual
In the FFI section, "whih takes an int" → "which takes an int".

jpellegrini:
Fix #897


egallesio:
Thanks @jpellegrini for the fix. However, GitHub is unhappy because, I have also rebuilt the PDF. It shoul be OK but I have to close it manually.

Issue #856, opened 2025-10-30
`expt` doesn't seem to deal with large fixnum exponents


jpellegrini:
`expt` doesn't seem to deal with certain exponents well:

```
stklos> (expt 2 1000000000000000000)
Floating point exception   stklos
```

The exponent is a fixnum, but too large...

jpellegrini:
Maybe the libgc API has something that can be used to determine if we will have enough memory? But this may be too much of a wish, I suppose... Some other Schemes also crash in teh same situation

egallesio:
Hi @jpellegrini, 

I have added a small handler for this case, but was not able to do a good job except exiting with a STklos message, rather than a system one. 

Now:
```scheme
stklos> (expt 2 1000000000000000000)
Received a SIGFPE signal (which cannot be trapped).
Exit.
$ 
```

Not marvelous...

jpellegrini:
Maybe STklos could automatically restart in such situation?
Something like,

1. When called, `stklos` forks itself and wait for the child
2. The child (knowing it is the child) starts the VM
3. When SIGFPE, set a global "`please_restart`" variable
4. When the child exits, the parent checks the `please_restart` var, and decides to either fork again (if the value was 1) or exit (if the value was 0, which means the user requested to exit normally)

I didn't think for too long about this, so my idea may be totally wrong, but it would likely be a small change that ensures that STklos won't crash. When STklos restarts, it would print a message like
```
STklos was freshly restarted, and all bindings, objects, threads and everything 
else were removed.
```
Maybe there could even be a procedure to restart freshly, `(stklos-restart)`?

jpellegrini:
I tried to get this working in #902 but I'm not sure it's worth it. If it is not, we can close the PR and the issue.

PR #902, opened 2026-06-27
Restart after SIGFPE

jpellegrini:
This is an idea of how one could deal with #856 

It doesn't work perfectly yet, since I think I messed with signal handling and now, for example, `^C` causes STklos to exit, and after exiting STklos the terminal needs a `reset`, but it shows that we could, in theory, restart STklos like that.

I'm not sure if this is worth though, since all variables and everything will be lost on restart.

### DIFF
--- a/src/signal.c
+++ b/src/signal.c
@@ -49,6 +49,10 @@
  */
 static SCM signals[NSIG];
 
+extern int _stklos_restart;    /* buffer used to writ to the pipe below */
+extern int restart_pipefd[2];  /* pipe for communicating the REPL (or parent) that we want
+                                  to be restarted. */
+
 
 /* ====================================================================== */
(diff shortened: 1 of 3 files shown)


jpellegrini:
This would be much better using `pipe()`...  So a exit code wouldn't be reserved for restarting.
I'll work on it. But still not sure it makes sense! :grin:

jpellegrini:
Now using pipes!

But still somewhat clumsy... And it breaks the handling of some signals (now ^C exits the REPL).

jpellegrini:
This is `x.stk`:

```scheme
(display -10)
(newline)
(display  (expt 2 1000000000000000000))
(display " OK ")
(newline)
```

STklos now prints `-10`, but does not segfault. It exits (before writing OK - I can't tell why).

PR #903, opened 2026-06-28
Let parameters be declared immutable

jpellegrini:
For internal use only... But maybe useful (?)

### DIFF
--- a/src/parameter.c
+++ b/src/parameter.c
@@ -79,6 +79,9 @@ SCM STk_set_parameter(SCM param, SCM value)
 
   verify_parameter(param);
 
+  if (BOXED_INFO(param) & PARAMETER_CONST)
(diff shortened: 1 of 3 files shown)


egallesio:
Hi @jpellegrini,

I'm back. It was longer than expected, but I'm able to work (slowly). 
I've read your PR. I have no problem with your code, but some concern with the fact to lock the value of a parameter object. IMHO, that seems to me to be contradictory to the very concept of a parameter. 
Anyway, if you think this could be useful, I can happily merge your code.

jpellegrini:
> some concern with the fact to lock the value of a parameter object. IMHO, that seems to me to be contradictory to the very concept of a parameter.

Yes. That's why I included it as an internal procedure (name beginning with a `%`). 

Now I can't remember anymore why I wanted to make some parameter immutable... :grin:

PR #905, opened 2026-06-29
Support SRFI-261: Portable SRFI Library Reference

jpellegrini:
The SRFI is natively supported, and the modifications are:

1. `(srfi srfi-NNN)` is normalizes to `srfi/NNN` for library names
2. Some standard SRFI aliases would be added

However, I am not sure about implementation of (2).

Do we add `(%make-copy-module srfi/NAME srfi/N)` to all relevant SRFI implementations, in their `.stk` files under `lib/srfi`? I believe adding the names as aliases in `lib/srfi.stk` doesn't work.

### DIFF
--- a/SUPPORTED-SRFIS
+++ b/SUPPORTED-SRFIS
@@ -148,4 +148,5 @@ implemented in latest version is available at https://stklos.net/srfi.html):
     - SRFI-253: Data (Type-)Checking
     - SRFI-258: Uninterned symbols
     - SRFI-260: Generated symbols
+    - SRFI-261: Portable SRFI Library Reference
     - SRFI-264: String Syntax for Scheme Regular Expressions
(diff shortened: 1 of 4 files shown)


egallesio:
Hi @jpellegrini, 

I have changed a lot the structure of the SRFI   database (and moved it in `etc`) directory since it is not used by a running STklos (it is only used at compile time to build the `lib/srfi-data.scm` file).

You can now import srfi-180 with 
```scheme 
(import (srfi 180))
;; or  
(import (srfi json)
```
The documentation about using SRFIs have also updated (it was false on some points).

Now `cond-expand` and `require-feature` eventually load AND import the SRFI. So, in the REPL the `,r` is more convenient to use.

```scheme
stklos> ,r lists
stklos> (take '(1 2 3) 2)
(1 2)
```

PR #906, opened 2026-06-30
Add SRFI 267 (raw string syntax)

jpellegrini:
Tests are missing, because the test suite is a bit tricky to adapt. But I ran it and the tests and all worked fine!

I know STklos has herestrings, but well, it's a SRFI... :grin:


jpellegrini:
OK, tests added and it seems to work fine! :smile:

egallesio:
Hi @jpellegrini,

I have finally changed the implementation in pure Scheme. Finally, it was just

```scheme
    (%add-sharp-reader #\"
                       (lambda (port)
                         ;; We are sure that current character is the opening #\"
                         (read-char port)  ;; skip it
                         (read-raw-string-after-prefix port)))
```

Having an official syntax for entering weird strings is a good thing, and, you are right, it is fine to have in STklos.

PR #910, opened 2026-07-31
Fix SRFI 253 lambda-checked to check returns even in simple cases

aartaka:
Where the mere lambda was generated previously


aartaka:
CC @egallesio

PR #912, opened 2026-08-08
Make SRFI 273’ define-values-checked stricter

aartaka:
Sorry, noticed this bug only now. Thus a follow up :sweat_smile:


egallesio:
Merged! 
Thanks.

PR #913, opened 2026-08-12
Better `sqrt` for big rationals

jpellegrini:
```scheme
(define x (- (expt 9 -10000) 1))
(define q (* x x))
(define d (denominator q))
(define n (numerator q))

Before: (sqrt (+ 1 q)) => +nan.0
Now:    (sqrt (+ 1 q)) => 1.4142135623730951
```

When computing `sqrt` of a rational, first get the numerator and denominator and try to compute `sqrt` on them. if either is infinite, then we may try converting to inexact to see if it helps.  Of course, if the two square roots are finite, we keep them, for they're exact numbers.

The patch is only a bit large because of the comment -- otherwise, it's quite small.


egallesio:
Great. That's really better. 
Thanks @jpellegrini

PR #915, opened 2026-08-13
Update libffi from 3.5.2 to 3.8.0

jpellegrini:
There are some performance improvements, and there's also this new feature that could be useful in future STklos versions:

* `FFI_TYPE_VECTOR` (SIMD) type support with libffi-computed layout, for aarch64 and x86-64.

* `ffi_call_plan_size` — report the total memory a reusable call plan owns, for embedders that account for the memory held by long-lived plans.


egallesio:
Merged. Thanks.

PR #917, opened 2026-08-18
Fix exactness of exponentiation result

jpellegrini:
I know `number.c` is quite large, and this is mostly related to exactness, so if you think it's not interesting, @egallesio , we can close.

Before:
```scheme
(expt (+ (expt 7 100)) 1/2) => 1.79846504264741e+42  ; exact answer exists, not given!
(expt (+ (expt 7 101)) 1/2) => 4.75829124448823e+42  ; no exact answer
```

Now:
```scheme
(expt (+ (expt 7 100)) 1/2) => 1798465042647412146620280340569649349251249
(expt (+ (expt 7 101)) 1/2) => 4.75829124448823e+42
```

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -4379,6 +4379,55 @@ static SCM my_expt(SCM x, SCM y)
       return fixnum_exponent_expt(x, INT_VAL(y));
 
     case tc_rational:
+      if (INTP(x)) { /* Treat special cases where x = 0, 1 or -1 */
+        if (INT_VAL(x) == 0) return MAKE_INT(0);        //  0
+        if (INT_VAL(x) == 1) return MAKE_INT(1);        // +1
+      }
+      if (INTP(x) || BIGNUMP(x)) {
+        /* y = m/n */
+        SCM m = RATIONAL_NUM(y);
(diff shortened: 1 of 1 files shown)


egallesio:
We have a problem here. 

```scheme
(expt -1 1/2) => SIGSEGV
(expt 27 1/3) => SIGSEGV
(expt -inf.0 1/2)  => +inf.0+inf.0i  
;; BUT 
(sqrt   -inf.0) => 0+inf.0i
```

I have not looked at the cause for the SIGSEGV, but we are incoherent between `expt` with `y=1/2` and `sqrt`. Since there are a lot of corrections
for `sqrt` and `expt`, I'm just thinking if it would be more easy to have a solid implementation of `expt` and just implement `sqrt(x)` as `expt(x, 1/2)`.
What do you think?

jpellegrini:
Oops... I'll look deeper into this

jpellegrini:
@egallesio I see you merged this. I fixed the segfault in another PR.

Regarding the consistency with sqrt, I'll take a look.

I'm not  sure it would halp to have a sincle expt procedure. I'll look into the workings of it all again later

PR #921, opened 2026-09-02
Print procedure / closure with arity whenever possible

aartaka:
This small change improves REPL experience somewhat, helping people to remember the shape of procedure arguments even when named formals are not available. I personally missed it direly, coming from Chibi and Chicken.

There should be no trade-offs as it’s essentially constant string printing based on an integer constant dispatching.

@egallesio how does it look?

### DIFF
--- a/src/proc.c
+++ b/src/proc.c
@@ -75,6 +75,20 @@ static void print_lambda(SCM closure, SCM port, int mode)
   if (formals != STk_false) {
     STk_nputs(port, " ", 1);
     STk_print(formals, port, mode);
+  } else {
+    switch (CLOSURE_ARITY(closure)) {
+      case 0:  STk_fprintf(port, ""); break;
+      case 1:  STk_fprintf(port, " _"); break;
+      case 2:  STk_fprintf(port, " _ _"); break;
+      case 3:  STk_fprintf(port, " _ _ _"); break;
+      case 4:  STk_fprintf(port, " _ _ _ _"); break;
+      case 5:  STk_fprintf(port, " _ _ _ _ _"); break;
+      case -1: STk_fprintf(port, " . _"); break;
+      case -2: STk_fprintf(port, " _ . _"); break;
+      case -3: STk_fprintf(port, " _ _ . _"); break;
+      case -4: STk_fprintf(port, " _ _ _ . _"); break;
+      default: break;
+    }
   }
   STk_nputs(port,"]", 1);
 }
--- a/src/proc.c
+++ b/src/proc.c
@@ -77,16 +77,16 @@ static void print_lambda(SCM closure, SCM port, int mode)
     STk_print(formals, port, mode);
   } else {
     switch (CLOSURE_ARITY(closure)) {
-      case 0:  STk_fprintf(port, ""); break;
-      case 1:  STk_fprintf(port, " _"); break;
-      case 2:  STk_fprintf(port, " _ _"); break;
-      case 3:  STk_fprintf(port, " _ _ _"); break;
-      case 4:  STk_fprintf(port, " _ _ _ _"); break;
-      case 5:  STk_fprintf(port, " _ _ _ _ _"); break;
-      case -1: STk_fprintf(port, " . _"); break;
-      case -2: STk_fprintf(port, " _ . _"); break;
-      case -3: STk_fprintf(port, " _ _ . _"); break;
-      case -4: STk_fprintf(port, " _ _ _ . _"); break;
+      case 0:  STk_fprintf(port, " ()"); break;
+      case 1:  STk_fprintf(port, " (_)"); break;
+      case 2:  STk_fprintf(port, " (_ _)"); break;
+      case 3:  STk_fprintf(port, " (_ _ _)"); break;
+      case 4:  STk_fprintf(port, " (_ _ _ _)"); break;
+      case 5:  STk_fprintf(port, " (_ _ _ _ _)"); break;
+      case -1: STk_fprintf(port, " _"); break;
+      case -2: STk_fprintf(port, " (_ . _)"); break;
+      case -3: STk_fprintf(port, " (_ _ . _)"); break;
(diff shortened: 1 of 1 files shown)


jpellegrini:
Hello @aartaka !

I', not sure if @egallesio  woudl agree, as this makes the printed procedure take more space, but if he agrees, then wouldn't it be better to have a default way to represent large arities? For example, if a procedure has 8 parameters, the arity could be represented as `(- - - - - ...)`, or even `( <8 parameters> )`. For example,

```scheme
(define (x a b c) -1)
(define (y q w e r t y u i o p a s d f g h) -2)
```

x would have arity `(- - -)` but y would be `( <16 parameters> )`

aartaka:
@jpellegrini oh, right! I though that procedures have a restricted number of inspectable args, based on `%procedure-arity` definition. But it seems they indeed have unbounded arity. I agree that these need to be represented too, although I think that with longer arglists we get rapidly diminishing returns.

The matter of representation is important, but I really have no ideas beyond representing the arglist literally with these underscores (or dashes.) It’s most straightforward for interactive use. 

Representing things with `<X arguments>` et al. needs to happen only if it saves space, which, in case of `<X arguments>` only happens when there are, like, more than 7 args. It’s a rare case, so maybe we shouldn’t optimize for it? I mean, I’m even fine with this PR as-is, because I doubt many programmers write procedures with large arities in plain Scheme.

aartaka:
@jpellegrini I’m also not sure about your `- - - ...` notation. One of the most important pieces of info on a procedure is the _exact_ number of arguments it accepts. Approximating it with an ellipsis is quite confusing. So either we show these in full (long and maybe unnecessary,) explain them (with `<X arguments>` or the like,) or not even try (like this PR, limiting itself to 4-5 args.)

jpellegrini:
>  I doubt many programmers write procedures with large arities in plain Scheme.

OK,  but if a procedure (remember it may have been generated by a macro, and so it could have 200 parameters) has too many parameters, using `()` to represent them is misleading... It looks like you're saying "zero parameters"

jpellegrini:
> I’m also not sure about your `- - - ...` notation

OK, that was not a good example... I was just thinking that not printing anything when there areseveral arguments may be misleading.
But I really don't have much of a preference for the rest

aartaka:
@jpellegrini 

> > I doubt many programmers write procedures with large arities in plain Scheme.
> 
> OK, but if a procedure (remember it may have been generated by a macro, and so it could have 200 parameters) has too many parameters, using `()` to represent them is misleading...

There’s a way to avoid that: not listing any arglist. This is what the implementation in the PR does—it simply prints nothing when arity falls outside of expected numbers. See the `default:` case.

jpellegrini:
> prints nothing when arity falls outside of expected numbers

OK, I don't have a strong opinion here. @egallesio tell us what you think... :smile:

aartaka:
@jpellegrini you did raise a valid point though: it’s impossible to infer whether `#[procedure foo]` refers to:

- Procedure of 0 args
- Procedure of unknown / non-provided args
- Procedure of 5+ args

This problem can be solved through wrapping things in parens, much like Chicken does with `#<procedure (scheme#eval x . env)>`. Or by printing a number instead of an arglist, like Chibi does with `#<procedure eval 1+>`. Chibi’s approach is more scalable, but I really like these positional underscores, they are much more representative than some number.

aartaka:
Okay, I checked the formals code, and it

- prints a list when there are positional args
- and prints a symbol when there’s only the rest argument.

So our faux argument lists need to be wrapped into parens too. I just pushed a change doing that. It also solves the problem of not knowing whether the arguments are not provided or there are 0 of them. No parens—no arg info!

egallesio:
> > prints nothing when arity falls outside of expected numbers
> 
> OK, I don't have a strong opinion here. @egallesio tell us what you think... 😄

Perhaps, it's time to set the `compiler:keep-formals` parameter object to `#t`by default. It augments the size of the image, but the delta seems reasonable (even with current  RAM prices).  It makes things also a bit slower but it should not be noticeable. 
If we keep formals, I would really like to have them for primitives too. I have some results for that, and I would like to have something like chicken for that. Something like `#[primitive (cons x y)]` or `#[primitive (newline :optional port)`. 

More on this soon.

aartaka:
@egallesio so what are the next steps for this PR? I think it should be mergeable regardless of the formals fetching mechanism. Right?

aartaka:
@egallesio also, `#[primitive (cons x y)]` should probably be `#[primitive cons (x y)]`, as closures printed with formals have formals as a separate list.

aartaka:
@egallesio I went ahead and added arity printing to primitives too. I realize this is not a general solution you might’ve thought about, but it works miraculously, is fast, and simple to an extreme.

egallesio:
I'm sorry for my long delay to respond.

Primitive and lambdas have now their parameter printed. W

We have now
```scheme
        stklos> car                                      ;; a simple primitive  (a subr1)
        #[primitive (car x)]
        stklos> newline                             ;; another one (a subr2)
        #[primitive (newline #:optional port)]
        stklos> +                                         ;; a vsubr primitive (general case)
         #[primitive (+ . arg)]
        stklos> map                                    ;; a vsubr  primitive (with forced arguments in C source)
        #[primitive (map proc list ...)]
        stklos> cadr                                     ;; a lambda
        #[closure (cadr x)]
        stklos> (lambda (a b . c) 1)            ;; an anonymous lambda
        #[closure (#p7f9341f1f7c0 a b . c)]
```

About `map`, this is for now the only one where the default arguments were changed. There are a lot of other function that could be better documented. 

Parameters for primitives doesn't cost a lot (there are statically compiled as a C string). For the lambdas, even if the current commit displays the real parameters, perhaps it could be useful to take your code in the case where `compiler:keep-formals` is false. I would be better than no arguments at all. What do you think @aartaka?

aartaka:
@egallesio so this PR is no longer necessary? Nice!

PR #924, opened 2026-09-03
Fix (expt negative rational)

jpellegrini:
The result will be complex.

Before: `(expt -1 2/3) => 1`
Now:    `(expt -1 2/3) => -0.5+0.866025403784439i`


jpellegrini:
#931 should fix this too...

PR #925, opened 2026-09-04
Preserve complex exactness when mixing reals and exact complexes.

jpellegrini:
When we do `(* +3i 2.0)` we should get `+6.0i` because STklos has mixed exactness for complexes.

 However, the current code does `mul2(r1,o2)` and  we will have real part "0 times 2.0" which  does NOT result in 0, but in 0.0 (inexactness contaminates).  But in  this case it makes no sense. Since STklos uses mixed exactness in complexes, it makes sense to return `+6.0i`. So we test for `real part == 0`, and when it is, we just use plain `MAKE_INT(0)` when calling `make_compelx`. 

The same for division.

Almost all Scheme implementations seem to do this: when mixed exactness is supported, then the above results in `+6.0i`. The same is true for Common Lisp:

| system      | mixed complexes?  | (* 2.0 +3i) | note                  |
|-------------|------------------|-------------|-----------------------|
| bigloo      | no complexes     |             |                       |
| chez        | N                | 0.0+6.0i    |                       |
| chibi       | Y                | 0+6.0i      |                       |
| chicken     | N                | 0.0+6.0i    |                       |
| cyclone     | N (inexact only) | 6.0i        | 0.0+2i == +2i         |
| gambit      | Y                | +6.i        |                       |
| gauche      | N                | 0.0+6.0i    |                       |
| guile       | N                | 0.0+6.0i    |                       |
| kawa        | Y                | +6.0i       |                       |
| loko        | N                | 0.0+6.i     |                       |
| mit         | Y                | +6.i        |                       |
| sagittarius | N                | 0.0+6.0i    |                       |
| wile        | N                | 0.0+6.0i    |                       |
| ypsilon     | N                | 0+6.0i      | not sure what happens |
|-------------|------------------|-------------|-----------------------|
| ABCL        | N                | #C(0.0 6.0) |                       |
| Clisp       | Y                | #C(0 6.0)   |                       |
| ECL         | N                | #C(0.0 6.0) |                       |
| GCL         | N                | #C(0.0 6.0) |                       |
| SBCL        | N                | #C(0.0 6.0) |                       |

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -2627,7 +2627,18 @@ SCM STk_div2(SCM o1, SCM o2)
           return make_complex(div2(add2(mul2(r1, r2), mul2(i1, i2)), tmp),
                               div2(sub2(mul2(i1, r2), mul2(r1, i2)), tmp));
         }
-        case tc_real:     // fallthrough
+            /* Below: when we do (/ +6i 2.0)  we should get +3.0i.
+                      However, if we fall through and use mul2(r1,o2)
+                      we will have real part "0 / 2.0" which does
+                      NOT result in 0, but in 0.0 (inexactness contaminates).
+                      But in this case it makes no sense. Schemes do vary
+                      with respect to this, but since STklos uses mixed
+                      exactness in complexes, it makes sense to return
+                      +3.0i. So we test for r1 == 0. */
+        case tc_real:     if (r1 == MAKE_INT(0))
+                            return make_complex(MAKE_INT(0),
+                                                div2(i1, o2));
+                          // fallthrough
         case tc_rational: // fallthrough
         case tc_bignum:   // fallthrough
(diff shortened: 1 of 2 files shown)


egallesio:
Thanks this PR @jpellegrini .

It is merged. 

When working on this PR? I asked myself if the value `0+6i` should not be printed as `+6i`. After all the latter can be read back without problems. This is not the most common convention as seen in your table, but it seems nicer. What do you think?

jpellegrini:
> asked myself if the value 0+6i should not be printed as +6i

It would be more concise, and perhaps more readable (maybe not?)
Anyway, I'd be happy with it (although I would be even happier if it was configurable - but I know, that would be a little too much config granularity... :grin: )

PR #926, opened 2026-09-04
Treat exact zeros in make-polar

jpellegrini:
```scheme
(make-polar 0 x) => 0
(make-polar x 0) => x
```

(Exact zero magnitude means exact zero; exact zero  angle means exact zero imaginary part)
```
stklos> make-polar 
#[primitive (make-polar a m)]
```
Oops. The first argument is magnitude, and the second is the angle!


egallesio:
It is merged now.

PR #927, opened 2026-09-04
Name cut-generated procedures with %cut% and %cute%

aartaka:
This is useful in debugging. Throwaway lambdas made with cut/cute might be erroneous, and being able to see at a glance that they are cut-generated might narrow the bug search area tremendously.


aartaka:
@egallesio @jpellegrini I’ve amended the code to use actual named lambdas, as discussed in #928. I believe it’s ready for merge.

PR #931, opened 2026-09-04
Rewrite `my_expt` and `STk_sqrt`

jpellegrini:
Hi @egallesio !

I have rewritten `my_expt`. Not sure if it's good, but it does look cleaner. what do you think?

## Issues

1. ~There is still a test that fails, because... Even though the method for doing `(expt z 3)` (for complexes) is the **same** for doing `(* z z z)`, they seem to differ!~
2. There are lots of repeated tests in the `if`s, but the compiler (GCC) should be able to factor out those...
3. ~`sqrt` is still not using it. That would be a next step.~

## Changes

Split the cases in a more readable way:

```
1. expt_exact_positivefixnum
   /* x^y, with x EXACT and y POSITIVE FIXNUM */
   (Just renamed fixnum_exponent_expt)
2. my_expt_real_positivefixnum
   /* x^y, with x REAL and y POSITIVE FIXNUM */
4. my_expt_exact_x_rational_y
   /* x^y, with x EXACT and y RATIONAL */
5. expt_via_log
   /* x^y, GENERAL exp-log method */
```

Create helpers:

1. invert(x,y)
2. flip(x,y)

Rewrite `my_expt`:

```
      x         y     result
=====================================================
-----------------------------------------------------  y = 0:
                0     1
              0.0     1.0
-----------------------------------------------------  x = 0, 1, -1:
       1             1
      -1             1 or -1   * but y not complex!
       0     -a+bi   error
       0      < 0    error
       0      >=0    0
----------------------------------------------------- y = 1, 1.0, 0.0:
                1    x
     0.0      0.0    1.0
              1.0    inex(x)
     0.0    -a+bi    error
-----------------------------------------------------  x REAL:
    -1.0       big   x
    +1.0       big   +-x (parity of y)
    -0.0       big   +-0.0 (parity of y)
    +0.0      +big   +0.0
] 0.0,+1.0[   +big   +0.0
]-1.0, 0.0[   +big   0.0 * parity(y)
   +real      +big   +inf.0
   -real      +big   +inf.0 * parity(y)
    real   +fixnum   expt_real_positivefixnum
    -0.0             +-0.0 (parity of y)
     0,0             0.0
   -real      real   make-rect(0, imag-part (pow(x,y)))
   +real      real   pow(x,y)
//   +-0.0 +rational   0.0
//   +-0.0 -rational   +inf.0
   +real +rational   exp( log(x) * y )
   -real +rational   make-rect(0,  imag-part( exp( log(x) * y ) ) )
    real   complex   exp( log(x) * y )
    real       < 0   1 / exp(x,-y)
----------------------------------------------------- x EXACT except 0, 1:
   exact      +big   error
   exact   +fixnum   expt_exact_positivefixnum(x,y)
   exact   -fixnum   1 / expt_exact_positivefixnum(x,y)
   exact +rational   my_expt_exact_x_rational_y(x,y)
   exact -rational   1 / my_expt_exact_x_rational_y(x,y)
   exact     -real   make-rect(0,  imag-part( pow(x,y) )
   exact     +real   pow(x,y)
   exact   complex   exp( log(x) * y )
----------------------------------------------------- x COMPLEX:
 complex      +real  expt_exact_positivefixnum(x,y)      IF y represents int
 complex      -real  1 / expt_exact_positivefixnum(x,y)  IF y represents int
 complex      +int   expt_exact_positivefixnum(x,y)
 complex      -int   1 / expt_exact_positivefixnum(x,y)
 complex             exp( log(x) * y )
```

### DIFF
--- a/src/number.c
+++ b/src/number.c
@@ -4666,6 +4666,14 @@ static SCM my_expt(SCM x, SCM y) {
                                                  ? double2real(-0.0)
                                                  : double2real(0.0);
     if (REALP(x) && REALP(y))                return my_expt_real_real(x, y);
+    /* (expt -inf.0 1/2) => +inf.0i
+       (expt -inf.0 1/n) => +inf.0+inf.0i with signs changing accordig to the
+                            exp-log method. */
+    if (IS_INFP(x) &&
+        RATIONALP(y) &&
+        (RATIONAL_NUM(y)!=MAKE_INT(1) ||
+         RATIONAL_DEN(y)!=MAKE_INT(2)))        return expt_via_log(x,y);
+
     if (REALP(x) && RATIONALP(y))            return negativep(x)
                                                  ? make_complex(MAKE_INT(0),
                                                                 COMPLEX_IMAG(expt_via_log(x,y)))
--- a/src/number.c
+++ b/src/number.c
@@ -4323,6 +4323,83 @@ static SCM my_expt_real_positivefixnum(SCM x, SCM y) {
   return my_expt_basic(x, INT_VAL(y));
 }
 
+static inline int represents_exact(SCM x) {
+    /* Will return 1 iff x represents some number with integer or rational components exactly.
+       2/3    => 1
+       2      => 1
+       1.2    => 0
+       1.0    => 1
+       1+2i   => 1
+       2.0-3i => 1
+       2/3-4i => 1
+       2.2-4i => 0
+     */
+    return (isexactp(x) ||
+            REALP(x) && REAL_REPRESENTS_INT(REAL_VAL(x)) ||
+            (RATIONALP(x) && represents_exact(RATIONAL_DEN(x)) && represents_exact(RATIONAL_NUM(x))) ||
+            (COMPLEXP(x) && represents_exact(COMPLEX_REAL(x)) && represents_exact(COMPLEX_IMAG(x))))
+        ? 1 : 0;
+}
+
+static inline SCM expt_via_log(SCM x, SCM y) {
+    /* x^y, GENERAL exp-log method */
+   SCM z = my_exp(mul2(my_log(x),y));
+   /* The exp-log method above is not perfect, and it introduces
+      error. We'll try to fix.
+      If:
+
+         1. x is neither inifnite nor NaN and
+         2. The exponent is either a fixnum or rational which
+            does NOT contain bignums,
+
+      then we do 3 iterations of Newton's method, which should be
+      enough to get the best possible approximation to the actual
+      value.          */
+   if (( (finitep(x) && !STk_isnan(x)) || !REALP(x) ) &&
+       (INTP(y) || (RATIONALP(y) &&
+                    INTP(RATIONAL_DEN(y)) &&
+                    INTP(RATIONAL_NUM(y))))) {
+       /* y is either p/1 or p/q. */
+       SCM p = INTP(y) ? y           : RATIONAL_NUM(y);
+       SCM q = INTP(y) ? MAKE_INT(1) : RATIONAL_DEN(y);
+       /*
+         Newton's method:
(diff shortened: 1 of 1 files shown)


jpellegrini:
The failing test is this:

```scheme
(define x 0.33333333333333331482961625624739099293947219848633)

(- (* x x x) (expt x 3)) => 6.93889390390723e-18
```
However, at least so far as I can see, the method for computing `(expt x n)` when `n` is a positive fixnum is... To repeatedly multiply! So I can't tell exactly why they differ. I'll look into that later.

Anyway, what do you think?

I have not used the `switch(typeof(_))` style because some of the cases were like `isexactp(_)` and some special cases are like `if (x == MAKE_INT(0)`...

jpellegrini:
I found the problem. Will fix and commit...

jpellegrini:
OK, the new `my_expt`  is ready!

1. There are lots of repeated tests in the ifs, but the compiler (GCC) should be able to factor out those...
2. `sqrt` is still not using it. That would be a next step.

jpellegrini:
And I just committed one more patch to get `sqrt` and `expt ` consistent.

Before, using this PR:
```
1. (sqrt -inf.0)     => 0+inf.0i
2. (expt -inf.0 1/2) => +inf.0+inf.0i
    
3. (sqrt -10000000000000000000.0 )    => 0+3162277660.16838i
4. (expt -10000000000000000000.0 1/2) => 1.9363366072702e-07+3162277660.16838i
```
    
Now, after this last push, they agree.

egallesio:
Ouch, that looks impressive! Not had time yet to see effectively the code, but I'll do this asap. Thanks by advance.

jpellegrini:
I'm still working on this one, @egallesio ... Wait a bit more

jpellegrini:
@egallesio I think this one is ready. The commit message is more up to date than the PR description. It seems to work fine, and it's more precise than the previous version, besides being also more readable. And sqrt and expt now use the same algorithms!

jpellegrini:
By the way, now:
```scheme
(define x 0.33333333333333331482961625624739099293947219848633)

(- (* x x x) (expt x 3)) => 0.0
```

jpellegrini:
Also,
```scheme
(define z 2-1i)          ;; <= It is EXACT!
(define z3 (* z z z))
z3                        
 => -142+65i
(expt z3 1/3)            ;; EXACT exponent!
 => 2-1i                 ;; EXACT answer! :)
```

This is now unique to STklos (others will bring an inexact answer, and not exactly 2-1i). 

Of course, this won't always happen, because not always Gaussian integer a+bi, with a, b integers, will have the property that you take its nth power and take the principal value of the nth root, and will get the same number:

```scheme
(define z 2-5i)
(* z z z)                  => -142+65i
(define z3 (* z z z))
(expt z3 1/3)              => 3.33012701892219+4.23205080756888i
```
Also, the precision gain is limited. No miracles performed! :grin:

jpellegrini:
The only thing I'm not sure about is the Newton-Raphson refinement after doing expt via exp(log(...)). It does get more precision in several cases, but also makes the computation slower -- but only when inexact complexes are computed (in other cases the exp-log code is not used).
I even thought it could be possible to make it configurable... :grin:

egallesio:
Hi @jpellegrini. 

I have merged your (great) PR. While correcting a regression (the one expressed in PR #913) I have seen  new issue:

```scheme
stklos> (define foo (- (expt 10 1000) 1)) 
;; foo
stklos> (expt foo 1/2)
+inf.0
stklos> (expt foo 1/3)
+nan.0           ;; should be +inf.0
```
The case 1/2 is specially looked at, but in the general case, the Newton method seems to be incorrect with very big bignums. I was unable to correct it.  It would be cool if you have time to have a look at this problem.

jpellegrini:
@egallesio  I'll take a look!

PR #932, opened 2026-09-07
case-lambda: Parse args smarter to enable more useful arity inspection

aartaka:
This makes `case-lambda` to generate more inspectable and useful arglists, in the traces of [similar Chibi pull request](https://github.com/ashinn/chibi-scheme/pull/1189), for example:

``` scheme
(case-lambda ((a b c) 1) ((a b) 1))
;; #[closure 7efcf8226f80 (a b . args13)]
```

@egallesio how does it look?

### DIFF
--- a/lib/bonus.stk
+++ b/lib/bonus.stk
@@ -2352,17 +2352,34 @@ doc>
 doc>
 |#
 (define-macro (case-lambda . clauses)
-  (let ((len           (gensym))
-        (args          (gensym))
-        (len-prim      (gensym 'length))
-        (apply-prim    (gensym 'apply))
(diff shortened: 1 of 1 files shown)


egallesio:
Interesting!!! I 'll look to integrate this PR with my current version of the procedure printing. 
In any case, it is a very good idea.

egallesio:
It is merged now. We had some problem of hygiene which were, hopefully, detected by the tests.  
I also made some rewriting but, anyway everything is her now. 
Thanks @aartaka for this PR.

aartaka:
@egallesio thanks!

Issue #935, opened 2026-09-10
Wright-Cartwright-Shinn compatible matcher?


aartaka:
So [Wright-Cartwright-Shinn `match`](https://srfi.schemers.org/srfi-204/srfi-204.html) seems to be the majority consensus for pattern matching. While STklos uses Bigloo `match-case` with a peculiar syntax. This obstructs writing portable code by forcing one to write a lot of portability wrappers and unhygienic macros.

So it would be nice to have a Wright-Cartwright-Shinn `match` in STklos.

It can start as simple as [80 LoC match-simple](https://synthcode.com/scheme/match-simple.scm), if you don’t want to bring in the whole thing.

@egallesio @jpellegrini thoughts?

jpellegrini:
@aartaka I like the  idea! There was a SRFI proposed for that but it was withdrawn for lack of progress (204).
Maybe SRFI 241 could be interesting but as you already know it's not the de-facto standard (and it is a bit tricky to implement - it rebinds quasiquote).
So I'd say sure, it's a great idea.  @egallesio what do you think?

jpellegrini:
But if your goal is the full SRFI 204 API, see that STklos already has withdrawn SRFIs implemented. There was nothing inherently wrong with SRFI 204, it just wasn't making progress in the SRFI process.

aartaka:
@jpellegrini I don’t want to stall the progress, so I suggested `match-simple.scm` as the initial step. It’s roughly equivalent to the current Bigloo matcher and is a subset of SRFI 204, so it’s a good overall direction. And it can be further extended in case we ever need it!

egallesio:
I agree @aartaka that the Bigloo match case is no more convenient (it was a long time ago, but it seems a bit alien today). Furthermore, its compilation does not integrate well with current system. 

I wasn't aware of the existence of the [80 LoC match-simple](https://synthcode.com/scheme/match-simple.scm)  It is interesting. 
I'll take a look at it. However, I'm not sure if it will easily work in STklos, since it uses extensively hygienic macros which are badly implemented in STklos. 


> I suggested `match-simple.scm` as the initial step. It’s roughly equivalent to the current Bigloo matcher and is a subset of SRFI 204, so it’s a good overall direction. And it can be further extended in case we ever need it! 

Let's see what we can do with it.

aartaka:
@egallesio I can rewrite match-simple in `define-macro` if you want, it’s pretty simple (🥁)

[...]

PR #937, opened 2026-09-12
Add an adaptation of Alex Shinn’s match-simple

aartaka:
Hi! This adds a version of [match-simple.scm](https://synthcode.com/scheme/match-simple.scm) rewritten in non-hygienic `define-macro`-s. Feature set should be roughly the same, except that it also allows non-quoted constants (chars, strings, numbers, booleans,) where Shinn’s matcher would require quoting them.

One significant problem is that `_` pattern is breaking the macro. @egallesio any way to allow `_` as an aux syntax?

How does it look in general?


jpellegrini:
Already @aartaka ? You mentioned only two days ago that you'd  work on it... :smile:

@egallesio look at this:

```scheme
(pp (macro-expand '(match (list 1 2 3 4) ((a b c d) (+ a b c d)))))

(let ((expr2562 (list 1 2 3 4)))
  (cond ((and (pair? expr2562)
              (let ((car2563 (car expr2562))
                    (cdr2564 (cdr expr2562)))
                (and #t
                     (and (pair? cdr2564)
                          (let ((car2565 (car cdr2564))
                                (cdr2566 (cdr cdr2564)))
                            (and #t
                                 (and (pair? cdr2566)
                                      (let ((car2567 (car cdr2566))
                                            (cdr2568 (cdr cdr2566)))
                                        (and #t
                                             (and (pair? cdr2568)
                                                  (let ((car2569 (car cdr2568))
                                                        (cdr2570
                                                          (cdr cdr2568)))
                                                    (and #t
                                                         (null? cdr2570)))))))))))))
         (let* ((car2571 (car expr2562))
                (a car2571)
                (cdr2572 (cdr expr2562))
                (car2573 (car cdr2572))
                (b car2573)
                (cdr2574 (cdr cdr2572))
                (car2575 (car cdr2574))
                (c car2575)
                (cdr2576 (cdr cdr2574))
                (car2577 (car cdr2576))
                (d car2577)
                (cdr2578 (cdr cdr2576)))
           (+ a b c d)))))
```

Seems quite short, compared to the one you showed in #935

aartaka:
@jpellegrini I mean, I can shorten it even further, but I decided that a minimal implementation is more reviewable than an optimized one. To lessen the load on @egallesio :wink:

aartaka:
One thing that isn’t there in this PR is record destructuring. It’s definitely possible to do, and Bigloo match has it. But, again, let’s merge this first and extend later.

egallesio:
Thank you very much @aartaka for this contribution. It looks great.  
I can't wait to test it and integrate it into STklos.

egallesio:
> One significant problem is that `_` pattern is breaking the macro. @egallesio any way to allow `_` as an aux syntax?

I'll take a look at it.

aartaka:
>  One significant problem is that _ pattern is breaking the macro. @egallesio any way to allow _ as an aux syntax?
>
> I'll take a look at it.

Oh, no, sorry, that was silly me not importing (smatch). I’ve fixed it
and now work with broken tests! Will ping you once it all passes!

aartaka:
Okay I’ve fixed tests. I commented out one of them because supporting it
would take a lot of work for a really marginal corner-case. @egallesio,
this is ready for review.

aartaka:
> Okay I’ve fixed tests. I commented out one of them because supporting it
> would take a lot of work for a really marginal corner-case. @egallesio,
> this is ready for review.

To be fair, match-simple.scm does not pass this test either.

lib/smatch.stk
```diff
@@ -0,0 +1,304 @@
...
+      `(let ([,expr-var ,expr])
+         (cond
+          ,@(map
+             (lambda [clause]
+               (match-case
+                clause
+                [(?patt . ?body)
+                 `(,(clause->condition expr-var patt)
+                   (let* (,@(clause->bindings expr-var patt))
+                     ,@body))]))
```
egallesio: We should add a systematic else clause such as
 (else (error 'match "no matching clause for ~S" ,expr-var))

To yield an error when we have no matching clause

lib/smatch.stk
```diff
@@ -0,0 +1,304 @@
...
+           ,@clauses))))
+
+  (define-macro (match-let . body)
+    (if (symbol? (car body))
+        (let* ([loop-name (car body)]
+               [bindings (cadr body)]
+               [vars (map (lambda _ (gensym "match-let-var")) bindings)]
+               [body (cddr body)])
+          `(let ,loop-name (,@(map (lambda [var binding]
+                                     (list var (cadr binding)))
```
egallesio: should be binding[s]

egallesio:
@aartaka, some informations on this PR:
I'm still working on this PR. I have fixed little issues, but mainly, I rewrote it to not use `match-case` (the idea, is to replace it by your code, when it will be finalized). The remaining problem, I would like to solve is that 

```scheme
match '(1 2) ((a a) a) (_ 'no))
```
 returns 2 instead of `no`. 
I know that `match-simple.scm` does not manage this case, but it would be cool, if I could fix it.

aartaka:
> @aartaka, some informations on this PR:
> I'm still working on this PR. I have fixed little issues, but mainly, I rewrote it to not use match-case (the idea, is to replace it by your code, when it
> will be finalized).

Thanks for keeping me updated! 

> The remaining problem, I would like to solve is that
>
> match '(1 2) ((a a) a) (_ 'no))
>
> returns 2 instead of no.
> I know that match-simple.scm does not manage this case, but it would be cool, if I could fix it.

Wow, that’s a really hard problem! I didn’t dare to approach it,
sticking to match-simple feature set for implementability. But if you’re
feeling adventurous, then sure, your call!

Issue #943, opened 2026-09-17
`#e-0.1` => `+1/10`  (`#e` ignores the sign)


jpellegrini:
I'm working on the fix for this along with SRFI 270...

jpellegrini:
Also,

```
stklos> #e+3.21i
**** Error:
%read: bad sharp syntax in "#e+3"

stklos> #e-3.21i
**** Error:
%read: bad sharp syntax in "#e-3"
```

PR #944, opened 2026-09-17
Add SRFI 270 - Hexadecimal Floating-Point Constants

jpellegrini:
In `number.c`:

* added `<string.h>`, for `memmove`. This is because we use snprintf to convert numbers into hexfloat form, and it outputs them with a "0x" prefix, which needs to be removed. We could have used a for, but `memmove` is used in other parts of STklos.

* New function `number2Chexfloatstring` that converts a Scheme REAL into hexfloat format, and handles NaNs, infs and signed zero. This uses the GMP function `gmp_snprintf`.

* In number2Cstr, do not reject base 16 for `tc_real`. Instead, call the new function `number2Chexfloatstring`

* Function `compute_exact_real` now accepts a "base" argument. Two old problems were also fixed (see below).

* In the sample implementation `write-hexadecimal-float` was written in Scheme, but we need to make `number->string` also understand hex floats. We have implemented this in C, so that `number->string` with base 16 does the same. We then have just a wrapper in `lib/srfi/270.stk` just for calling `number->string`.

* Function `read_integer_or_real`: treat base 16 differently.
  - Do NOT change 'd' and 'f' into 'e' (obviously)
  - Note the place of 'p'
  - Other

* `Cstr2simple_number`: acept the first argument as pointer-to-pointer, and change it so we can skip the sharp markers.

* `STk_Cstr2number`: pass address of first pointer to `Cstr2simple_number`, and other minor adjustments.

Old problems fixed:

```
stklos> #e+3.21i
**** Error:
%read: bad sharp syntax in "#e+3"

stklos> #e-3.21i
**** Error:
%read: bad sharp syntax in "#e-3"
```


jpellegrini:
Wil Fix #943

